PR #16, opened 2016-11-08
Implement faction-based cost reduction locations

ystros:
Introduces a new class, `FactionCostReduction`, that reduces the cost of the next marshaled card for the faction specified in the constructor. `FactionCostReduction` replaces `DragonstonePort`, `BloodOrangeGrove` and `HeartTreeGrove` which were all identical minus the faction code used. 

It also introduces a function `factionCostReductionLocation` which takes a faction code and returns a card implementation that uses `FactionCostReduction` for that faction code. This function is now used to generate both the replaced reducer locations as well as implement the remaining reducer locations.

### DIFF
--- a/server/game/cards/locations.js
+++ b/server/game/cards/locations.js
@@ -35,11 +35,11 @@ locations['01040'] = {
     }
 };
 
-// 01059 - Dragonstone Port
-class DragonstonePort {
-    constructor(player, card) {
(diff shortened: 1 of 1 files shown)


cryogen:
:+1: This is exactly what I was going to do this evening, thank you :)

PR #19, opened 2016-11-09
Allow initiative modifying cards

ystros:
This allows cards that modify initiative (such as Kingsroad - see [full list](https://thronesdb.com/find?q=x%3AInitiative&sort=name&view=card)) to be implemented. The game now asks `Player` for a list of all cards in play that modify the initiative. The player determines this by looking for an initiative value on the card - plots inherently have one, it can be added to cards that modify initiative.

A similar approach can be taken for cards that modify the amount of gold collected at the start of the marshal phase, as well as cards that modify the reserve value. I do have a couple concerns that I'd like feedback on, which I'll detail as separate comments on the code.

As always, feel free to scrap if you don't like this direction.

### DIFF
--- a/server/game/cards/locations.js
+++ b/server/game/cards/locations.js
@@ -2,6 +2,16 @@ const _ = require('underscore');
 
 var locations = {};
 
+// 01039 - The Kingsroad
+locations['01039'] = {
+    register: function(game, player, card) {
+        // TODO: Full Kingsroad behaviour.
+        card.initiative = 1;
+    },
+    unregister: function(game, player, card) {
+    }
+};
+
 // 01040 - The Roseroad
 class TheRoseRoad {
     constructor(player, card) {
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -210,10 +210,6 @@ class Game extends EventEmitter {
                     highestInitiative = totalInitiative;
                     highestPlayer = p;
                 }
-                var explanation = _.map(initiativeCards, c => {
-                    return c.name + ': ' + c.initiative;
-                });
-                this.addMessage(p.name + ' has ' + totalInitiative + ' initiative (' + explanation + ')');
             });
 
             _.each(this.players, p => {
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -202,10 +202,7 @@ class Game extends EventEmitter {
             var highestPlayer = undefined;
             var highestInitiative = -1;
             _.each(this.players, p => {
-                var initiativeCards = p.initiativeCardsInPlay();
-                var totalInitiative = _.reduce(initiativeCards, (memo, c) => {
-                    return memo + c.initiative;
-                }, 0);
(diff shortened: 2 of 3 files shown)


cryogen:
I like the idea, but I might go a slightly different route with something like Player having getTotalInititive() and getTotalIncome(), which uses the same sort of logic that you're proposing.  It's probably better than the implementation I currently have for roseroad.

ystros:
I can certainly add a method for getting the total. I think the list of cards will be needed in at least certain situations (plot Snowed Under which ignores initiative from the plot card but not other cards, plot Rains of Autumn which ignores gold derived from characters + locations). But it may not be necessary to use directly right now.

I did use the full list for the initiative calculation message to players. Let me know if you think that's too verbose and I'll remove.

Not sure you saw my review comments I made about how initiative was being set on cards, so I'll reproduce here:

I don't particularly like this approach of modifying the card (in the implementation for Kingsroad), but it worked as a proof-of-concept. It probably would be better if this value was in the database that the card info is pulled from. I know that the data is currently pulled from ThronesDB and stored locally.

It may make more sense to add an additional setup step (either in separate migrations file or in fetchdata.js) that augments the card data with additional metadata such as gold, initiative, reserve, etc modifiers. You might be able to read it from the text description from ThronesDB (often says things like "+1 Initiative") but that could get iffy. Thoughts?

Perley:
I guess we could add more metadata but will that necessarily take effect when the card is added or removed from play? We still need logic to apply and remove the modifiers (in the case more of initiative as that is more fluid) to a player at the appropriate times. I think storing it with a player as two values that dynamically get updated when appropriate cards are played makes the most intuitive sense to me. It's easily referenceable and I think more easy to calculate the total initiative/gold values than wrapping it in methods. A little crude yes but why get fancy?

cryogen:
My original plan was to do it like how the plots do it.  Modify the values on the card or in the state when the card is registered(put into play) and restore the original value when it is unregistered(leaves play).

In your example, Snowed under would have a plotRevealed handler that sets both plots initiative to 0 on reveal and then in unregister it will restore both plots back to their initiative values.

So the getInitiative() method on the player would look something like(very not valid pseudocode):

return this.activePlot.card.initiative + allCards.sum(card.initiative)

This falls down a bit when multiple cards affect a plot's initiative(not sure if there are any situations where this happens) but it can be worked around

I'm ok with the modifier value being in the card's implementation, I've always wanted to keep the static data (from thronesdb) separate to the implementation of the card.

I think the full list of modifiers is unnecessary.  I've tried to keep chat messages to when interactions happen.  If the initiative/gold modifier is printed on the card, then that should be enough for players to see what's going on.  As a game engine we need to explain interactions so that the players know what's happening and when to correct things.  For example when text on a card is resolved automatically.

Am I making any sense? I feel like I'm rambling

ystros:
Looks like this was ported in commit 7fd95eaffd10d0e6fa8864be81526ec616c9d13e. Closing!

PR #20, opened 2016-11-09
Small power comparison fixes and power claim fix

DrThyme:
Some small things I noticed while reading through the code, now since my node.js is currently quite rusty these fixes might ruin intended behavior and in that case feel free to ignore this pull request. But I'd thought it good to point them out. :)

Power comparison fix: All winner.getTotalPower() check for "> 15", wouldn't this lead to a player needing 16 power to win?

Power claim fix: When applying power claim to a losing player the code currently only checks to see if that player has any power (ie. loser.power > 0) and then subtracts the claim from that players power and adds it to the winner. This could lead to a player having negative power/a player getting power that didn't exist. An example would be that the losing player has 1 power and has just lost a power challenge with claim 2. 
This fix is supposed to correct that by taking the claim in steps of one power until there is no power left or no claim left. Might be fancier ways of fixing this problem though.

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -699,8 +699,8 @@ class Game extends EventEmitter {
             } else if (winner.currentChallenge === 'power') {
                 while (claim > 0) {
                     if (loser.power > 0) {
-                        loser.power -= 1;
-                        winner.power += 1;
+                        loser.power--;
+                        winner.power++;
                         claim--;
 
                         if (winner.getTotalPower() => 15) {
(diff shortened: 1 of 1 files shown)


cryogen:
:+1: Good catch.  This fixes an issue someone mentioned yesterday that I'd forgotten to document.  Just one thing, can you change the -= 1 and += 1 to -- and ++?  Be happy to merge it then :)

cryogen:
As for the win check, that was a quick and dirty implementation i did before i publicised the site a little as i thought people might want to know when they've won :) It need's to be better and preferably refactored to something like this.hasWon()

DrThyme:
Great! Fixed it to ++ and -- in the code. :)

PR #22, opened 2016-11-09
Bug fixes

Perley:
Fixes to ties on initiative and dominance. As well some small semantic changes to make it consistent with game language for readability (feel like this should be a thing but perhaps that's just me).

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -229,7 +229,7 @@ class Game extends EventEmitter {
             initiativeWinner.buttons.push({ command: 'firstplayer', text: 'Opponent', arg: 'opponent' });
         }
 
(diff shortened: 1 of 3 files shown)


server/game/game.js
```diff
@@ -257,20 +257,48 @@ class Game extends EventEmitter {
...
 
+            // determine initiative winner
             _.each(this.getPlayers(), p => {
-                p.revealPlot();
+                var playerInitiative = p.activePlot.card.initiative;
+                var playerPower = p.power;
+
+                if (playerInitiative === highestInitiative) {
+                    if (playerPower === highestPower) {
+                        var diceRoll = Math.floor((Math.random() * 20));
```
cryogen: _.random will keep is consistent with other places we ask for random numbers

PR #35, opened 2016-11-11
Always check win condition when power is gained.

ystros:
Previously power was being modified directly in multiple places and the win condition wasn't being checked in all of them. Additionally, the win condition was duplicated in multiple spots. This should make things a bit more consistent.

In the future this could also emit a 'powerGained' event, which may be needed for cards like [Edmure Tully](https://thronesdb.com/card/04041) and [Eddard Stark (WotN)](https://thronesdb.com/card/03003).


cryogen:
👍 💯

PR #53, opened 2016-11-12
Milk rework

cryogen:


server/game/player.js
```diff
@@ -24,10 +25,10 @@ class Player extends Spectator {
     searchDrawDeck(limit, predicate) {
         var cards = this.drawDeck;
 
-        if (_.isFunction(limit)) {
+        if(_.isFunction(limit)) {
```
ystros: Not related to the substance of this PR but the ESLint rules should be updated if this is the desired styling.

PR #61, opened 2016-11-16
Implement card preview in chat

ystros:
Reworks `Game.addMessage` to work as a general-purpose formatter when constructing messages for chat. It may now take a list of optional arguments that are interpolated into the message. 

```
// server side
this.addMessage('{0} discards {1}', player.name, card);

// what appears client side
Player 1 discard Tyrion Lannister
```

* In general, the usage of addMessage going forward should favor argument replacement over concating strings together. This provides consistency, allows other rich data to be interpolated into the message (player data, challenge type + icon, etc), and allows a way forward in the future should translation ever be desired.
* Arguments can be either strings or JSON data. Currently no enforcement of what type of JSON data, is left up to the client to interpret. 
* On the client side, JSON data is converted to the appropriate JSX. In this specific PR, when card JSON is passed to the client it will render the card name and allow players to hover over it for the card preview image. 

### Preview
chat-card-hover

This PR also converts all usages of addMessage to use the new formatter instead of manually building strings. I know this may conflict w/ the ongoing refactor, so let me know if you want me to re-PR it without the conversions, just the feature itself.

### DIFF
--- a/client/GameComponents/Messages.jsx
+++ b/client/GameComponents/Messages.jsx
@@ -15,6 +15,8 @@ class InnerMessages extends React.Component {
         this.state = {
             message: ''
         };
+
+        this.formatMessageText = this.formatMessageText.bind(this);
     }
 
     componentDidUpdate() {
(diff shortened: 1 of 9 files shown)


client/GameComponents/Messages.jsx
```diff
@@ -24,12 +26,30 @@ class InnerMessages extends React.Component {
...
+            return <div key={'message'+index++} className='message'>{this.formatMessageText(message.message)}</div>;
         });
 
         return messages;
     }
 
+    formatMessageText(message) {
+        var index = 0;
+        return _.map(message, fragment => {
+            if(fragment.code && fragment.label) {
```
ystros: This check felt hacky but we may be able to wait until we introduce other argument types (e.g. challenge icons) to deal with it.

server/game/game.js
```diff
@@ -22,7 +22,25 @@ class Game extends EventEmitter {
...
     addMessage(message) {
-        this.messages.push({ date: new Date(), message: message });
+        var args = Array.from(arguments).slice(1);
+        var formattedMessage = this.formatMessage(message, args);
+        this.messages.push({ date: new Date(), message: formattedMessage });
+    }
+
+    formatMessage(format, args) {
+        var messageFragments = format.split(/(\{\d+\})/);
+        return _.map(messageFragments, fragment => {
```
ystros: This whole method is still pretty ugly but at least it's server side now. Client is only responsible for rendering what the server tells it now.

PR #79, opened 2016-11-24
Fix keyword related issues

### LINKED ISSUE #76: Cards containing keywords in their text, but not keywords themselves, trigger those keywords
cards containing keywords in their text, but not keywords themselves, trigger those keywords.

Examples: 
1. Septa Mordane gives renown to Sansa Stark, but triggers renown on her own challenges.
2. Doreah gains insight only when attacking together with Lord or Lady, but triggers it on her own.

ystros:
* Parses out a list of keywords when a card is constructed based on the first line of the card text. Only exact matches are considered.
* No longer considers 'no attachment' to be a keyword. Instead, it's tracked separately and allows characters to exclude attachments that don't match a certain trait - [most notably weapon](https://thronesdb.com/find?q=x%3A%22no+attachments+except%22&view=card&sort=name&page=1).
* It provides a place that 'ambush' can be parsed in the future.

This PR makes me sad.

Fixes #76 .

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -26,13 +26,26 @@ class BaseCard {
         this.tokens = {};
 
         this.menu = _([]);
-        this.keywords = this.parseKeywords(cardData.text || '');
+        this.parseCard(cardData.text || '');
     }
 
-    parseKeywords(text) {
+    parseCard(text) {
         var firstLine = text.split('\n')[0];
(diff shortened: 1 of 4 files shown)


cryogen:
At least we only have to parse once, when the deck is selected.. and maybe we could even do this processing when we sync from thronesdb, in the future

PR #130, opened 2016-12-11
Card locations

ystros:
* Adds a location property on DrawCard to indicate where it is currently.
* Adds a moveCard method on Player to handle removing the card from its current position, adding it to the target position, and setting its new location.
* Somewhat simplifies drag-and-drop code due to the generalized moveCard method.
* Reimplements Player.discardCard as a saveable discard.
* Introduces a sacrificeCard method on Player for sacrifice abilities.
* Fixes various oddities on core Arya Stark implementation (cards ending up in discard pile face down, crashes when an attachment is drawn as her duplicate).

**Future improvments:**
* DrawCard.inPlay should be converted to a property that checks whether the current location is 'play area'.
* Introduce either constants or a set of enumerable objects for the locations instead of magic strings everywhere.
* Player.playCard should probably be cleaned up and no longer take the source list directly.
* Ultimately I think some of this behaviour belongs on card itself - the pattern of card.owner.discardCard(card) (controller instead of owner, move instead of discard, etc) comes up a few times.

Did my best to preserve the spirit of the existing tests if not the letter of it (converted a few things that were testing for internal method calls to test for the overall effects)


server/game/gamesteps/attachmentprompt.js
```diff
@@ -24,20 +24,8 @@ class AttachmentPrompt extends UiPrompt {
...
-            player.cardsInPlay = player.removeCardByUuid(player.cardsInPlay, attachmentId);
-        } else {
-            targetPlayer.attach(player, attachment, targetCard.uuid);
-            player.removeFromHand(attachmentId);
-        }
-
-        if(player.dropPending) {
-            player.discardPile = player.removeCardByUuid(player.discardPile, attachmentId);
-        }
+        targetPlayer.attach(player, attachment, targetCard.uuid);
```
cryogen: 👍

PR #146, opened 2016-12-14
document /pillage command in README

zacchiro:


cryogen:
Thanks! Want to put it on the new about page too?

PR #239, opened 2017-01-19
Allow event reactions to play at challenge prompt.

ystros:
Previously, event-based reactions were only playable if there was a
pause in the DUCK phase (e.g. another reaction fires, then the player
could play the event reaction card before completing the original
reaction). This made several cards hard to play.

This change allows such event cards to be played after the challenge has
been completed but before the next challenge has been initiated. It does
so by not immediately clearing the Game.currentChallenge property when a
challenge has been completed.


ystros:
After further consideration, these events are already playable at the Apply Claim prompt, so this change is unneeded. Playing won/lost events there is closer to where they should happen anyway (post U instead of post K in DUCK).

PR #241, opened 2017-01-19
reworking reconnecting to be less stupid and more resilient

cryogen:
Summary is:

remove the leavegame message and let the client figure out when the game is gone

Don't remove a player from the game when they disconnect, just mark them as disconnected

### DIFF
--- a/client/Application.jsx
+++ b/client/Application.jsx
@@ -81,16 +81,6 @@ class App extends React.Component {
             this.props.receiveJoinGame(game);
         });
 
-        socket.on('leavegame', (game, player) => {
-            var isMe = false;
-
-            if(player === this.props.username) {
-                isMe = true;
-            }
-
-            this.props.receiveLeaveGame(game, isMe);
-        });
-
         socket.on('gamestate', game => {
             this.props.receiveGameState(game, this.props.username);
         });
--- a/client/Application.jsx
+++ b/client/Application.jsx
@@ -168,7 +168,6 @@ App.propTypes = {
     receiveGameState: React.PropTypes.func,
     receiveGames: React.PropTypes.func,
     receiveJoinGame: React.PropTypes.func,
-    receiveLeaveGame: React.PropTypes.func,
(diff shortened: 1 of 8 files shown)


server/index.js
```diff
@@ -454,17 +413,45 @@ io.on('connection', function(socket) {
...
-        
-        if(!game.finishedAt) {
-            game.finishedAt = new Date();
-            game.saveGame();
-        }
-
         runAndCatchErrors(game, () => {
-            removePlayerFromGame(game, socket, 'has left the game');
+            if(game.players[socket.request.user.username]) {
+                game.players[socket.request.user.username].noReconnect = true;
```
ystros: There doesn't seem to be a difference between `Player.left` and `Player.noReconnect` anymore, as they're both always set in this method. Consider removing `Player.noReconnect` and checking `left` instead in the socket connection handler for reconnecting.

client/Application.jsx
```diff
@@ -81,16 +81,6 @@ class App extends React.Component {
...
         });
 
-        socket.on('leavegame', (game, player) => {
-            var isMe = false;
-
-            if(player === this.props.username) {
-                isMe = true;
-            }
-
-            this.props.receiveLeaveGame(game, isMe);
```
ystros: `receiveLeaveGame` isn't being used anywhere now. Remove the React property for the handler.

client/reducers/games.js
```diff
@@ -46,12 +46,15 @@ function games(state = {
...
             });
 
-            var currentState = state.currentGame;
+            var currentState = retState.currentGame;
+            if(!currentState) {
+                return retState;
+            }
 
-            if(currentState && currentState.players[action.username] && currentState.players[action.username].left) {
+            if(!currentState || !currentState.players[action.username] || currentState.players[action.username].left) {
```
ystros: This change makes spectating impossible. When joining as a spectator, it hits this line and immediately deletes the game state since spectators are not in the players list.

ystros:
Fixed the spectator problem I was seeing. 👍

PR #248, opened 2017-01-22
added isKneeled getter and Obara Sand

ryangately:


ystros:
Hi there! Thanks for contributing. A couple issues:

1. `challenge` isn't defined in your `isKneeled` method, so this code will crash. You could get the current challenge from `this.game.currentChallenge`. However:
2. In general, we've been moving away from overriding methods to provide abilities like this in favor of declaring their abilities. Take a look at the implementation for [Doreah](https://github.com/cryogen/throneteki/blob/master/server/game/cards/characters/04/doreah.js) for an example of a character whose ability only triggers during certain challenges.
3. You would need to implement an effect in [effects.js](https://github.com/cryogen/throneteki/blob/master/server/game/effects.js) that sets a flag on the card, then modify the "can add to challenge" check to not care if the card is kneeling when that flag is set.

cryogen:
I'd echo what @ystros has already said, and also this needs to be rebased now.

Thanks for your efforts!

PR #276, opened 2017-01-28
Password reset

cryogen:


client/Application.jsx
```diff
@@ -98,19 +101,35 @@ class App extends React.Component {
...
         if(index !== -1) {
             path = path.substr(index, 11);
             pathArg = this.props.path.substr(11 + 1);
         }
 
+        index = path.indexOf('/reset-password');
+        if(index !== -1) {
+            path = path.substr(index, 15);
+            idArg = this.getUrlParameter('id');
+            tokenArg = this.getUrlParameter('token');
```
ystros: Is it not possible to use `this.props.location.query` to retrieve these parameters instead of manually parsing w/ regexes?

server/api/account.js
```diff
@@ -1,75 +1,99 @@
...
     server.post('/api/account/check-username', function(req, res) {
-        db.collection('users').findOne({ username: { '$regex': new RegExp('^' + escapeRegex(req.body.username.toLowerCase()), 'i')}},
-        function(err, account) {
-            if(err) {
-                res.send({ message: '' });
-                logger.info(err.message);
-                return;
-            }
+        userRepository.getUserByUsername(req.body.username).then(user => {
+            logger.info(user);
```
ystros: Did you intend to keep this log message here? If so, printing out context information in addition to the user object would be useful.

server/api/account.js
```diff
@@ -1,75 +1,99 @@
...
+            if(user) {
                 res.send({ message: 'An account with that name already exists, please choose another' });
                 return;
             }
 
             res.send({ message: '' });
+        }).catch(err => {
+            logger.info(err);
+            res.send({ message: '' });
+            logger.info(err.message);
```
ystros: Double logging the error?

server/api/account.js
```diff
@@ -82,4 +106,105 @@ module.exports.init = function(server) {
...
+        userRepository.getUserById(req.body.id).then(user => {
+            if(!user) {
+                throw new Error('User id not found ' + req.body.id);
+            }
+
+            if(!user.resetToken) {
+                throw new Error('Got unexpected reset request for user ' + user.username);
+            }
+
+            resetUser = user;
```
ystros: I'm kind of iffy on setting external variables like this vs explicitly returning the data and piping it into the next `then` step. But it's probably fine? Might just be a purity thing on my part.

client/Application.jsx
ystros: Nevermind, it seems that this is part of react-router which isn't being used (and has since changed to `getQuery()` anyway)

server/api/account.js
cryogen: This bothers me too but the .update promise from monk doesn't callback with the db object so i wasn't sure how to do it

PR #291, opened 2017-02-02
Manage players through the `Game` class

ystros:
Previously, a lot of the work of adding players / spectators, managing
disconnecting, reconnecting and leaving games was being managed manually
through the socket handlers. This code has now been moved into the
`Game` class itself.

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -5,6 +5,7 @@ const uuid = require('node-uuid');
 const GameChat = require('./gamechat.js');
 const EffectEngine = require('./effectengine.js');
 const Effect = require('./effect.js');
+const Player = require('./player.js');
 const Spectator = require('./spectator.js');
 const GamePipeline = require('./gamepipeline.js');
 const SetupPhase = require('./gamesteps/setupphase.js');
@@ -55,6 +56,10 @@ class Game extends EventEmitter {
 
     isSpectator(player) {
         return player.constructor === Spectator;
+    }
+
+    hasActivePlayer(playerName) {
+        return !!this.players[playerName] && !this.players[playerName].left;
     }
 
     getPlayers() {
@@ -683,16 +688,6 @@ class Game extends EventEmitter {
         this.gameChat.addChatMessage('{0} {1}', player, message);
     }
 
-    playerLeave(playerName, reason) {
-        var player = this.players[playerName];
-
-        if(!player) {
-            return;
-        }
-
-        this.addMessage('{0} {1}', player, reason);
-    }
-
     concede(playerName) {
         var player = this.getPlayerByName(playerName);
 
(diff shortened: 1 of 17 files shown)


server/index.js
```diff
@@ -447,30 +435,12 @@ io.on('connection', function(socket) {
...
-
-                if(game.started && !game.finishedAt) {
-                    game.finishedAt = new Date();
-                    game.saveGame();
-                }                
-            }
+        runAndCatchErrors(game, () => {
+            game.leave(socket.request.user.username);
+            // Force update of game state to make sure spectators get disconnected.
+            socket.emit('gamestate', game.getState(socket.request.user.username));
```
ystros: Most of this refactor was just pushing existing code around, but this is an actual change. This used to only happen if a spectator left, but I didn't want to send in the socket / io object into game to only do it for spectators. I think it's fine (brief testing didn't come up with anything) but please review carefully.

server/game/game.js
```diff
@@ -57,6 +58,10 @@ class Game extends EventEmitter {
         return player.constructor === Spectator;
     }
 
+    hasActivePlayer(playerName) {
+        return !!this.players[playerName] && !this.players[playerName].left;
```
cryogen: Checking for truthiness is fine here, rather than forcing it to bool, but I'm happy to leave it as is too

server/game/game.js
```diff
@@ -829,14 +824,76 @@ class Game extends EventEmitter {
...
+            return false;
+        }
+
+        this.players[user.username] = new Player(socketId, user, this.owner === user.username, this);
+
+        return true;
+    }
+
+    isEmpty() {
+        return _.all(this.players, player => player.disconnected || player.left);
```
cryogen: How do you feel about renaming players to playersAndSpecatators?  Would you rather do it in a separate commit?

server/game/game.js
```diff
@@ -829,14 +824,76 @@ class Game extends EventEmitter {
...
+    }
+
+    leave(playerName) {
+        var player = this.players[playerName];
+
+        if(!player) {
+            return;
+        }
+
+        this.addMessage('{0} has left the game', player);
```
cryogen: One idea i had recently, that your implementation now makes easier, is to notify specifically if it's a spectator that's left/disconnected - dunno about anyone else but i find myself having to check if it was my opponent who left or not :)

server/index.js
cryogen: I actually think this is unnecessary as the game state is sent to all players and spectators below, if the game is still around by then

ystros: Just tested, this line is explicitly needed for spectators. This is because we delete the spectator from the players array, so when it updates the game state for everyone in the list they won't receive it (since they're not in the list). Without some sort of explicit update, the "Leave" button for spectators doesn't revert the UI back to the game list.

cryogen: That sounds familiar!  Probably why I added it in the first place :)

server/game/game.js
ystros: In general, I try to explicitly return a bool for methods that have `has` or `is` in the name.

server/game/game.js
ystros: Will do as part of this PR. 👍

server/game/game.js
ystros: Seems like a good candidate for a separate PR / commit after this refactor.

PR #305, opened 2017-02-04
Implement Alannys Greyjoy

DukeTax:


cryogen:
I think this is ok to go in, just would like @ystros to glance over the effect stuff

server/game/cards/characters/01/alannysgreyjoy.js
```diff
@@ -0,0 +1,16 @@
+const DrawCard = require('../../../drawcard.js');
+
+class AlannysGreyjoy extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: player.firstPlayer,
```
ystros: `player` doesn't exist in this context so it will crash. Also, condition must always be a function (and needs to be in this case anyway since first player can change each round). So you want this instead:
```javascript
condition: () => this.controller.firstPlayer,
```

Otherwise, looks good to me.

DukeTax:
Thanks for the feedback. If you don't mind me asking beginner questions: why doesn't player.firstPlayer exist in this context when it does in a similar context in the code for raidinglongship.js?

cryogen:
raiding longship is an action, which means a method is called when the player clicks on the menu item, this callback has the player as an argument

reactions and interrupts just happen in response to game events, it's not always possible to determine a player, so this.controller is used to mean "the player who currently controls this card".  Hope this helps

PR #312, opened 2017-02-06
Implement Greatjon's Vanguard

DukeTax:
Is there a way to implement 'no attachments' and the like? Didn't see it on other cards.


ystros:
Most attachment restrictions ('no attachments', 'No attachments except X') are parsed from the card text, no need to do anything.

DukeTax:
Cool!

PR #313, opened 2017-02-06
Implement Ser Ilyn Payne

DukeTax:
Not sure if card.getCost() is a thing.

### DIFF
--- a/server/game/cards/characters/02/serilynpayne.js
+++ b/server/game/cards/characters/02/serilynpayne.js
@@ -1,0 +1,37 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SerIlynPayne extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Kneel Ser Ilyn Payne to kill a character with printed cost 3 or less',
+            method: 'killChud'
+        });
+    }
+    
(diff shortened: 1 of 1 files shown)


server/game/cards/characters/02/serilynpayne.js
```diff
@@ -0,0 +1,37 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SerIlynPayne extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Kneel Ser Ilyn Payne to kill a character with printed cost 3 or less',
```
cryogen: Try to condense this, it's a little wordy and wont look so great on the menu

server/game/cards/characters/02/serilynpayne.js
```diff
@@ -0,0 +1,37 @@
...
+            method: 'killChud'
```
cryogen: can we call this something better? :)

cryogen:
The spacing is a little off in this file.  Are you using an editor that supports eslint? If not run eslint manually on the source to find issues like this

DukeTax:
I have just been typing directly on here so far, but I'm using an editor with eslint now. :)

PR #320, opened 2017-02-07
Reapply effects upon take control.

### LINKED ISSUE #317: Dominance Calculation
When I Night Gathers .../Yoren opponent character it appears it calculates 2x their STR for dominance.

untitled

### LINKED ISSUE #319: Stealing an Arbor (with Euron) doesn't add gold during the following marshalling phase
See title.

ystros:
* Adds a /take-control command (useful for debugging purposes + cards that haven't been implemented yet)
* Recalculates effects once a card changes controller to ensure the new controller receives the benefits of the effect (and the old controller loses it).

Fixes #317.
Fixes #319.

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -653,14 +653,19 @@ class Game extends EventEmitter {
             }
         }
 
-        if(message.indexOf('/take-control') === 0) {
+        if(message.indexOf('/give-control') === 0) {
             this.promptForSelect(player, {
                 activePromptTitle: 'Select a character',
-                waitingPromptTitle: 'Waiting for opponent to take control',
-                cardCondition: card => card.location === 'play area' && card.controller !== player,
+                waitingPromptTitle: 'Waiting for opponent to give control',
+                cardCondition: card => card.location === 'play area' && card.controller === player,
                 onSelect: (p, card) => {
-                    this.takeControl(player, card);
-                    this.addMessage('{0} uses the /take-control command to control {1}', p, card);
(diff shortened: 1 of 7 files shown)


cryogen:
I deliberately didn't implement a take control because of the (admittedly low) chance of it being abused by someone wanting to cause annoyance. If we're going to have it, can it be /give-control instead so we maintain ownership of a persons game state?

ystros:
Good point. Griefers ruin everything.

server/game/game.js
```diff
@@ -653,6 +653,27 @@ class Game extends EventEmitter {
             }
         }
 
+        if(message.indexOf('/give-control') === 0) {
+            this.promptForSelect(player, {
+                activePromptTitle: 'Select a character',
+                waitingPromptTitle: 'Waiting for opponent to give control',
+                cardCondition: card => card.location === 'play area' && card.controller === player,
```
cryogen: The location restriction can be lifted I feel, maybe anywhere but hand?

ystros: I'll add discard + dead as valid targets. Hand + draw deck seem invalid, probably don't want take control to work w/ plots, agendas, or faction cards.

ystros:
I think this now also fixes #317. Can't reproduce it on this branch but can on master. The combination of takeControl pushing the card onto cardsInPlay followed by Yoren calling moveCard to explicitly get it in the play area was causing the card to be in cardsInPlay twice. Hence the double dominance strength if it were standing.

PR #324, opened 2017-02-08
Implement Lannisport Guard

DukeTax:


server/game/cards/characters/05/lannisportguard.js
```diff
@@ -0,0 +1,19 @@
+const DrawCard = require('../../../drawcard.js');
+
+class LannisportGuard extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardEntersPlay: (e, card) => card === this
```
cryogen: You need to restrict this so that it doesnt trigger in setup

DukeTax:
I believe Littlefinger doesn't restrict in a similar way and I have put him in setup without issue before, although restricting the reaction to the marshalling phase is needed anyway, especially with all the Lanni jumper effects.

server/game/cards/characters/05/lannisportguard.js
```diff
@@ -0,0 +1,19 @@
+const DrawCard = require('../../../drawcard.js');
+
+class LannisportGuard extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardEntersPlay: (e, card) => card === this && this.game.currentPhase === 'marshal'
+            },
+            handler: () => {
+                player.drawCardsToHand(1);
```
ystros: `player` doesn't exist here, and the card ability is for each player to draw a card. So what you really want here is an each loop through each player (which will define a `player` variable for you to use), then draw the card inside that loop.
```javascript
_.each(this.game.getPlayers(), player => {
    player.drawCardsToHand(1);
});
```

You'll also need to pull in the Underscore object at the top of the file:
```
const _ = require('underscore');
```

ystros:
FYI, you **don't** need to check that it isn't the setup phase, the ability engine will do that for you. But the card text does explicitly say 'after you marshal' so good to check that phase is marshal phase.

DukeTax:
That makes sense. Thanks again for the feedback.

PR #332, opened 2017-02-09
Improve handling of military claim

ystros:
* Requires all characters be selected for claim in the same prompt.
* Since they're no longer chosen sequentially, it is no longer possible to select the same character multiple times for the same claim.
* Enforces that if there are not enough characters to fulfill claim then all characters must be selected.
* Cards that can leave play mid-kill (biggest example being Jory Cassel during 2 claim or Valar) no longer end up in the dead pile.
* Fixed a bug where if a select prompt rejected the 'Done' click, the UI and the actual selection would be out of sync.
* **Bonus:** the order you select characters is the order they will be killed. Useful for working around save order problems in the short term and naturally choosing the order things go in the dead pile.


server/game/player.js
```diff
@@ -757,7 +757,12 @@ class Player extends Spectator {
                 this.game.addMessage('{0} discards a duplicate to save {1}', this, character);
             }
         } else {
-            this.game.raiseEvent('onCharacterKilled', this, character, allowSave, () => {
+            this.game.raiseEvent('onCharacterKilled', this, character, allowSave, event => {
+                if(character.location !== 'play area') {
+                    event.cancel();
```
ystros: I did feel a bit icky about this but the alternative seems to be rewriting the entire kill + save system (again). That's on the menu as we move toward first player choosing the order of triggered abilities, but this seemed like a 'good enough' fix in the meantime.

cryogen:
Does this allow you to fulfil e.g. claim 2 when you only have 1 character in play? I can't really tell without testing it.  I'll merge it assuming that it does but please speak up if it doesnt :)

ystros:
Yes. If you have fewer characters than the claim amount, it requires you to select all characters you do have.

PR #336, opened 2017-02-10
User params and image_path

cedricfey:
Hello guys 
I wanted to play in french with throneteki and I notice I would just have to replace the assets.
So I took the jump and made this merge request. 
The purpose would be to play with your own scan, (and in my case in my own language).
So I created a params page and store a image_path in the db. So if you host some scans somewhere you can play with it. I think you can also use local file if you setup your browser correctly (or iccorectly).

This is my very first public merge request and my first steps in node so please tell me every thing I did wrong. Maybe I should have created a issue first. I don't know you tell me.

### DIFF
--- a/client/Application.jsx
+++ b/client/Application.jsx
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 import io from 'socket.io-client';
 
 import Login from './Login.jsx';
+import Params from './Params.jsx';
 import Logout from './Logout.jsx';
 import Register from './Register.jsx';
 import Lobby from './Lobby.jsx';
@@ -26,6 +27,7 @@ var notAuthedMenu = [
 ];
 
 var authedMenu = [
+    { name: 'Params', path: '/params' },
     { name: 'Logout', path: '/logout' }
 ];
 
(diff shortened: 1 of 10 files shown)


ystros:
I'll defer to @cryogen since he runs the show, but in general this seems like a reasonable (if not very user friendly) way to get translated card images in quickly.

I haven't pulled this locally to try it out but a couple of comments based solely on the code (I wouldn't bother putting in the work changing it yet if I were you until @cryogen weighs in):

1. 'Params' is not a very good name for this. I think either 'user settings' or 'account settings' would be more appropriate.
2. In general I think the project prefers to use camel case instead of snake case. The places that snake case are used currently are with data that was imported from ThronesDB. So `image_path` should probably be `imagePath` instead.

cryogen:
I meant to comment on this earlier. Until such time that we officially support translations I would prefer if things like this were done via some sort of client thing. A chrome add on or a grease monkey script or something

cedricfey:
Fair enough.
In the end I used **'Resource Override** and I hosted the pics.

PR #343, opened 2017-02-11
implement Crone of Vaes Dothrak

zacchiro:
[ reopening after inadvertent close due to branch reorganization on my fork ]

A few points worth noting:

I had to had a new game event "onCardDiscardedFromDraw", because apparently there isn't an event raised upon pillage and similar effects. I guess a better long-term solution for this situations would be a generic "onCardMove(source, target)" event, but I didn't feel like starting that journey. If I'm missing a better way to do this, please let me know.

I'm not sure how much automation should go into verifying the starting condition of this card. I've decided to be minimal for now, so for instance the player will be prompted to trigger Crone even if they have no standing Dothraki character. If you think that part should be automated, just let me know.

### DIFF
--- a/server/game/cards/characters/02/croneofvaesdothrak.js
+++ b/server/game/cards/characters/02/croneofvaesdothrak.js
@@ -3,14 +3,22 @@ const DrawCard = require('../../../drawcard.js');
 class CroneOfVaesDothrak extends DrawCard {
 
     canBeTriggered(event, player, card) {
-        return (card.getType() === 'character' && player !== this.controller)
+        var standingDothrakis = this.controller.cardsInPlay.filter(
+            card =>
+                card.getType() === 'character'
+                && !card.kneeled
+                && card.hasTrait('Dothraki'));
+
+        return (card.getType() === 'character'
+                && player !== this.controller
+                && standingDothrakis.length > 0);
     }
 
     setupCardAbilities(ability) {
         this.reaction({
             when: {
                 onCardDiscarded: (e, p, c) => this.canBeTriggered(e, p, c)
-            }
+            },
             handler: context => {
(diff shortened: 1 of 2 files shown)


ystros:
Per my comments in Discord, the bigger problem is that the onCardDiscarded event was only being fired when you dragged a card to the discard pile. Really it should be fired any time a discard happens.

Rather than introduce a specific event for having discarded from draw, I think it may be more reasonable to add an original location parameter to the event and check that. Otherwise we need 3 separate events - discarded from hand, discarded from draw, discarded from play. If you pull in this commit it should get rid of the need for the onCardDiscardedFromDraw event: https://github.com/ystros/throneteki/commit/b1610c3d687f9fd555048e4e7ee47a5095017398

(example usage w/ LoCR Cersei here: https://github.com/ystros/throneteki/commit/ebb6992860e10a48680c1e0ebfe17fb572c54aff)

Regarding automation, kneeling a Dothraki character is part of the cost for the reaction and generally we've been checking that the cost can be paid before prompting the user. We haven't been perfect about it, but it would probably be good to check if they have a standing Dothraki.

ystros:
Oops. I went a bit too far on the previous commit, broke some tests. I edited the commits in the previous message, check https://github.com/ystros/throneteki/commit/536c48c2f4e9be12e336c191f5556a0e3a5137b5 instead.

zacchiro:
@ystros: that generalization of onCardDiscarded is just perfect! It's indeed what I wanted to have, but didn't dare implementing it myself for fear or unexpected consequences :)
I've now cherry picked your commit in this pull request, reworked the base card implementation to rely on it, and also made the user prompt for using it conditional on the availability of standing Dothraki characters.

PR #352, opened 2017-02-12
implement Even-Handed Justice

zacchiro:
The implementation requires the user to *separately* choose standing characters controlled by each player. This is due to the fact that, AFAICT, there is no way to pass separate cardCondition functions to promptForSelect, when asking to select multiple cards. Without that we cannot enforce the fact that the 2 selected characters must belong to different players. Hence the awkward workflow. Let me know if there are better ways to do this right now.

Additionally, the message emitted after the event is completed shows card names, but they are not "hyperlinks" that allow hovering, as I didn't find a way to do that. If you can point me in the right direction, I'll be happy to improve that.

### DIFF
--- a/server/game/cards/events/03/evenhandedjustice.js
+++ b/server/game/cards/events/03/evenhandedjustice.js
@@ -1,0 +1,80 @@
+const _ = require('underscore');
+
+const DrawCard = require('../../../drawcard.js');
+
+class EvenHandedJustice extends DrawCard {
+
+    canPlay(player, card) {
+        if(player !== this.controller || this !== card) {
+            return false;
+        }
+
+        return (_.every(this.game.getPlayers(), this.hasStandingCharacters)
+                && super.canPlay());
+    }
+
+    hasStandingCharacters(player) {
+        return (player.cardsInPlay
+                .filter(card => card.getType() === 'character' && !card.kneeled)
+                .length > 0);
+    }
+
+    play(player) {
+        this.waitingPrompt = 'Waiting for opponent to use ' + this.name;
+
+        this.toKneel = [];  // characters to kneel (later)
+
+        // step 1: select opponent character
+        this.game.promptForSelect(player, {
+            activePromptTitle: 'Select a standing character controlled by your opponent',
+            waitingPromptTitle: this.waitingPrompt,
+            cardCondition: card =>
+                !card.kneeled
+                && card.getType() === 'character'
+                && card.controller !== this.controller,  // not event controller
+            onSelect: (player, card) => this.onFirstCardSelected(player, card),
+        });
+
+        return true;
+    }
+
+    onFirstCardSelected(player, card) {
+        this.toKneel.push(card);
+
+        // step 2: select event controller character
+        this.game.promptForSelect(player, {
+            activePromptTitle: 'Select a standing character of yours',
+            waitingPromptTitle: this.waitingPrompt,
+            cardCondition: card =>
+                !card.kneeled
+                && card.getType() === 'character'
+                && card.controller === this.controller,  // event controller
+            onSelect: (player, card) => this.onSecondCardSelected(player, card),
+        });
+
+        return true;
+    }
+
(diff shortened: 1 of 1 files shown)


server/game/cards/events/03/evenhandedjustice.js
```diff
@@ -0,0 +1,65 @@
...
+            onSelect: (player, card) => this.onCardsSelected(player, card),
+        });
+    }
+
+    onCardsSelected(player, card) {
+        this.toKneel.push(card);
+        _.each(this.toKneel, card => card.controller.kneelCard(card));
+
+        var kneeledCards = _.map(this.toKneel, card => card.name);
+        this.game.addMessage('{0} uses {1} to kneel {2}',
```
ystros: To make the card lists hyperlinked, take a look at how Wildfire builds up the `params` variable and then outputs its message (it's kind of annoying, we should make it easier at some point): https://github.com/cryogen/throneteki/blob/master/server/game/cards/plots/01/wildfireassault.js#L44

server/game/cards/events/03/evenhandedjustice.js
```diff
@@ -0,0 +1,65 @@
...
+            waitingPromptTitle: waitingPrompt,
+            cardCondition: card =>
+                !card.kneeled
+                && card.getType() === 'character'
+                && card.controller !== this.controller,
+            onSelect: (player, card) => this.toKneel.push(card),
+        });
+
+        // kneel controller character
+        this.game.promptForSelect(player, {
```
ystros: I think generally we've preferred generating the next prompt from inside the onSelect of the previous prompt. This will technically work, though. Deferring to @cryogen on this one.

server/game/cards/events/03/evenhandedjustice.js
```diff
@@ -0,0 +1,65 @@
...
+                !card.kneeled
+                && card.getType() === 'character'
+                && card.controller === this.controller,
+            onSelect: (player, card) => this.onCardsSelected(player, card),
+        });
+    }
+
+    onCardsSelected(player, card) {
+        this.toKneel.push(card);
+        _.each(this.toKneel, card => card.controller.kneelCard(card));
```
ystros: You should probably return without kneeling anything here unless the `toKneel` array contains cards for both players. Based on the way you've structured the rest of the card it might not be possible for that to happen, but if the card is restructured it might be (e.g. click 'Done' without selecting a card when I select my card, but select one for the opponent, thus kneeling only their card).

server/game/cards/events/03/evenhandedjustice.js
cryogen: Hmm, interesting.  Yeah I think we should stick with waiting for one thing to happen rather than queuing two at once.  Could cause interesting side effects if the prompt is cancelled for any reason

zacchiro: OK, I've now implemented this, chaining the two card selections.

server/game/cards/events/03/evenhandedjustice.js
zacchiro: Good point, better safe than sorry. I've implemented this too.

server/game/cards/events/03/evenhandedjustice.js
zacchiro: Gah, I'd rather not copy paste that in another card. If you don't mind accepting it as is (i.e., with just non-hoverable card names in the message), I prefer to wait for that "markup" feature to be available as a helper function from somewhere. If you really want this as a requirement to land the implementation of Even-Handed Justice, I guess I could do the helper refactoring myself, but no promises ;-)

ystros: As-is is fine with me, we're still going to have to go back and convert all the others when that helper/message feature is introduced.

zacchiro: OK, I'll then leave it as-is for now. I'll be happy to help out with the porting of this and other cards when the helper will be available.

PR #353, opened 2017-02-12
Implement Qhorin Halfhand

DukeTax:


server/game/cards/characters/04/qhorinhalfhand.js
```diff
@@ -0,0 +1,37 @@
...
+                    activePromptTitle: 'Select a character',
+                    waitingPromptTitle: 'Waiting for opponent to use ' + this.name,
+                    cardCondition: card => (
+                        card.location === 'play area' && 
+                        card.getType() === 'character' && 
+                        !card.isUnique() && 
+                        card.getStrength() < this.getStrength() &&
+                        card.controller !== this.controller),
+                    onSelect: (p, card) => {
+                        this.card.controller.killCharacter(this.card);
```
ystros: You don't want `this.card` here. You want to kill the card that was selected, which was passed as parameter `card`. Thus, it should be:

```javascript
card.controller.killCharacter(card)
```

server/game/cards/characters/04/qhorinhalfhand.js
```diff
@@ -0,0 +1,37 @@
...
+            when: {
+                afterChallenge: (event, challenge) => (
+                    challenge.challengeType === 'military' &&
+                    challenge.winner === this.controller &&
+                    challenge.isParticipating(this)
+                )
+            },
+            handler: () => {
+                this.game.promptForSelect(this.controller, {
+                    activePromptTitle: 'Select a character',
```
ystros: Maybe the prompt title should be 'Select a character to kill' instead? It can be figured out from context since you were just prompted to activate Qhorin, but it still can be confusing if you're not clear why you're selecting a character.

PR #367, opened 2017-02-14
Implement Gendry

DukeTax:
1. Checking if dominance is lost in the forced reaction this way would not work in solo test games. Cryogen mentioned having the event pass the loser. If someone could do that for me or point out to me how to, gladly. :)

2. I guess it is possible to choose to discard a power when Gendry doesn't have any to prevent the sacrifice. Is this important (enough)?

### DIFF
--- a/server/game/cards/characters/02/gendry.js
+++ b/server/game/cards/characters/02/gendry.js
@@ -8,7 +8,7 @@ class Gendry extends DrawCard {
             },
             handler: () => {
                 this.modifyPower(1),
-                this.game.addMessage('{0} uses {1} gain a power on {1}', this.controller, this);
+                this.game.addMessage('{0} uses {1} gain a power on {2}', this.controller, this, this);
             }
         });
 
--- a/server/game/cards/characters/02/gendry.js
+++ b/server/game/cards/characters/02/gendry.js
@@ -1,0 +1,58 @@
+const DrawCard = require('../../../drawcard.js');
+
+class Gendry extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onDominanceDetermined: (event, winner) => this.controller === winner
+            },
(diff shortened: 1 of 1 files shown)


server/game/cards/characters/02/gendry.js
```diff
@@ -0,0 +1,58 @@
...
+        });
+
+        this.forcedReaction({
+            when: {
+                onDominanceDetermined: (event, winner) => this.game.getOtherPlayer(this.controller) === winner
+            },
+            handler: () => {
+                this.game.promptWithMenu(this.controller, this, {
+                    activePrompt: {
+                        menuTitle: 'Discard 1 power from ' + this.name + ' or sacrifice a bastard?',
```
cryogen: Does this look ok in the UI? Seems a bit long

server/game/cards/characters/02/gendry.js
```diff
@@ -0,0 +1,58 @@
...
+            },
+            handler: () => {
+                this.modifyPower(1),
+                this.game.addMessage('{0} uses {1} gain a power on {2}', this.controller, this, this);
+            }
+        });
+
+        this.forcedReaction({
+            when: {
+                onDominanceDetermined: (event, winner) => this.game.getOtherPlayer(this.controller) === winner
```
cryogen: Change this to event, winner, loser, then check this.controller === loser

I will do the code to support this parameter as a separate task

DukeTax:
Ah, didn't understand your first comment at first, but it was referring to my update about addMessage. I just changed it because I saw it done like that on some other card and thought I'd conform, but I changed it back. I haven't gotten throneteki running locally yet (getting closer though) so I haven't tested the menu but I shortened it to be like The Hound; ie ask for one option (the least bad, generally) and if not do the other. If you want something else let me know.

PR #368, opened 2017-02-14
implement Shierak Qiya

zacchiro:


server/game/cards/events/04/shierakqiya.js
```diff
@@ -0,0 +1,48 @@
...
+            waitingPromptTitle: 'Waiting for opponent to use ' + this.name,
+            cardCondition: card =>
+                card.location === 'play area'
+                && this.game.currentChallenge.isParticipating(card),
+            onSelect: (p, card) => this.onCardSelected(p, card),
+        });
+    }
+
+    onCardSelected(player, card) {
+        player.kneelCard(player.faction);
```
cryogen: Note to self; I think this is better served with a specific method, something like player.kneelFactionCard

player.kneelCard will raise stand/kneel events that we dont necessarily want to do for faction cards.

PR #370, opened 2017-02-14
Adding ser lancer and fixing tower of the hand

petarpetkov:
Tower of the hand should return the selected controller character to his hand as well as kneel the location itself.

### DIFF
--- a/server/game/cards/characters/05/serlancellannister.js
+++ b/server/game/cards/characters/05/serlancellannister.js
@@ -3,18 +3,22 @@ const DrawCard = require('../../../drawcard.js');
 class SerLancelLannister extends DrawCard {
     setupCardAbilities(ability) {
         this.persistentEffect({
-            condition: () => this.getOtherLannisterLords().length == 1,
+            condition: () => !this.getSingleOtherLannisterLordOrLady(),
             match: this,
-            effect: ability.effects.dynamicStrenght(() => this.getOtherLannisterLords().first().getStrength())
+            effect: ability.effects.dynamicStrenght(() => this.getSingleOtherLannisterLordOrLady().getStrength())
         });
     }
 
-    getOtherLannisterLords() {
+    getSingleOtherLannisterLordOrLady() {
         var cards = this.controller.cardsInPlay.filter(card => {
             return card.getFaction() === "lannister" && (card.hasTrait('Lord') || card.hasTrait('Lady')) && card.getType() === 'character' && card !== this;
         });
 
-        return cards;
+        if (cards.length === 1) {
+            return cards.first();
+        }
+
+        return;
     }
 
 }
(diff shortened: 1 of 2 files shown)


petarpetkov:
http://thronesdb.com/card/05014

This is ser lancel.

server/game/cards/locations/03/towerofthehand.js
```diff
@@ -33,6 +33,8 @@ class TowerOfTheHand extends DrawCard {
 
     onOpponentCardSelected(player, card) {
         card.controller.returnCardToHand(card);
+		this.selectedCard.controller.returnCardToHand(this.selectedCard);
```
cryogen: Good catch.  There seems to be a spacing issue here though, can you check?  Also can you update the text below to say '{0} kneels {1}' rather than uses

server/game/cards/characters/05/serlancellannister.js
```diff
@@ -0,0 +1,25 @@
...
+class SerLancelLannister extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => this.getOtherLannisterLords().length == 1,
+            match: this,
+            effect: ability.effects.dynamicStrenght(() => this.getOtherLannisterLords().first().getStrength())
+        });
+    }
+
+    getOtherLannisterLords() {
```
cryogen: Can this method do the logic for checking that there is only one other, and if there is, return it, and if there is not, return undefined?

Also can its name be something like getSingleOtherLannisterLordOrLady ?

petarpetkov:
I have made the changes. Sorry this is the first time ever I touch javascript and I am not really sure of what goes where yet.

server/game/cards/characters/05/serlancellannister.js
```diff
@@ -0,0 +1,29 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SerLancelLannister extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => !this.getSingleOtherLannisterLordOrLady(),
+            match: this,
+            effect: ability.effects.dynamicStrenght(() => this.getSingleOtherLannisterLordOrLady().getStrength())
```
ystros: Typo - dynamicStrength.

server/game/cards/characters/05/serlancellannister.js
```diff
@@ -0,0 +1,29 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SerLancelLannister extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => !this.getSingleOtherLannisterLordOrLady(),
```
ystros: This condition seems wrong. Shouldn't it evaluate truthy when there *is* another single Lannister lord or lady? This will evaluate true when there isn't one.

petarpetkov: totally right I was thinking of another card when doing that.

ystros:
👍

cryogen:
Thanks!

PR #371, opened 2017-02-14
Implement Robb Stark [ATSK]

DukeTax:


server/game/cards/characters/04/robbstark.js
```diff
@@ -0,0 +1,58 @@
...
+
+class RobbStark extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => (
+                this.game.currentChallenge &&
+                this.game.currentChallenge.challengeType === 'military'
+            ),
+            match: this,
+            effect: ability.effects.modifyStrength(this.numberOfLoyalChars())
```
ystros: This will need to use the dynamic strength effect. Currently what this will do is set the strength modifier to be the number of loyal cards in play at the start of the game, which is 0.

```javascript
effect: ability.effects.dynamicStrength(() => this.numberOfLoyalChars())
```

This will recalculate the strength modifier every time characters enter / leave play and periodically.

server/game/cards/characters/04/robbstark.js
```diff
@@ -0,0 +1,58 @@
...
+        });
+
+        this.action({
+            title: 'Initiate ability',
```
cryogen: Can we change this to something like 'Stand and remove a character from the challenge' - that reflects what the ability is and isn't too long

server/game/cards/characters/04/robbstark.js
```diff
@@ -0,0 +1,58 @@
...
+            waitingPromptTitle: 'Waiting for opponent to use ' + this.name,
+            onSelect: (player, card) => this.onCardSelected(player, card)
+        });
+    }
+
+    onCardSelected(player, card) {
+        card.controller.standCard(card);
+        this.game.currentChallenge.removeFromChallenge(card);
+
+        this.game.addMessage('{0} uses {1} to stand {2} and remove it from the challenge', player, this, card);
```
cryogen: Minor nitpick, characters are people so 'them' fits better here than 'it' :)

server/game/cards/characters/04/robbstark.js
DukeTax: I asked you about this in the chat a couple of days ago, where I believed you echoed my sentiments. :P Personally I would love it if any action was titled 'Initiate ability' like here, whether it is on Nymeria Sand, Iron Fleet Scout or Selyse Baratheon. I don't like the labels used currently, mostly from the 'Kneel this card to " variety. They look too much like 'Select this card', they are usually incomplete because of space constraints (like here, no room to mention non-King constraint) and they're sometimes just plain wrong (Iron Fleet Scout when you're first player for example). Having them all named the same is nice from a templating perspective, it stands out from 'Select this card' and if you don't know what the ability does, the card is shown large since you're mouse-over at that point anyway. Your call though. :)

PR #392, opened 2017-02-16
document ESLint usage in README.md

zacchiro:
Note: the instructions on how to run eslint singles out the client/ and server/
dir. This is because we cannot simply tell people to run 'eslint .' from the
root dir, as some built bundles will make eslint barf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,5 +76,19 @@ NODE_ENV=production ./node_nodules/.bin/webpack -p --config webpack.config.produ
 NODE_ENV=production PORT=4000 node .
 ```
 
-###Build Status
+### Coding Guidelines
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,5 +76,19 @@ NODE_ENV=production ./node_nodules/.bin/webpack -p --config webpack.config.produ
...
-###Build Status
+### Coding Guidelines
+
+All JavaScript code included in Throneteki should pass (no errors, no warnings)
+linting by [ESLint](http://eslint.org/), according to the rules defined in
+`.eslintrc` at the root of this repo. To manually check that that is indeed the
+case install ESLint and run
+
+```
+eslint client/ server/
```
cryogen: may want to add test/ in there as well, but i'll do it after i merge this

PR #393, opened 2017-02-16
Minor UI change to PlayerStats

khyinn:
Just displaying images instead of text for PlayerStats


cryogen:
Interesting idea.  Can you include a screenshot to save me checking out the whole branch and seeing for myself?

khyinn:
playerstats

khyinn:
And small things added again
playerstats

khyinn:
playerstats

Trying to replace the First Player DIV with an image in PlayerStats. Not sur for the player-info part but the our-side part seems to work as expected.

ystros:
I like the look of the icons (as long as there is alt/title for accessibility) but not sure I like the icon to indicate first player. It seems a bit too subtle in comparison to the current marker, and takes up precious room next to user name (consider longer names).  Just my $0.02.

khyinn:
The image can easily be replaced by something better. For the precious room next to user name, i'm thinking about changing the left side and stats width. We also can reduce a little the username font size (something like 10px instead of 14px actually). This can save some space.

cryogen:
I think I'd like to see the text as well as an icon, alt/hover isn't very good ux on mobile, and until people get used to the icons it may be confusing at a glance

khyinn:
OK, reverting the FirstPlayerIndicator changes so you can add the other changes to the main branch. Oher icons (for gold, power, reserve and claim) are taken from the game so they can be added without text.

khyinn:
capture d ecran 26

Another thing i'm testing is adding a background to the site so we can play on a beautiful board.
In addition to this change, i'm currently testing some transparency for panels.

It's just here for comments and ideas before making a commit.

cryogen:
Thanks.  any further changes please open a new PR.  The background image idea is a good one, not sure about the actual image.  And I think I'll ultimately want a way for people to turn it off (or pick another one)

khyinn:
Sure, we need a way to change it or turn it off (a settings panel for users along with the way to change locale :))

PR #398, opened 2017-02-16
Fix Summer Harvest from granting too much gold

### LINKED ISSUE #261: Summer Harvest granting too much gold
Just had a Summer Harvest vs A Noble Cause with 2 roseroads out grant me 16 gold in marshalling. No idea what could have caused it; we were in the second plot stack already though.

ystros:
Previously, if you went through all of your plots and played Summer
Harvest a second time, it would retain its income value from the first
time in addition to how much it would normally gain this round. Now
instead of setting a gold modifier, it sets the base income value which
is overwritten each time. This prevents the previous bug and should
allow interactions with Varys' Riddle to be correct.

Fixes #261.

### DIFF
--- a/server/game/cards/plots/04/summerharvest.js
+++ b/server/game/cards/plots/04/summerharvest.js
@@ -9,7 +9,7 @@ class SummerHarvest extends PlotCard {
                     return true;
                 }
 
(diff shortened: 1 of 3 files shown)


cryogen:
If those tests dont scream "we need a test harness" i dont know what will :)

PR #399, opened 2017-02-17
Implement Wildling Bandit

DukeTax:


server/game/cards/characters/05/wildlingbandit.js
```diff
@@ -0,0 +1,18 @@
+const DrawCard = require('../../../drawcard.js');
+
+class WildlingBandit extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => (
+                this.game.currentChallenge &&
+                this.game.currentChallenge.isAttacking(this) &&
+                this.controller.gold < this.game.getOtherPlayer(this.controller).gold),
```
ystros: This condition will work for 2 player games, but will crash when there's only 1 player. Generally we've taken an approach where cards should not crash (though not necessarily trigger) in 1 player testing.

PR #400, opened 2017-02-17
Implement Myrcella Baratheon

DukeTax:


server/game/cards/characters/04/myrcellabaratheon.js
```diff
@@ -0,0 +1,36 @@
...
+        });
+
+        this.persistentEffect({
+            condition: () => !this.getKingsInPlay(),
+            match: this,
+            effect: ability.effects.addKeyword('Renown')
+        });
+    }
+
+    getKingsInPlay() {
```
ystros: This function is incorrect. `_.each` will loop through the cards, but does not return the list of filtered cards. Additionally, you're not using the returned cards, only checking whether there are any. Finally, using `=> { ... }` opens up a lambda function which you need to explicitly return a value from, whereas `=> ( ... )` will implicitly return the result of whatever expression is in parentheses.

I would suggest rewriting this to be `areNoKingsInPlay` that returns a boolean instead using either `_.any` or `_.all`. e.g.
```javascript
areNoKingsInPlay() {
    return !_.any(this.game.getPlayers(), player => {
        return player.cardsInPlay.any(card => card.getType() === 'character' && card.hasTrait('King'));
    });
}
```

(probably want to double check the logic + syntax there, just quickly typing this here and not somewhere that would reveal obvious errors)

DukeTax: Ugh, sorry about that mess. I'll take a good look at it later so I can make sure I understand what went wrong. Thanks as usual.

Issue #405, opened 2017-02-17
Styling of game list with long titles overflows


ystros:
![screen shot 2017-02-17 at 9 01 24 am](https://cloud.githubusercontent.com/assets/145077/23075024/c238b2f8-f4ef-11e6-8b40-a886a11553d3.png)

Might be better if it stretches some (and a limit on the title length should be added to prevent abuse)

PR #409, opened 2017-02-17
Limit game name to 140 characters, inspired by twitter char count (#405)

nbelza:
Added a 140 character limit for the game name with a character counter located to the top right of the text input.

screen shot 2017-02-17 at 4 01 32 pm
screen shot 2017-02-17 at 4 01 57 pm
screen shot 2017-02-17 at 4 04 45 pm
(As ystros suggested, we want to prevent full texts of War & Peace as game titles)

Also added CSS adjustments to keep the text within its container, wrapping accordingly
screen shot 2017-02-17 at 3 25 52 pm
screen shot 2017-02-17 at 3 27 56 pm

### DIFF
--- a/client/NewGame.jsx
+++ b/client/NewGame.jsx
@@ -28,7 +28,7 @@ class InnerNewGame extends React.Component {
     }
 
     onNameChange(event) {
-        this.setState({gameName: event.target.value});
+        this.setState({gameName: event.target.value.substr(0,140)});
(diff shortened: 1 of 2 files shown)


cryogen:
I'm not sure the character count is overly necessary, but I'm not sure it hurts anything either.  I will implement the limit server side, don't worry about that

PR #415, opened 2017-02-18
implement most of the missing challenge events

zacchiro:
The new class is called ChallengeEvent and by default checks whether we're
currently in a challenge or not. Additionally it can be passed arguments to
declaratively request checking that we're in a challenge of a given type and/or
participating in a given role (attacker/defender).

### DIFF
--- a/server/game/cards/challengeevent.js
+++ b/server/game/cards/challengeevent.js
@@ -1,0 +1,42 @@
+const DrawCard = require('../drawcard.js');
+
+class ChallengeEvent extends DrawCard {
+    // event card that can only be played during a challenge
+
+    constructor(owner, cardData, challengeType, participatingAs) {
+        // - challengeType (optional): make the event playable only on a
+        //   challenge of the given type
+        //
+        // - participatingAs (optional): make the event playable only when the
+        //   event controller is participating as either 'attacker' or
+        //   'defender'
+        //
+        super(owner, cardData);
+
+        this.challengeType = challengeType;
+        this.participatingAs = participatingAs;
+    }
+
+    canPlay(player, card) {
+        if(player !== this.controller || this !== card) {
+            return false;
+        }
+
+        var challenge = this.game.currentChallenge;
+        if(!challenge) {
+            return false;
+        }
+        if(this.challengeType && challenge.challengeType !== this.challengeType) {
+            return false;
+        }
+        if((this.participatingAs === 'attacker' && challenge.attackingPlayer !== this.controller) ||
+           (this.participatingAs === 'defender' && challenge.defendingPlayer !== this.controller)) {
+            return false;
+        }
+
+        return super.canPlay(player, card);
+    }
+
+}
+
+module.exports = ChallengeEvent;
--- a/server/game/cards/events/01/winteriscoming.js
+++ b/server/game/cards/events/01/winteriscoming.js
@@ -1,0 +1,18 @@
+const ChallengeEvent = require('../../challengeevent.js');
+
+class WinterIsComing extends ChallengeEvent {
+
+    // TODO implement restriction "(Max 1 per challenge.)"
+
+    play(player) {
+        this.untilEndOfChallenge(ability => ({
+            match: card => card === player.activePlot,
+            effect: ability.effects.modifyClaim(1)
+        }));
+    }
+
+}
+
+WinterIsComing.code = '01159';
+
+module.exports = WinterIsComing;
--- a/server/game/cards/events/04/thereismyclaim.js
+++ b/server/game/cards/events/04/thereismyclaim.js
@@ -1,0 +1,55 @@
+const _ = require('underscore');
+
+const ChallengeEvent = require('../../challengeevent.js');
+
+class ThereIsMyClaim extends ChallengeEvent {
+
+    // TODO implement restriction "(Max 1 per challenge.)"
+
+    isTyrellCharacterInHand(card) {
+        return card.location === 'hand'
+            && card.getType() === 'character' && card.getFaction() === 'tyrell';
+    }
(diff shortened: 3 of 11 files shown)


zacchiro:
WARNING: I've submitting this now to gather feedback on the general idea.
If you like it, I'll add to this PR the implementation of all the [relevant cards](https://thronesdb.com/find?q=x%3A%22play+only+during+a%22&sort=name&view=spoiler)

There seem to be 8 of them, and currently we only have Ours is the Fury in Throneteki. Implementing the other 7 with this helper should be quasi-trivial, and will serve as further validation for this idea.

What do you think?

ystros:
Not opposed to this approach, it should help clean up some of the repetitious logic. However, I do plan on redoing how cards are played / marshaled / ambushed / etc, so this code may end moving around or evaporating entirely.

I don't quite have 100% concrete thoughts on that yet, though, so I would say move forward with it unless there are other objections.

server/game/cards/challengeevent.js
```diff
@@ -0,0 +1,48 @@
...
+        //   challenge of the given type
+        //
+        // - partecipatingAs (optional): make the event playable only when the
+        //   event controller is partecipating as either 'attacker' or
+        //   'defender'
+        //
+        super(owner, cardData);
+
+        this._challengeType = challengeType;
+        this._partecipatingAs = partecipatingAs;
```
ystros: Spelling - `participatingAs`. Also unsure of the convention of prefixing these variables with underscore. I understand what you're going for here, but it's not something we've done elsewhere.

zacchiro: Fixed spelling and dropped _.

zacchiro:
Following @ystros and @cryogen comments, I went ahead with this idea.

This pull request now contains the implementations for most of the challenge event cards in the meta (with the exception of There Is My Claim and The Sword in the Darkness), all based on the new ChallengeEvent class.

As part of this I also introduced a new effect to (temporarily) set the max number of challenges of a given type to an arbitrary value (= 0 in my use case, but might be other for different needs).

Reviews and comments welcome.

server/game/cards/challengeevent.js
```diff
@@ -0,0 +1,46 @@
...
+        super(owner, cardData);
+
+        this.challengeType = challengeType;
+        this.participatingAs = participatingAs;
+    }
+
+    canPlay(player, card) {
+        if(player !== this.controller || this !== card) {
+            return false;
+        }
```
cryogen: this needs to go on the base card - that's more a note for myself than to you, though :)

server/game/cards/challengeevent.js
```diff
@@ -0,0 +1,46 @@
...
+
+        var challenge = this.game.currentChallenge;
+        if(!challenge) {
+            return false;
+        }
+        if(this.challengeType
+           && challenge.challengeType !== this.challengeType) {
+            return false;
+        }
+        if((this.participatingAs === 'attacker'
```
cryogen: I find this block quite hard to read -  I can tell you've tried to wrap to 80 chars, but it's ok to go a bit longer these days, as much as 160 even.  Perhaps putting the two bracketed conditionals on their own line might make this clearer

server/game/cards/events/01/winteriscoming.js
```diff
@@ -0,0 +1,18 @@
+const ChallengeEvent = require('../../challengeevent.js');
+
+class WinterIsComing extends ChallengeEvent {
+
+    // TODO implement restriction "(Max 1 per challenge.)"
```
cryogen: This probably means bringing events under the effects engine somehow, @ystros FYI

cryogen:
Other than a couple of minor things ive mentioned, looks good.  It's ok to go longer than 80 chars wide, too :)

zacchiro:
On Sun, Feb 19, 2017 at 01:22:51AM -0800, Stuart Walsh wrote:
> +        if(this.challengeType
> +           && challenge.challengeType !== this.challengeType) {
> +            return false;
> +        }
> +        if((this.participatingAs === 'attacker'
> 
> I find this block quite hard to read - I can tell you've tried to wrap
> to 80 chars, but it's ok to go a bit longer these days, as much as 160
> even.  Perhaps putting the two bracketed conditionals on their own
> line might make this clearer

More than a strict 80 char limit, I generally tend to not have multiple
logical clauses on the same line, because they make *further* changes
you make to those lines harder to read in diffs.  But here it doesn't
matter, because the left hand of the && here is just verifying the thing
is defined.  And of course I'll happily adapt to the code base style :)

Fixed in the pull request update I just pushed. Ditto for the other
"complex" boolean test in the same class.

Cheers.

zacchiro:
just added a15c1da23e1882e88ed9593b829d60f5b7307f24, which implements There Is My Claim

cryogen:
That reads much better, I'm ok splitting conditionals, but if they're in the same parens makes sense to keep them together

server/game/cards/events/04/thereismyclaim.js
```diff
@@ -0,0 +1,55 @@
+const _ = require('underscore');
+
+const ChallengeEvent = require('../../challengeevent.js');
+
+class WinterIsComing extends ChallengeEvent {
```
cryogen: copy and paste :)

zacchiro:
On Sun, Feb 19, 2017 at 10:05:43AM -0800, Stuart Walsh wrote:
> +class WinterIsComing extends ChallengeEvent {
> 
> copy and paste :)

Gah, I suck. It's interesting (and now that I looked into it fully
understandable) that it actually worked nonetheless :)

Fixed now. Sorry for forcing you through another review/accept cycle.

PR #426, opened 2017-02-19
Implement Arya Stark [WotN]

DukeTax:


server/game/cards/characters/03/aryastark.js
```diff
@@ -0,0 +1,36 @@
...
+                    activePromptTitle: 'Select a character',
+                    waitingPromptTitle: 'Waiting for opponent to use ' + this.name,
+                    onSelect: (player, card) => this.onCardSelected(player, card)
+                });
+            }
+        });
+    }
+
+    onCardSelected(player, card) {
+        this.controller.sacrificeCard(this);
```
ystros: This is relatively new (just merged today), but costs such as sacrificing the card should now be specified when you declare the reaction. See https://github.com/cryogen/throneteki/blob/master/docs/implementing-cards.md#paying-additional-costs-for-reactions-and-interrupts

So adding `cost: ability.costs.sacrificeSelf()` to the reaction declaration and removing this line will do the same thing.

server/game/cards/characters/03/aryastark.js
```diff
@@ -0,0 +1,36 @@
...
+                    activePromptTitle: 'Select a character',
+                    waitingPromptTitle: 'Waiting for opponent to use ' + this.name,
+                    onSelect: (player, card) => this.onCardSelected(player, card)
+                });
+            }
+        });
+    }
+
+    onCardSelected(player, card) {
+        card.controller.killCard(card);
```
ystros: Sorry, missed this in initial review - should be `killCharacter` not `killCard`.

Issue #430, opened 2017-02-20
Settings page for choosing a game background


khyinn:
Just a discussion to make it working before making a PR.
I've created a Settings page only when logged in. On this page, the user can select a background for playing the game on a beautiful board.
Here is an example
![capture d ecran 27](https://cloud.githubusercontent.com/assets/25751963/23108438/9de30f78-f70d-11e6-9966-51106ac3cd32.png)
I've also played with panels transparency to make it more polished.
I've made my Settings page with an input text where you type the name of the background to load.
![capture d ecran 28](https://cloud.githubusercontent.com/assets/25751963/23108471/1284faf8-f70e-11e6-87d2-46c154b100c9.png)
But, this is not really user-friendly and i think backgrounds preview buttons are better than this crap. I'm not very good with react and node so i can't figure it out. That's why i ask to you :)
Here is my Settings.jsx. The more important to keep is the user.settings.userBackground wich i'm using for the rest of the code.

```
import React from 'react';
import _ from 'underscore';
import $ from 'jquery';
import {connect} from 'react-redux';

import Link from './Link.jsx';

import * as actions from './actions';

class InnerSettings extends React.Component {
    constructor() {
        super();

        this.state = {
            error: '',
            userBackground: user.settings.userBackground,
            validation: {}
        };

        this.onChange = this.onChange.bind(this);
        this.onSave = this.onSave.bind(this);
    }

    onChange(field, event) {
        var newState = {};

        newState[field] = event.target.value;
        this.setState(newState);
        console.log(newState);
    }

    onSave(event) {
        event.preventDefault();
        this.setState({ error: '' });

        $.ajax({
            url: '/api/account/settings',
            type: 'POST',
            data: JSON.stringify({ userBackground: this.state.userBackground }),
            contentType: 'application/json'
        }).done((data) => {
            if(!data.success) {
                this.setState({ error: data.message });
                return;
            }else{user.settings.userBackground = this.state.userBackground}
        }).fail((xhr) => {
            console.log(xhr);
            if(xhr.status === 401) {
                this.setState({ error: 'something wrong' });
            } else {
                this.setState({ error: 'Could not communicate with the server.  Please try again later.' });
            }
        });

    }

    render() {
        var fields = [
            {
                name: 'userBackground',
                label: 'Game Background',
                placeholder: 'my/background/path',
                inputType: 'text'
            }
        ];

        var fieldsToRender = [];
[...]

PR #431, opened 2017-02-20
Implementing Settings page and user backgrounds

khyinn:
Solving #430


khyinn:
Argh, something is wrong... But it's here just for cryogen to look at and make the changes needed...

khyinn:
Travis CI build fails saying that user isn't known but it's working... Maybe i missed something in GameBoard.jsx.

cryogen:
There is now a settings page, backgrounds are likely to be something in the future once we actually get some that can be used

PR #433, opened 2017-02-20
Implement Stannis Baratheon [TIMC] (partial)

DukeTax:
The effect of the reaction (character that cannot stand) is not enforced, but displaying the message still makes it clear you have triggered him and who was targeted.


cryogen:
I'm not now sure how I feel about the effect not being implemented but being printed out.  It might lead to confusion from people think it's broken.  I will ponder on this some more

PR #434, opened 2017-02-20
Implement Salladhor Saan

DukeTax:


server/game/cards/locations/04/salladhorsaan.js
```diff
@@ -0,0 +1,34 @@
...
+                this.game.promptForSelect(this.controller, {
+                    activePromptTitle: 'Select a card to put into play',
+                    waitingPromptTitle: 'Waiting for opponent to use ' + this.name,
+                    cardCondition: card => (
+                        card.location === 'hand' &&
+                        card.controller === this.controller &&
+                        ((card.hasTrait('Warship') && card.getType() === 'location') ||
+                        (card.hasTrait('Weapon') && card.getType() === 'attachment'))),
+                    onSelect: (player, card) => {
+                        player.moveCard(card, 'play area');
```
ystros: This will work for the locations but not for attachments. Attachments require you choose something to attach the card on, which is done from `Player.promptForAttachment`.

ystros: Edit: You can probably use `player.playCard` instead using the 'forcePlay' parameter. i.e.
```javascript
player.playCard(card, true)
```

That should do the full 'put into play' action, including prompting for attachment.

PR #435, opened 2017-02-20
Implement Ser Kevan Lannister

DukeTax:


server/game/cards/characters/05/serkevanlannister.js
```diff
@@ -0,0 +1,32 @@
...
+                this.game.promptForSelect(this.controller, {
+                    activePromptTitle: 'Select location or attachment',
+                    waitingPromptTitle: 'Waiting for opponent to use ' + this.name,
+                    cardCondition: card => (
+                        card.location === 'discard pile' &&
+                        card.controller === this.controller &&
+                        (card.getFaction() === this.getFaction() || card.getFaction() === 'neutral') &&
+                        (card.getType() === 'location' || card.getType() === 'attachment')),
+                    onSelect: (player, card) => {
+                        player.moveCard(card, 'play area');
```
ystros: Same problem as Salladhor Saan, won't work with attachments. `player.playCard(card, true)` should resolve it though.

PR #441, opened 2017-02-21
Catch serialization errors while handling errors.

ystros:


cryogen:
I tried this before and it never trips.  Doesn't hurt to have it though

PR #444, opened 2017-02-21
Implement Red Keep Spy

DukeTax:


server/game/cards/characters/05/redkeepspy.js
```diff
@@ -0,0 +1,44 @@
...
+                }
+            },
+            handler: () => {
+                var otherPlayer = this.game.getOtherPlayer(this.controller);
+
+                if(!otherPlayer) {
+                    return;
+                }
+
+                if(this.controller.hand.size() <= otherPlayer.hand.size()) {
```
cryogen: I would argue that this check should be part of the event, so that it doesn't trigger if this condition is not true

server/game/cards/characters/05/redkeepspy.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+    
+    hasMoreCardsInHand() {
+        var otherPlayer = this.game.getOtherPlayer(this.controller);
+
+        if(!otherPlayer) {
+            return false;
+        }
+
+        if(this.controller.hand.size() <= otherPlayer.hand.size()) {
```
cryogen: I'll merge it as-is, but just FYI you could write this is as return this.controller.hand.size() > otherPlayer.hand.size()

DukeTax:
Thanks, makes sense. :)

PR #453, opened 2017-02-22
implement Riverrun

zacchiro:


server/game/cards/locations/04/riverrun.js
```diff
@@ -0,0 +1,35 @@
...
+                        return true;
+                    }
+                    return false;
+                }
+            },
+            handler: () => {
+                if(!this.tullyCharacter) {
+                    return false;
+                }
+                this.controller.kneelCard(this);
```
ystros: Use the new `cost` property instead (and then you don't need to check !this.kneeled in the event). https://github.com/cryogen/throneteki/blob/master/docs/implementing-cards.md#paying-additional-costs-for-reactions-and-interrupts

server/game/cards/locations/04/riverrun.js
```diff
@@ -0,0 +1,35 @@
...
+                this.tullyCharacter.modifyPower(1);
+                this.tullyCharacter = undefined;
+
+                this.game.addMessage('{0} uses {1} to have {2} gain 1 power',
+                                     this.controller, this, this.tullyCharacter);
```
ystros: You set this.tullyCharacter to undefined in the line before this. ;-)

zacchiro: Gah, silly me (and nice catch!). Fixed.

server/game/cards/locations/04/riverrun.js
zacchiro: The new cost framework is very very very cool. Thanks for pointing me to it.

Fixed this PR so that it uses it.

PR #456, opened 2017-02-22
Implement White Raven

DukeTax:


server/game/cards/characters/04/whiteraven.js
```diff
@@ -0,0 +1,38 @@
...
+                    (this.anyPlotHasTrait('Summer') || this.anyPlotHasTrait('Winter')))
+            },
+            handler: () => {
+                this.game.addPower(this.controller, 1);
+                this.game.addMessage('{0} uses {1} to gain 1 power for their faction', this.controller, this);
+            }
+        });
+    }
+
+    anyPlotHasTrait(trait) {
```
cryogen: We should consider adding this to Game, if it could be useful for any other cards

PR #458, opened 2017-02-22
Update chat to log pillaged card for /pillage and onPillage.  

nbelza:
Also update /discard x phrasing, depending on number of cards discarded.


cryogen:
You need to update the tests, too :)

PR #462, opened 2017-02-23
port cycle #1 characters that kneel/stand to the cost framework

zacchiro:


cryogen:
Would Selmy's standing really be considered a cost?  I get why you've done it but I'm not sure it fits.  @ystros ?

ystros:
Card text says "stand Ser Barristan Selmy to save that character", which is the standard rule construction for "[pay cost] to [do thing]". So yes, in this instance standing is a cost. For ability cancellers, Selmy will stand no matter what, just the character won't be saved.
https://thronesdb.com/card/02107

cryogen:
Fair enough.  I suppose I could just have read the comment on thronesdb for the card! :+1:

zacchiro:
On Thu, Feb 23, 2017 at 12:27:50AM -0800, Brian Upton wrote:
> Card text says "stand Ser Barristan Selmy to save that character",
> which is the standard rule construction for "[pay cost] to [do
> thing]". So yes, in this instance standing is a cost.

Yup, that was my line of thinking.

Thanks @cryogen for the review, and for the merge! :)

PR #465, opened 2017-02-23
implement Ser Edmure Tully

zacchiro:
This is an implementation of Edmure. But there is a BUT (that is also noted as
a TODO in the card code). It will trigger even when power is *moved* between
cards, rather than only when it is *gained*. To achieve the proper semantics
however we need to:

- introduce a new event (!= onCardPowerChanged) that is raised when power is
  moved between cards

- add a new function to move power, so that cards that need to do so have a
  high-level way of implementing it, rather than using two distinct power
  modifications on two cards

- review all current uses of power modifications to check whether they should
  be ported to the new high level effect

Do we want to do the above as a condition for having Edmure implemented, or do
we prefer to have a suboptimal implementation of Edmure in right now, and
improve it later?

### DIFF
--- a/server/game/cards/characters/04/seredmuretully.js
+++ b/server/game/cards/characters/04/seredmuretully.js
@@ -1,0 +1,59 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SerEdmureTully extends DrawCard {
+
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                // TODO this should trigger only when power is *gained*, but currently also
+                // triggers when power is *moved* between cards. To make the distinction we
+                // need a new high-level event and review all uses of card power modifications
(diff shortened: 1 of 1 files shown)


DukeTax:
Even though it has the potential to cause issues (people pissed something that is illegal being 'allowed'  by the system) I'd still say go for it, since the vast majority of cards gain power, not get it moved to them.

On another note; renown is optional in this game, which means that when a player is facing an Edmure and he has 1 character with renown, he probably doesn't use renown anyway and stuff has to be reverted.

cryogen:
Implementing this card means I'm probably going to have to make renown a trigger :(

zacchiro:
@cryogen @cavnak yeah, we definitely need to make renown optional.
But TBH I think we needed to do so anyway, so here's our chance :)

PR #471, opened 2017-02-24
Implement framework for cost reducing effects.

ystros:
Introduces the CostReducer class that represents an active cost
reduction effect. These reducer objects can be added to the player
object, which are then used to calculate reductions on matching cards.

This change does not replace the current cost calculations yet.

### DIFF
--- a/server/game/costreducer.js
+++ b/server/game/costreducer.js
@@ -1,0 +1,42 @@
+const _ = require('underscore');
+
+class CostReducer {
+    constructor(game, source, properties) {
+        this.game = game;
+        this.source = source;
+        this.uses = 0;
+        this.limit = properties.limit;
+        this.match = properties.match;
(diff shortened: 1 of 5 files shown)


ystros:
The next PR in this sequence (once this and #470 are merged) will convert the existing reducer cards to use these effects and get rid of the old `canReduce` and `reduce` methods. To give an idea of usage, Kingsroad would be applied via:

```javascript
    targetType: 'player',
    targetController: 'current',
    effect: ability.effects.reduceNextMarshalledCardCost(3, card => card.getType() === 'character')
}));
```

Things that have a persistent effect such as Paxter Redwyne look more like this:
```javascript
this.persistentEffect({
    targetType: 'player',
    targetController: 'current',
    effect: ability.effects.reduceFirstPlayedCardCostEachRound(1, card => card.getType() === 'event')
});
```

PR #474, opened 2017-02-24
Implement Sweet Donnel Hill

DukeTax:
Quite sure this should be able to be done more eloquently. :P

### DIFF
--- a/server/game/cards/characters/05/sweetdonnelhill.js
+++ b/server/game/cards/characters/05/sweetdonnelhill.js
@@ -1,0 +1,25 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SweetDonnelHill extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => this.game.currentChallenge && this.game.currentChallenge.isDefending(this),
+            match: (card) => this.game.currentChallenge.isAttacking(card) && card.getType() === 'character',
+            targetController: 'any',
+            effect: [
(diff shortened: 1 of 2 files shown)


cryogen:
Ideally this needs to be done as a single effect.  Also ideally it would just work on the list of current known keywords, which we shouldn't have to duplicate in too many places.

Pretty sure @ystros will agree, but feel free to say otherwise :)

DukeTax:
Yep. Also ability.effects.removeKeyword() doesn't exist. :)

ystros:
I would recommend creating a new `removeAllKeywords` effect (in addition to the `removeKeyword` effect that's missing). You could literally implement it like you have here for now, it'll just be shoved in a more appropriate place:
```javascript
removeAllKeywords: function() {
    return [
        this.removeKeyword('Ambush'),
        this.removeKeyword('Insight'),
        this.removeKeyword('Intimidate'),
        this.removeKeyword('Pillage'),
        this.removeKeyword('Renown'),
        this.removeKeyword('Stealth'),
        this.removeKeyword('Terminal'),
        this.removeKeyword('Limited')
    ]
}
```

Ideally you can pull out the ValidKeywords constant in basecard.js into a commonly accessible file and map that here. Not sure it's worth doing quite yet (and think there's an argument for creating a richer Keyword class to factor out some of the behaviour associated with keywords, instead of just having a list of string names)

cryogen:
I'm 99.5% sure that someone will forget to update both places when there's a new keyword added :)

(I'm not 98% sure that it wont happen because i've just said that)

PR #480, opened 2017-02-24
port core set locations that kneels to the cost framework

zacchiro:
with the exception of reducers, as they're coming in a separate PR by ystros

Note: I've assumed here that the cost framework inhibits weird things like
kneeling a location that is not in the play area, which seems to be an
assumption we already make for other types cards, e.g., characters. As such,
I've removed a bunch of checks on the card being in the play area as part of
this PR. If this assumption is not sound, please shout :-)


server/game/cards/locations/01/castleblack.js
```diff
@@ -1,15 +1,16 @@
...
         this.action({
             title: 'Kneel this card to stand and give a defending character +2 STR',
+            cost: ability.costs.kneelSelf(),
             method: 'kneel'
         });
     }
 
     kneel(player) {
-        if(this.location !== 'play area' || !this.game.currentChallenge || this.kneeled) {
+        if(this.location !== 'play area' || !this.game.currentChallenge) {
```
ystros: When using the cost framework, these types of `if`s should be converted to the `condition` property on action instead. Otherwise, the cost will be paid before it discovers that the ability can't be used. https://github.com/cryogen/throneteki/blob/master/docs/implementing-cards.md#checking-ability-restrictions

So in this case:
```javascript
this.action({
    title: 'Kneel this card to stand and give a defending character +2 STR',
    condition: () => this.game.currentChallenge,
    method: 'kneel'
});
```

(See Wildling Horde, it's similar)

server/game/cards/locations/01/highgarden.js
```diff
@@ -1,15 +1,16 @@
...
         this.action({
             title: 'Kneel this card',
+            cost: ability.costs.kneelSelf(),
             method: 'kneel'
         });
     }
 
     kneel(player) {
-        if(this.kneeled || this.controller.gold < 1 || this.location !== 'play area' || !this.game.currentChallenge) {
+        if(this.controller.gold < 1 || !this.game.currentChallenge) {
```
ystros: Similar to above, should be a `condition` on the action.

server/game/cards/locations/01/ironfleetscout.js
```diff
@@ -1,15 +1,16 @@
...
         this.action({
             title: 'Kneel this card to give a character +1 STR',
+            cost: ability.costs.kneelSelf(),
             method: 'kneel'
         });
     }
 
     kneel(player) {
-        if(this.location !== 'play area' || !this.game.currentChallenge || this.kneeled) {
+        if(!this.game.currentChallenge) {
```
ystros: Ditto

server/game/cards/locations/01/castleblack.js
zacchiro: thanks @ystros ! all instances adapted as you suggested, and i'll take care of this in future porting work too

PR #485, opened 2017-02-25
fix Forgotten Plans condition check to verify current phase on players

zacchiro:
The .phase attribute is not defined on Game objects, but as the condition was
negated in this plot, it always succeeded. The proper way to do this is
checking .phase on Player objects.

We now check it on both player at once, which is OK as phases are always
updated simultaneously.


zacchiro:
On Sat, Feb 25, 2017 at 09:08:33PM -0800, Brian Upton wrote:
> A simpler change would be to use game.currentPhase (which does exist
> whereas game.phase does not).

Ah, yes, *much* simpler :-) Thanks for noticing, fixed.

PR #490, opened 2017-02-26
dedup 'waiting for opponent to use' messages via new 'source' property

zacchiro:
The new property can be passed to various user prompt helpers and will be used
to create a default waitingPromptTitle when one is not explicitly given.


cryogen:
Happy to merge this but be advised I've merged in new card implementations that will also need to be converted

zacchiro:
On Sun, Feb 26, 2017 at 05:04:44AM -0800, Stuart Walsh wrote:
> Happy to merge this but be advised I've merged in new card implementations
> that will also need to be converted

Sure, just done this in #491.

PR #493, opened 2017-02-26
Implement Chataya's Brothel

DukeTax:


server/game/cards/locations/04/chatayasbrothel.js
```diff
@@ -0,0 +1,33 @@
+const DrawCard = require('../../../drawcard.js');
+
+class ChatayasBrothel extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kneel a character to gain gold',
+            phase: 'marshal',
+            limit: ability.limit.perPhase(2),
+            handler: context => {
+                this.game.promptForSelect(context.player, {
```
ystros: The character kneel is a cost for the ability, and should be specified using the `cost` property. Should be as simple as:
```javascript
cost: ability.costs.kneel(card => card.hasIcon('intrigue')),
```

(the framework should check the location and controller for you, if it doesn't then that's a bug with the framework)

PR #494, opened 2017-02-26
implement Renly Baratheon (For Family Honor)

zacchiro:
To that end, raise a new onInsight event, carrying the list of cards drawn
using the Insight keyword.

Passing note: I find super annoying that the "drawn" parameter of the new event
is either a singleton element (a card) or an array of cards. But I've done it
that way for consistency with what we do for onPillage/discardFromDraw. IMHO
it'd be nice to uniform all of those to always return an array of cards.

### DIFF
--- a/server/game/cards/characters/04/renlybaratheon.js
+++ b/server/game/cards/characters/04/renlybaratheon.js
@@ -1,0 +1,29 @@
+const DrawCard = require('../../../drawcard.js');
+
+class RenlyBaratheon extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onInsight: (event, challenge, card, drawn) => {
+                    this.drawnCard = drawn;
+
+                    // postpone the check about drawn card loyalty, to avoid
+                    // leaking game state to the opponent
+                    return true;
+                }
+            },
+            handler: () => {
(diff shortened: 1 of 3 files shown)


cryogen:
This potentially leaks game state to your opponent.  You'll only get the prompt if you draw a loyal card so your opponent can guess that youve drawn a loyal card if they see "waiting for opponent..."

zacchiro:
On Sun, Feb 26, 2017 at 12:43:51PM -0800, Stuart Walsh wrote:
> This potentially leaks game state to your opponent.  You'll only get
> the prompt if you draw a loyal card so your opponent can guess that
> youve drawn a loyal card if they see "waiting for opponent..."

Good point. So, how do you want to go about avoiding this?

The first idea that comes to mind is *always* prompt the user, meaning
that if the controller chose to trigger when they've drawn a non-loyal
card, nothing would happen.  This would be more annoying in terms of
user experience than the current implementation, but it doesn't leak
info --- opponent cannot distinguish the situation in which you've drawn
a loyal card and decided not to trigger Renly, and the case you've drawn
a non loyal card.

Other ideas?

cryogen:
What you said is all I could think of..

zacchiro:
OK, adapted to do so.

PR #504, opened 2017-02-27
Implement The Valyrian

DukeTax:


server/game/cards/locations/04/thevalyrian.js
```diff
@@ -0,0 +1,31 @@
+const DrawCard = require('../../../drawcard.js');
+
+class TheValyrian extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kneel to give attacking character +STR',
+            condition: this.game.currentChallenge && this.game.currentChallenge.defendingPlayer.gold >= 1 && this.game.currentChallenge.attackers.length >= 1,
+            cost: ability.costs.kneelSelf,
```
ystros: kneelSelf is a function, should be:
```
cost: ability.costs.kneelSelf()
```

PR #505, opened 2017-02-27
Implement Renly Baratheon (TtB)

ystros:


ystros:
Let me know if you think the character test is too much. I don't most cards needs such a test, or even this one specifically TBH, but it was a good opportunity to help flesh out the test framework a bit more.

zacchiro:
FWIW, I think *every* card should have a test for its specification. It's the only way to make sure we can in the long term make significant refactorings with confidence we aren't breaking the world with them.
If the tests becomes too long to write or too onerous to maintain, we should look at ways of simplifying them (e.g., by factoring out deck archetypes, or common play scenarios), rather than avoid adding them.
YMMV

PR #507, opened 2017-02-27
port attachments that kneel to the cost framework

zacchiro:


server/game/cards/attachments/02/mareinheat.js
```diff
@@ -17,7 +18,7 @@ class MareInHeat extends DrawCard {
     }
 
     kneel(player) {
-        if(this.kneeled || this.location !== 'play area' || !this.game.currentChallenge) {
+        if(!this.game.currentChallenge) {
```
ystros: This should be a `condition` of the action.

server/game/cards/attachments/04/motherofdragons.js
```diff
@@ -15,13 +16,12 @@ class MotherOfDragons extends DrawCard {
             return false;
         }
 
-        if(this.kneeled || !player.cardsInPlay.any(card => {
-            return challenge.isParticipating(card) && card.hasTrait('Dragon');
-        })) {
+        if(!player.cardsInPlay.any(
+            card => challenge.isParticipating(card) && card.hasTrait('Dragon')
```
ystros: This (and the if checking that there is a challenge above it) should be converted to a `condition` for the action.

server/game/cards/attachments/01/heartsbane.js
```diff
@@ -1,19 +1,19 @@
...
-            title: 'Kneel this card to give attached card +3 STR',
+            title: 'Kneel this card to give attached character +3 STR',
+            cost: ability.costs.kneelSelf(),
             method: 'kneel'
         });
     }
 
     kneel(player) {
-        if(!this.parent || this.kneeled || !this.game.currentChallenge || !this.game.currentChallenge.isParticipating(this.parent)) {
+        if(!this.parent || !this.game.currentChallenge || !this.game.currentChallenge.isParticipating(this.parent)) {
```
ystros: This should be converted to a `condition` for the action.

zacchiro:
On Mon, Feb 27, 2017 at 01:10:57PM -0800, Brian Upton wrote:
> ystros commented on this pull request.

Thanks, just sent in take 2.

About this:

> -        if(!this.parent || this.kneeled || !this.game.currentChallenge || !this.game.currentChallenge.isParticipating(this.parent)) {
> +        if(!this.parent || !this.game.currentChallenge || !this.game.currentChallenge.isParticipating(this.parent)) {

is checking this.parent on attachments whose abilities get triggered
still needed? (for now I left it in)

ystros:
Unfortunately, I think the parent checks are still necessary. I've tried to take some of them out but its apparently lead to weird issues if the attachment happens to trigger right as its parent is leaving play.

PR #509, opened 2017-02-27
Implement Ser Davos Seaworth [GoH]

DukeTax:
Hope this is correct.


server/game/gamesteps/challenge/choosestealthtargets.js
```diff
@@ -28,6 +28,8 @@ class ChooseStealthTargets extends BaseStep {
 
         this.game.addMessage('{0} has chosen {1} as the stealth target for {2}', this.challenge.attackingPlayer, target, character);
 
+        this.game.raiseEvent('onStealth', this.challenge, character, target);
```
ystros: I'd call the event `onBypassedByStealth` instead, but the arguments and usage look good.

PR #513, opened 2017-02-28
Implement Silent Sisters

DukeTax:


silentsisters.js
```diff
@@ -0,0 +1,14 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SilentSisters extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: this,
+            effect: ability.effects.dynamicStrength(() => this.controller.deadPile.length)
```
ystros: `deadPile` unfortunately is not an array but is an Underscore object that wraps the array. So you'll need to use `size()` here instead of `length`. Otherwise, 👍

PR #515, opened 2017-02-28
Port Harrenhal to cost framework and fix put into play effects

DukeTax:


server/game/cards/locations/04/harrenhalffh.js
```diff
@@ -1,38 +1,36 @@
...
 class Harrenhal extends DrawCard {
-    setupCardAbilities() {
+    setupCardAbilities(ability) {
         this.action({
             title: 'Kneel Harrenhal to put a character in play',
-            method: 'kneel',
-            phase: 'challenge'
+            cost: ability.costs.kneelSelf(),
+            phase: 'challenge',
+            handler: context => {
```
cryogen: I'm not super thrilled about how much this has increased the indentation vs separate methods

ystros: I think in this case it's because the onSelect for the prompt hasn't been separated into its own method. Otherwise, this is literally the same level of indentation as reactions / interrupts.

I plan on pulling most (thought probably not 100%) of the select prompts into another action/reaction/interrupt property, which will further collapse things like this, but have been busy working on other things.

cryogen: Yeah it's mainly the pulling stuff out of methods I was commenting on.  Not sure it adds anything but does make readability somewhat worse

DukeTax: I thought it reads better like this but what do I know. You want separate methods always, or possibly only above a certain amount of lines?

ystros: The intention was to have `handler` to replace `method`, since it is far more consistent with the way reactions and interrupts are done. It can also open the door to prefab handlers later (e.g general 'draw card', 'gain gold', 'kill selected character', etc).

If we do start converting them back to `method`, keep in mind that the signature is currently different between the two - handler takes a context, while the method version takes player, arg, then context. The context object **is** necessary as it provides information about the costs paid (which can sometimes involve player choices). This difference in signatures was done to ensure backward compatibility while things were updated to use `handler` instead.

If that's not the direction we're going, a concerted effort should be made to have method just take a context instead (aka update every single usage in a single PR).

cryogen: I'm fine with handler replacing method, just if the callbacks to promptWithSelect are more than a couple of lines they should be in their own method is all i was saying

PR #516, opened 2017-02-28
Implemented Arya's Gift and fixed Nymeria 

nbelza:
When attachments are moved from one character to another, they now keep their duplicates rather than discarding them.

### DIFF
--- a/server/game/cards/attachments/02/lady.js
+++ b/server/game/cards/attachments/02/lady.js
@@ -1,47 +1,60 @@
 const DrawCard = require('../../../drawcard.js');
 
 class Lady extends DrawCard {
+    constructor(owner, cardData) {
+        super(owner, cardData);
+        this.oldOwner = null;
+    }
+
     setupCardAbilities(ability) {
         this.whileAttached({
             effect: ability.effects.modifyStrength(2)
         });
         this.action({
             title: 'Pay 1 gold to attach Lady to another character',
-            method: 'reAttach'
+            method: 'reAttach',
+            limit: ability.limit.perPhase(1)
         });
     }
 
     reAttach(player) {
-        if(!this.parent || this.controller.gold < 1) {
+        this.oldOwner = this.parent;
+        if(!this.oldOwner || this.controller.gold < 1) {
             return false;
         }
 
-        player.removeAttachment(this, false);
-        player.promptForAttachment(this);
+        player.moveCard(this, 'play area');
 
-        player.gold -= 1;
-
-        this.triggered = true;
+        this.game.promptForSelect(this.controller, {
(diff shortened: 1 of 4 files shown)


nbelza:
screen shot 2017-02-28 at 4 52 05 pm
screen shot 2017-02-28 at 4 52 19 pm
screen shot 2017-02-28 at 4 52 30 pm
screen shot 2017-02-28 at 4 53 16 pm
screen shot 2017-02-28 at 4 54 14 pm
screen shot 2017-02-28 at 4 54 22 pm
screen shot 2017-02-28 at 4 54 27 pm

server/game/cards/attachments/02/lady.js
```diff
@@ -1,47 +1,60 @@
 const DrawCard = require('../../../drawcard.js');
 
 class Lady extends DrawCard {
+    constructor(owner, cardData) {
+        super(owner, cardData);
+        this.oldOwner = null;
```
ystros: This initialization (and the one on Nymeria) isn't strictly necessary. It does let the reader of the code know to expect this variable later but doesn't really fit with other places we've used similar patterns.

server/game/cards/attachments/02/lady.js
```diff
@@ -1,47 +1,60 @@
...
-            if(card.name === 'Sansa Stark' && card.kneeled) {
-                player.standCard(card);
-            }
-
-            this.triggered = false;
+    moveAttachment(player, newOwner) {
+        var targetPlayer = this.game.getPlayerByName(newOwner.controller.name);
+        targetPlayer.attach(player, this, newOwner.uuid);
+        player.gold -= 1;
+        if(newOwner.name === 'Sansa Stark' && newOwner.kneeled) {
```
ystros: You should set `this.oldOwner` to null here (and also in Nymeria). 

Scenario:
I use Lady to reattach to a different character (oldOwner set to original character).
Opponent uses something to remove Lady (goes back to your hand or you fish it out of discard later).
I try to play Lady again on the original character. I can't because of the added check in `canAttach` says I can't attach it to the old owner (which is still set).

ystros:
On the one hand, I think have a `moveAttachment` method on player would have been simpler. On the other, I like the UI mechanism of moving the attachment into the play area. Makes it more obvious what's going on to the opponent.

server/game/cards/attachments/02/lady.js
nbelza: Fixed!

server/game/cards/attachments/02/lady.js
nbelza: Fixed!

PR #517, opened 2017-03-01
Introduce a 'put into play' method

ystros:
* Replaces usages of `playCard(card, true)` a.k.a force-play with a new `putIntoPlay` method.
* I'll fix usages introduced in #515 once it's merged (force play will continue to work even with this change, putIntoPlay is just simpler / reads better)
* `putIntoPlay` will ultimately be used by the forthcoming marshaling refactor


zacchiro:
@ystros: #518 (Ghosts of Harrenhal) which I just submitted also uses `playCard(card, true)`. Let me know if you want me to rebase it on top of this (#517) PR.

DukeTax:
Would it be possible to have the onCardEntersPlay event raised by putIntoPlay pass the location the card came from? Useful for this for example: https://thronesdb.com/card/06011.

ystros:
@zacchiro - Don't worry about it, it can be converted after merge.
@DukeTax  - Yeah, we'll probably have to add that, but I would wait until the PR that actually implements that card.

PR #518, opened 2017-03-01
implement Ghosts of Harrenhal

zacchiro:


ystros:
Looks good but may cause some arguments in the short term when played after things like Valar or Wildfire. Currently you can't choose the order characters killed by those multi-kill plots, they're pretty much killed in the order they originally entered play. 2+ military claim does let you choose the order though (implicitly, by the order you selected characters).

zacchiro:
On Wed, Mar 01, 2017 at 12:24:37PM -0800, Brian Upton wrote:
> Looks good but may cause some arguments in the short term when played
> after things like Valar or Wildfire. Currently you can't choose the
> order characters killed by those multi-kill plots, they're pretty much
> killed in the order they originally entered play. 2+ military claim
> does let you choose the order though (implicitly, by the order you
> selected characters).

So, yes, I thought about that too.  The reason why I decided to
implement it nonetheless, is that I don't see us having an anti-argument
solution for this specific problem anytime soon (or ever, TBH).

It is currently *possible* to reorder things in the dead pile, but
that's annoying, as you've to move cards back and forth between the in
play are and that pile. We can improve over that by adding some sort of
drag & drop based reordering within the windows used to show cards,
which will apply to all piles (discard, dead, draw deck, plots, etc).
But this will not solve the arguments problem, as people will *always*
forget to do so when multiple characters die.  The only way to make
players to think about the ordering, is to *force* them to choose the
ordering when multiple characters side, but that would be annoying in
99% of the games, i.e., every time Ghosts of Harrenhal is not played.
I don't think we want to add that annoyance just for the 1% of the
remaining games.

Hence my personal conclusion that having Ghosts of Harrenhal implemented
now is better than not having it. YMMV :)

cryogen:
When this goes in(now) we need a more explicit way of choosing which characters die first.  I'm not a fan of having the dead pile able to be reordered - I'd rather stuff went in in the right order - that or they "fix" this card themselves

PR #519, opened 2017-03-01
Implement The Tumblestone

DukeTax:


server/game/cards/locations/06/thetumblestone.js
```diff
@@ -0,0 +1,28 @@
...
+                    if(!card.hasTrait('House Tully') || card.getType() !== 'character' || power === 0 || this.tokens['gold'] === 0) {
+                        return false;
+                    }
+                    
+                    this.standCard = card;
+                    return true;
+                }
+            },
+            handler: () => {
+                this.removeToken('gold', 1);
```
ystros: This should be a cost, though we don't have a 'removeToken' cost currently. if you're not comfortable writing a new type of cost, that's fine, add a TODO comment here so we don't forget about it.

ystros:
Note to @cryogen - Test failure here will be remedied once #522 is merged.

PR #522, opened 2017-03-01
Add "discard 1 gold"-cost to the cost framework

DukeTax:
Don't see discarding more than 1 gold as a cost or discarding different types of tokens as a cost happening soon, so I kept it simple.

Edit: Hmm, probably better to do it right the first time, have it take type and amount arguments...

### DIFF
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -95,6 +95,20 @@ const Costs = {
                 context.source.controller.standCard(context.source);
             }
         };
+    },
+    /**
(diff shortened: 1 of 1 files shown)


ystros:
This could be made more generic, but like you said, there doesn't seem to be a reason to yet - it's mostly just discarding 1 gold from the card. 👍

DukeTax:
On third thought, I'll keep it like this for now. Easy enough to fix later if it ever becomes an issue.

PR #525, opened 2017-03-02
Implement /bestow command.

ystros:


cryogen:
This is arguably the same as /token gold x but i guess it's useful to have a separate command like /power

PR #534, opened 2017-03-04
Fix Cersei LoCR + implement Tywin LoCR

### LINKED ISSUE #184: Cersei (LoCR) triggers for each card discarded
Cersei should only trigger once when discard happens, not per card. So if claim 2 or playing Without His Beard for example, it's prompting to trigger for each card discarded.

ystros:
* Ensures Cersei LoCR only gains power once for each 'group' of cards discarded. Fixes #184.
* Implements Tywin LoCR.

💥 💥 💥  **Please review carefully** 💥 💥 💥 

Fixing Cersei was the easier part. In order for Tywin to be implemented, two things needed to be changed.

1. Discarding cards is now asynchronous. Tywin's effect needs to be able to replace the card being discarded, and so must stop the flow of the normal discard until the choice has been made by his player. Unfortunately, this meant introducing callback methods on `discardFromDraw` and `discardAtRandom`, which previously assumed that the discard was done synchronously.
2. The parameters for these events needed to be consolidated into an object. The reasoning for this was two-fold. First and most importantly, since Tywin replaces the cards that get discarded, it was necessary to be able to change the card being discarded for subsequent handlers in the event chain. The easiest way to do this was to put it as a property on an object that could be updated. Second, the event was already at 4 parameters, which gets really confusing when defining event handlers. This breaks the pattern used by every other event we've defined so far (though there may be other events that can benefit from this approach).

I'm not terribly happy with any of this but unsure of any alternate approaches and It Works™️. Open to suggestions if there's a nicer path here.


server/game/cards/characters/02/croneofvaesdothrak.js
```diff
@@ -5,16 +5,16 @@ class CroneOfVaesDothrak extends DrawCard {
     setupCardAbilities(ability) {
         this.reaction({
             when: {
-                onCardDiscarded: (event, player, card, allowSave, originalLocation) =>
-                    ((originalLocation === 'hand' || originalLocation === 'draw deck')
-                     && card.getType() === 'character'
-                     && player !== this.controller)
+                onCardDiscarded: (event, params) =>
```
cryogen: I find this really hard to read.  params doesnt really mean anything, and is confused with the params a few lines down.

I don't know what the solution is, though.  Maybe One Day(tm) the event itself should just hold the relevant properties in a context-specific manner

ystros: I agree, it'd be way better if it were all just on the event object (which is where I wanted to move ultimately anyway). Let me take a crack at that here, though it'll probably require a new variant of `raiseEvent`.

ystros: How about something more like this? https://github.com/ystros/throneteki/compare/fix-cersei-finally-i-guess-sigh...ystros:merged-events?expand=1

**Pros:**
* Gets rid of the secondary `params` object by just merging it directly into the event.
* No longer need to access event.params directly in the handlers, so that can clean quite a bit up.
* Still fulfills all the requirements introduced by Tywin (replaceable event params).

**Cons:**
* We have two different ways of raising events now. I could make it do the merge if the first parameter has a special flag on it (e.g. `raiseEvent('onCardDiscarded', { merge: true, card: card })`), or just by detecting that it's a single non-class object parameter, but that feels a bit too magical. Thought it would be better to be explicit (in the short term while there's two ways, at least).
* Converting all events over (which should be done rather than maintain both forever) will be super painful. I'm not even sure there's a good migration path - just a catastrophic 'convert it all' commit for each event.

(those cons still existed with the `params` argument, it was just less explicit)

cryogen:
this has conflicts now, sorry

ystros:
Resolved conflicts.

PR #540, opened 2017-03-05
Implement King Robb's Host

DukeTax:


server/game/cards/characters/03/kingrobbshost.js
```diff
@@ -0,0 +1,53 @@
...
+        this.game.addPower(loser, -power);
+        card.modifyPower(power);
+
+        this.game.addMessage('{0} uses {1} to move {2} power from {3}\'s faction card to {4}', 
+                              this.controller, this, power, loser, card);
+                        
+        return true;
+    }
+
+    anyPlotHasTrait(trait) {
```
cryogen: Feels like this should be a method on game, in case anything else wants to do a similar check (of which I'm sure there are a few)

PR #546, opened 2017-03-06
Fix King Robb and convert to new syntax

### LINKED ISSUE #544: Robb Stark (Across the Seven Kingdoms) is prompted to remove characters in non-military challenges
Robb's action says:
Action: While Robb Stark is participating in a Military challenge, choose a participating non-King character. Stand and remove that character from the challenge. (Limit once per challenge.)

My opponent was able to kneel, defend, and trigger Robb Stark to stand Brienne of Tarth during a Power challenge which I initiated with Brienne of Tarth.

ystros:
Previously, King Robb's action was able to be used during any challenge
he participated in. Now it will only properly trigger for military
challenges.

Fixes #544.

### DIFF
--- a/server/game/cards/characters/04/robbstark.js
+++ b/server/game/cards/characters/04/robbstark.js
@@ -14,25 +14,34 @@ class RobbStark extends DrawCard {
         this.action({
             title: 'Stand and remove a character from the challenge',
             method: 'initiate',
-            limit: ability.limit.perChallenge(1)
+            limit: ability.limit.perChallenge(1),
+            condition: () => this.isParticipatingInMilitaryChallenge(),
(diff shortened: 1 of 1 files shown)


server/game/cards/characters/04/robbstark.js
```diff
@@ -14,25 +14,34 @@ class RobbStark extends DrawCard {
         this.action({
             title: 'Stand and remove a character from the challenge',
             method: 'initiate',
```
cryogen: I think you need to remove this as well?

ystros: :woman_facepalming: Yup, fixed.

PR #549, opened 2017-03-06
Implement Storm's End

DukeTax:


server/game/cards/locations/06/stormsend.js
```diff
@@ -0,0 +1,34 @@
...
+                    source: this,
+                    cardCondition: card => card.getType() === 'character',
+                    onSelect: (player, cards) => this.onSelect(player, cards)
+                });
+            }
+        });
+    }
+    
+    onSelect(player, cards) {
+        cards[0].modifyPower(1);
```
cryogen: I know you "know" there's only 2, but i still think a _.each would be beneficial here

cryogen:
This needs to be rebased

PR #550, opened 2017-03-06
Implement Bestow keyword prompt

ystros:
* Adds a prompt when a Bestow card enters play asking the player to select the amount they want to place. If the player doesn't have the full potential bestow amount, it will only prompt up to however much gold they do have. 
screen shot 2017-03-06 at 1 28 57 pm
* Minor fix to The Tumblestone to not prompt when the character gaining power is already standing.


cryogen:
:+1:  just need rebasing

ystros:
Conflicts resolved.

PR #552, opened 2017-03-07
port remaining cards that use playCard(card, true) to putIntoPlay

zacchiro:
Clean up after #517


ystros:
I'm surprised this isn't conflicted because #545 already did this.

zacchiro:
That's indeed ... weird. Given the timing I can only conclude it was a race condition, and I submitted it (and GitHub checked for conflicts) before yours was merged. Anyway, this PR is pointless now, closing.

PR #553, opened 2017-03-07
make Melisdanre trigger on R'hllor events

### LINKED ISSUE #416: Melisandre doesn't trigger on events
Melisandre doesn't trigger when you play Seen in Flames. Her code references "onCardEntersPlay" and I guess events never enter play.

zacchiro:
Fixes #416

### DIFF
--- a/server/game/cards/characters/01/melisandre.js
+++ b/server/game/cards/characters/01/melisandre.js
@@ -4,7 +4,10 @@ class Melisandre extends DrawCard {
     setupCardAbilities(ability) {
         this.reaction({
             when: {
-                onCardEntersPlay: (e, card) => card.controller === this.controller && card.hasTrait('R\'hllor')
+                onCardEntersPlay: (event, card) =>
+                    card.controller === this.controller && card.hasTrait('R\'hllor'),
+                onCardPlayed: (event, player, card) =>
+                    card.controller === this.controller && card.hasTrait('R\'hllor')
             },
             limit: ability.limit.perRound(1),
             handler: () => {


server/game/cards/characters/01/melisandre.js
```diff
@@ -4,7 +4,8 @@ class Melisandre extends DrawCard {
     setupCardAbilities(ability) {
         this.reaction({
             when: {
-                onCardEntersPlay: (e, card) => card.controller === this.controller && card.hasTrait('R\'hllor')
+                onCardPlayed: (event, player, card) =>
```
ystros: Switching to this event will break Mel for any instance where a R'hllor card is put into play, as this event is only fired when the player actively plays the card from hand.

In general I think this event should be moved to be fired *only* when an event is played, as it makes the semantics weird right now (implies that marshaling / ambushing is "playing" which does not match the language of the rules / cards.). Then you can have handlers for both events.

ystros: Oops - Mel is only supposed to work on marshaling anyway (which the implementation doesn't currently enforce). My criticism of the onCardPlayed event still stands, though - it doesn't match the language of the rules.

zacchiro:
On Tue, Mar 07, 2017 at 05:12:50PM -0800, Brian Upton wrote:
> Oops - Mel is only supposed to work on marshaling anyway (which the
> implementation doesn't currently enforce). My criticism of the
> onCardPlayed event still stands, though - it doesn't match the
> language of the rules.

I'm not sure I get how you suggest we proceed here.

For the first part, you're saying that onCardPlayed will trigger even if
the card hasn't been marshaled (e.g., by card abilities), right? If so,
I guess we need to better separate events, or is there any other way
you're thinking of?

For the onCardPlayed, are you suggesting we rename it? If so then, any
naming suggestion?

ystros:
Sorry. I was saying that onCardPlayed is currently being triggered for the following situations:
* Event card is played.
* Card is marshaled
* Card is ambushed

This would have been a problem if Mel should fire with put-into-play effects as I initially thought, but she should only trigger upon marshal, so your change would have worked fine. However, since it's called onCardPlayed, and the rules don't really use the word 'played' when marshaling / ambushing, the event should really only trigger for the following situation:
* Event card is played

The only usage of the onCardPlayed event currently is Great Kraken, which seems to have been a throwback from before 'onCardEntersPlay' existed. I've removed that usage in #556.

My suggestion is the following:
* Do not fire onCardPlayed at all when marshaling / ambushing a card.
* Continue to fire onCardPlayed when an event-type card is played (basically, move the raiseEvent from `Game.playCard` and into `Player.playCard` where event cards are handled).
* Mel listens to both `onCardEntersPlay` (to handle marshaling of R'hllor cards) and `onCardPlayed` (to just handle events). Catelyn Stark WotN is another example of an ability listening on two events.

As a separate issue (which doesn't necessarily need to be handled on this PR but it's kind of trivial) is that Mel's current implementation is allowing her to trigger when a R'hllor card enters play outside of marshaling. The example that came up in lobby chat was Ghosts of Harrenhal bringing a Fiery Follower back into play, which inappropriately triggered Mel. Just need to check the phase in the `onCardEntersPlay` handler (e.g. Northern Rookery).

I don't know. It's probably fine to merge this as is and fix other things separately. I just noticed the huge potential problem with the semantics of the onCardPlayed event. It "works" in this particular case, but could cause problems with any other "when X card is played" abilities which expect **only** event cards down the line if not corrected.

zacchiro:
On Wed, Mar 08, 2017 at 01:18:14AM -0800, Brian Upton wrote:
> I don't know. It's probably fine to merge this as is and fix other
> things separately.

Thanks for clarifying all this, @ystros.

I'm totally fine with putting this on hold for now. The implementation
is buggy-ish right now in master, and it will be only marginally better
if we merge this as-is. I'll try to give it a stab as a more general
proper fix, along the lines you suggested --- although probably as
separate PRs, we'll see.  I think merging this could wait for that to
happen.

@cryogen: feel free to tag this PR as "temporarily on hold", if you have
one such tag :-)

ystros:
With #573 merged, `onCardPlayed` is only fired for event cards now. So it should be able to proceed with this PR by keeping the existing `onCardEntersPlay` handler and adding a separate handler for `onCardPlayed`.

zacchiro:
Thanks for noticing @ystros. PR updated, and tested that it works. This should now be good to go.

PR #567, opened 2017-03-10
Implement Jojen Reed

DukeTax:


server/game/cards/characters/04/jojenreed.js
```diff
@@ -0,0 +1,61 @@
...
+            },
+            handler: () => {
+                var otherPlayer = this.game.getOtherPlayer(this.controller);
+
+                if(!otherPlayer) {
+                    return;
+                }
+
+                var ownCard = this.controller.drawDeck.value()[0];
+                var opponentCard = otherPlayer.drawDeck.value()[0];
```
ystros: These will both read better using `first()`. e.g.
```javascript
var opponentCard = otherPlayer.drawDeck.first()
```

PR #572, opened 2017-03-11
log details of moved cards when they move from/to public places

zacchiro:
Currently the list of public places is hard coded in the Game.drop() method, as
they are just strings throughout the code. If/when they become abstract
objects (e.g., Pile, Places, or something such), we should make sure they have
isPublic() properties to make this and similar checks more clean.

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -288,7 +288,18 @@ class Game extends EventEmitter {
         }
 
         if(player.drop(cardId, source, target)) {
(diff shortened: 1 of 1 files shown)


zacchiro:
It just occurred to me that 'play area' should also be in the list of public places for this PR, but *only* if the setup phase is done. If not, we will leak information if players manually fiddle with setup cards before setup is over. Any suggestion about how to best check this?

PR #580, opened 2017-03-12
Implement Bear Island Host

DukeTax:


ystros:
Test failure is unrelated and a problem on master, so 👍

PR #583, opened 2017-03-13
Clear games when worker is restarted or times out

ystros:
Previously, if a game server worker timed out, the games would remain in
the lobby view even though they were no longer running / accessible.
Additionally, if a worker was restarted, any games that were on that
worker would also remain stuck in the lobby.

### DIFF
--- a/server/gamerouter.js
+++ b/server/gamerouter.js
@@ -106,6 +106,7 @@ class GameRouter extends EventEmitter {
         var message = JSON.parse(msg.toString());
         switch(message.command) {
             case 'HELLO':
+                this.emit('onWorkerStarted', identityStr);
                 this.workers[identityStr] = {
                     identity: identityStr,
                     maxGames: message.arg.maxGames,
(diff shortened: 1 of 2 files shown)


ystros:
@cryogen Let me know if you think this is coupling things too much between the router and the lobby. The lobby already knows a little about workers (sets the `node` property on `game`), so there's some small coupling there already, but this is ever so slightly more. I do think this cleanup is necessary but maybe there's a better way?

I also feel like there's some cleanup that needs to be done for any sockets that were connected to games that were on the worker, but it's unclear what's appropriate in these situations nor does there seem to be an easy way to get ahold of those sockets.

cryogen:
No this is pretty much the direction I was going in - this was to be my next port of call after getting it live, that and getting nodes to resync after they reconnect - nodes should be able to operate independently of the main server.  This took a backseat because the zmq version on the live server is too old to properly support what I need for this to work

PR #586, opened 2017-03-13
Implement Vaes Tolorro

DukeTax:


server/game/cards/locations/04/vaestolorro.js
```diff
@@ -0,0 +1,35 @@
...
+                    
+                    return true;
+                }
+            },
+            cost: ability.costs.kneelSelf(),
+            handler: () => {
+                if(this.pendingCard.getPower() >= 2) {
+                    var power = this.pendingCard.getStrength() === 0 ? 2 : 1;
+                } else {
+                    power = 1;
```
ystros: Not sure how @cryogen feels style-wise but I'm generally not fond of this construct when a variable is defined in one block and then used in another sibling-level block. It's technically valid Javascript but doesn't read well. I'd suggest either declare the variable above the `if` block with initial value 1 (which gets rid of the need for the `else`) or using `Math.min` to one-line it.

cryogen: Agreed

PR #587, opened 2017-03-13
Implement Tower of the Sun

DukeTax:


server/game/cards/locations/04/towerofthesun.js
```diff
@@ -0,0 +1,20 @@
+const DrawCard = require('../../../drawcard.js');
+
+class TowerOfTheSun extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onCardPlayed: (event, player, card) => card.getType() === 'event'
```
ystros: The `getType()` check here will be redundant once #573 is merged but doesn't hurt anything and is necessary right now. So 👍

DukeTax: I know, #573 was the reason I did this one. The reason I checked getType() (embarassingly) is that ESLint bugs me if I don't do anything with card. What's the better solution for that? (thanks as usual)

ystros: Once it's merged, you could use this instead:
```javascript
onCardPlayed: () => true
```

However, I'd recommend leaving things like you have them now. #573 is a bigger chunk of code, so there's no guarantee it will be merged before this PR. If this PR were merged and #573 wasn't, Tower of the Sun would activate for every card marshaled in addition to events being played, which would be annoying.

PR #591, opened 2017-03-14
Add /count-dominance command

DukeTax:
This prints current dominance strength for both players.

Was requested by mplain (thronesdb admin) a while back and thought it was a good suggestion.


zacchiro:
Can you sort the list of commands in About.jsx alphabetically? I did that a while back — and still thinks it's very useful, given the amount of commands we now have — but it looks like it's slowly slipping away again.
Other than that, thumbs up on this new command!

cryogen:
I don't disagree that this is a useful suggestion, but I'm not sure it fits as a command.

PR #592, opened 2017-03-14
Apply Wraiths in their Midst to Rains plots.

### LINKED ISSUE #590: Wraiths in their midst doesn't reduce the reserve on a plot revealed with Rains

ystros:
Previously, Wraiths in their Midst was not being applied to plots
revealed by The Rains of Castamere. This has been resolved by converting
it into a persistent effect.

Fixes #590.


cryogen:
We still have cards that haven't been converted? scary

PR #603, opened 2017-03-15
Always remove player from pending game when left

ystros:
Previously, if a player left a game, the game listing in the lobby would
continue to list that player in the game.


ystros:
Regresses crash when you try to spectate your own game after leaving, need to rework.

PR #618, opened 2017-03-17
Implement Motley

DukeTax:


server/game/cards/attachments/03/motley.js
```diff
@@ -0,0 +1,41 @@
...
+            },
+            handler: () => {
+                this.parent.controller.discardAtRandom(1);
+                this.game.addMessage('{0} uses {1} to discard 1 card at random from {2}\'s hand', 
+                                      this.controller, this, this.parent.controller);
+            }
+        });
+        this.reaction({
+            when: {
+                onDefendersDeclared: (event, challenge) => challenge.isDefending(this.parent)
```
ystros: These two don't need to be defined as two separate reactions. You can define it as one reaction and have the `when` contain both event. See Catelyn Stark WotN for an example.

PR #623, opened 2017-03-18
Implement Jon Snow [WotW]

DukeTax:


server/game/cards/characters/07/jonsnow.js
```diff
@@ -0,0 +1,36 @@
...
+                    this.game.addMessage('{0} uses {1} to stand each attacking Wildling character', this.controller, this);
+                },
+                'Stand Nights Watch': () => {
+                    this.controller.cardsInPlay.each(card => {
+                        if(this.game.currentChallenge.isDefending(card) && card.isFaction('thenightswatch') && card.getType() === 'character') {
+                            card.controller.standCard(card);
+                        }
+                    });
+
+                    this.game.addMessage('{0} uses {1} to stand each defending The Night\'s Watch character', this.controller, this);
```
cryogen: if you use {2} instead of "The Night's Watch" and then put in an extra argument of "thenightswatch" this will display the icon instead of the text

PR #625, opened 2017-03-18
improve Growing Strong log message to highlight affected cards

zacchiro:


zacchiro:
Note: while this change seems relatively safe — and consistent with what other cards do, e.g., Wildfire Assault — I couldn't actually test this locally before submitting (because rebuilding client code is currently broken for me). So testing would be appreciated :)

zacchiro:
Tests with this work locally, and CircleCI doesn't show the log of what/why failed. I suspect it was a transient issue in the CI environment. @cryogen: can you kick the build again?

PR #628, opened 2017-03-19
Implement Maester Aemon [WotW]

DukeTax:
This can probably be done a lot cleaner, but it works so 'good enough'? (style or other comments welcome ofcourse!)


cryogen:
Not sure what @ystros thinks, but I think this needs to be hooked into the core engine somehow, rather than trying to do claim itself

ystros:
I agree. The approach I'd take is extract a much as possible of `ChallengeFlow.applyClaim` (basically everything inside the event but not the event itself) into its own game step class, then have Aemon queue that with an object that has all the properties of a challenge.

ystros:
#729 will allow us to apply claim directly in this case - just create an appropriate looking challenge object (`challengeType`, `winner`, `loser` and `claim` must be set) and queue the `ApplyClaim` step.

PR #632, opened 2017-03-19
sort user list in the main lobby

zacchiro:


zacchiro:
user list is now sorted server side, so this PR is no longer needed

PR #638, opened 2017-03-19
add automatic parsing of deck title/faction/agenda to the deck builder

### LINKED ISSUE #633: single text box entry for the deckbuilder
Currently the deckbuilder requires user to specify deck name / faction / agenda in dedicated text boxex, and everything else (draw deck / plot deck) in a separate text box that is parsed on the fly. That makes it annoying inserting deck in textual format, as they usually have everything in a txt file.

It would be better to parse everything from a single big text box (as ThronesDB allows to do, for what is worth), removing the need of manually entering name/faction/agenda.

Note that this is separate from #358, as independently from that we will always need a way to manually enter/edit decks.

zacchiro:
Complete ThronesDB-like textual decks can now be pasted into the "Cards"
section of the deck builder. If a deck "header" --- made of deck title,
faction, agenda, possibly with multiple banners in the case of "Alliance" ---
is detected, it will be used to pre-populate the values of the various deck
builder dropdown selectors.

Fixes #633

### DIFF
--- a/client/DeckEditor.jsx
+++ b/client/DeckEditor.jsx
@@ -194,6 +194,42 @@ class DeckEditor extends React.Component {
     onCardListChange(event) {
         var split = event.target.value.split('\n');
 
+        var headerMark = _.findIndex(split, line => line.match(/^Packs:/));
+        if(headerMark >= 0) { // ThronesDB-style deck header found
+                              // extract deck title, faction, agenda, and banners
+            var header = _.filter(_.first(split, headerMark), line => line !== '');
+            split = _.rest(split, headerMark);
+
+            if(header.length >= 2) {
+                this.setState({ deckName: header[0] });
+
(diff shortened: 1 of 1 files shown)


zacchiro:
Couple of caveats:

- the synchronization between the text box and the dropdown menus is currently one way: changes in the text box will update the dropdown, but not viceversa. Make it bidirectional is feasible, but a bit annoying to code. Alternatively, as a third option, we can simply remove the deck header from the text box as soon as a change is made via the dropdown menus. IMO this issue can just be ignored, given we already don't round trip with what the user pasted when we edit an already inserted deck.

- the logic of the various setState() calls is duplicated wrt other code paths. It's just a few lines of code, but still. @cryogen has pending refactoring of the entire deck builder, so maybe this issue could be postponed/folded into that.

PR #665, opened 2017-03-22
Implement Tommen Baratheon

DukeTax:


DukeTax:
Thoughts on adding a message in case a challenge is unopposed but the bonus is not gained? Normally we don't have messages for constant abilities (except Red Keep I guess, need to remove that one), but this is such a rare situation that it might throw people off when it happens without a message.

cryogen:
I think it should have a message.

Deferring review of the actual implementation to @ystros

server/game/cards/characters/05/tommenbaratheon.js
```diff
@@ -0,0 +1,32 @@
...
+class TommenBaratheon extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => this.controller.hand.size() === 0,
+            targetType: 'player',
+            effect: ability.effects.cannotGainChallengeBonus()
+        });
+
+        this.persistentEffect({
+            condition: () => this.opponentHasNoCardsInHand(),
```
ystros: The following should be equivalent to your two separate effects and will get rid of the extra function:
```javascript
this.persistentEffect({
  targetType: 'player',
  targetController: 'any',
  match: player => player.hand.size() === 0,
  effect: ability.effects.cannotGainChallengeBonus()
});
```

server/game/effects.js
```diff
@@ -408,6 +408,16 @@ const Effects = {
             }
         };
     },
+    cannotGainChallengeBonus: function() {
+        return {
+            apply: function(player) {
+                player.gainsChallengeBonus = false;
```
ystros: Since 'cannot' is absolute, it's better for the flag to be named w/ the word cannot, set to `true` by the effect and be defaulted to `false` on the card (or otherwise just never set). When it isn't set up that way, it allows the potential for another effect to set `gainsChallengeBonus` back to `true`, thus negating the 'cannot' which the rules do not allow. See the other 'cannot' effects.

ystros:
Re: messages, it should be relatively easy to add a message in `ChallengeFlow` when they cannot gain bonus power. If you want a message to the effect that 'Tommen prevents player from gaining bonus power', that's going to be super tricky to work into the effects engine since it's timing base. I have (fairly deprioritized plans) to add effect message capabilities when an effect starts, but haven't given much thought to having it output messages on certain events (or non-events in this case).

PR #666, opened 2017-03-22
Implement The New Gift

DukeTax:
The challenges action is working, but the marshalling action is not for some reason...


ystros:
Marshaling action should start working properly when #667 is merged.

PR #674, opened 2017-03-24
Implement Southron Messenger

DukeTax:


server/game/cards/characters/07/southronmessenger.js
```diff
@@ -0,0 +1,30 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SouthronMessenger extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardEntersPlay: (event, card) => (
+                    card === this &&
+                    this.game.currentChallenge &&
+                    this.game.currentPhase === 'challenge')
```
cryogen: This is redundant when you're checking that there's a current challenge

server/game/drawcard.js
```diff
@@ -155,6 +155,22 @@ class DrawCard extends BaseCard {
         return icons;
     }
 
+    getNumberOfIcons() {
+        var count = 0;
```
cryogen: return _.size(this.icons)

would also work and be less code :)

PR #681, opened 2017-03-25
Implement Rattleshirt

DukeTax:


server/game/cards/characters/07/rattleshirt.js
```diff
@@ -0,0 +1,18 @@
+const DrawCard = require('../../../drawcard.js');
+
+class Rattleshirt extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => (
+                this.game.currentChallenge &&
+                this.game.currentChallenge.isAttacking(this) &&
+                this.game.currentChallenge.attackers.length === 1),
```
cryogen: I think challenge needs a isAttackingAlone() method

PR #682, opened 2017-03-25
Implement Weirwood Bow

DukeTax:


DukeTax:
'Undo last commit' in VS Code didn't do what I thought it would...

PR #684, opened 2017-03-25
Implement Whisper Campaign

DukeTax:


server/game/cards/plots/07/whispercampaign.js
```diff
@@ -0,0 +1,26 @@
+const PlotCard = require('../../../plotcard.js');
+
+class WhisperCampaign extends PlotCard {
+    setupCardAbilities() {
+        this.forcedReaction({
+            when: {
+                onCardEntersPlay: (event, card) => {
+                    if(!(card.getType() === 'character' && !card.hasIcon('intrigue'))) {
```
cryogen: This is pretty confusing, can you rewrite it to be less double negative? :)

Something like 

```if(card.getType() !== 'character' || !card.HasIcon('intrigue')) {
     return false;
}```

DukeTax:
The double negative allowed you to basically read the card from the check, but I don't care either way. :)

PR #686, opened 2017-03-26
Add card name to ability prompts.

ystros:
Now when a character ability triggers a prompt, the name of the card
will be displayed in the menu prompt. This should help reduce confusion
during nested ability windows where a new window is opened and resolved
and then it returns to the original window.


ystros:
Looks like this currently:
screen shot 2017-03-25 at 12 28 13 pm

I tried distinguishing the card name from the phase indicator a few ways but couldn't get it looking much better than that. Should be a good enough starting point, let me know if you have a better idea of how it should look.

cryogen:
This should probably use the cards title rather than its name including packname, but i'll merge as is and may tinker with it

PR #688, opened 2017-03-26
Restrict Littlefinger to marshalling phase

### LINKED ISSUE #687: Ghosts of Harrenhal + Littlefinger
When Littlefinger is put into play via Ghosts of Harrenhal, his ability triggers. It shouldn't.

DukeTax:
Fixes #687.


mplain:
This is not an ideal fix. During the marshaling phase I might want to use Arianne to put Littlefinger into play, so that I can marshal Arianne again using my Martell discounts -- a reasonably probably scenario, and this will now trigger Littlefinger's ability.

PR #698, opened 2017-03-27
Protect against double kill through burn effects.

### LINKED ISSUE #678: Benjen triggers twice with one copy in hand and one in play
Happened to me yesterday: had 1 copy of Benjen in play and 1 in hand. The one in play dies, i got prompted to trigger his ability and I do. After that, I got prompted a second time. I denied the trigger,but I suspect it was the copy in hand triggering.

ystros:
Previously, burn effects could end up killing a character twice as
effects are reapplied. This could end up triggering reactions multiple
times innapropriately. Now the `killByStrength` guards against this
condition.

Fixes #678.

### DIFF
--- a/server/game/effects.js
+++ b/server/game/effects.js
@@ -240,7 +240,9 @@ const Effects = {
     },
     killByStrength: {
         apply: function(card, context) {
-            if(card.getStrength() <= 0) {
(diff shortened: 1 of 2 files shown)


ystros:
Integration tests introduced here will need to be updated once #694 is merged.

PR #712, opened 2017-03-31
fix Devoted Bloodrider to boost only bloodriders you control

### LINKED ISSUE #711: after controller change, Devoted Bloodrider gives +1 to the wrong bloodriders
Seen yesterday playing: I marshaled a Devoted Bloodrider from the discard pile of my opponent using Night Gathers. My opponent had bloodriders in play. As a result, both mine (as Devoted controller) and their (as Devoted owner) bloodriders got the +1 strength modifier. Expected behavior is that only mine (as controller) should have been boosted.

zacchiro:
Fixes #711


ystros:
I think this issue is a problem with take control and not with Devoted Bloodrider. Your change here should be identical to what the current implementation is, since controller is checked by the `targetController` property (which defaults to the current controller).

PR #718, opened 2017-03-31
Improve faction icon colors

DukeTax:
Changed the faction icon colors to better match their faction. Took the colors from the colored banners on the main page of thronesDB, with the exception of Stark (white), NW (black) and Targaryen (darkish red), because of background concerns

### DIFF
--- a/less/icons.less
+++ b/less/icons.less
@@ -48,8 +48,6 @@
   &:before {
     content: "\e600";
     color: #e3d852;
-    //background-color: #FECF03;
-    //color: grey;
   }
 }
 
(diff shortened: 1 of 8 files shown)


less/icons.less
```diff
@@ -48,7 +48,8 @@
   &:before {
     content: "\e600";
     //background-color: #FECF03;
-    color: grey;
+    //color: grey;
```
ystros: Can't really blame you for doing this since it was done previously with the background colors, but generally we should prefer removing old code rather than commenting it out.

ystros:
Deferring approval of the color changes themselves to @cryogen.

cryogen:
Screenshots would help, I'm far too lazy to check out the branch and look myself :)

PR #723, opened 2017-04-01
add integration test for Wildling Bandit

zacchiro:


zacchiro:
Note: as far as I can tell (and observed in play) Wildling Bandit is currently broken, hence this new test will fail for the time being. I won't bother filing an issue for it, let's just fix it before merging this test.

test/server/cards/characters/05/05041-wildlingbandit.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+            this.completeMarshalPhase();
+        });
+
+        describe('when attacking a more wealthy opponent', function() {
+            beforeEach(function() {
+                this.skipActionWindow();
+
+                this.player1.clickPrompt('Military');
+                this.player1.clickCard(this.bandit);
```
ystros: Thanks for doing this, but your test is not quite correct. Attackers aren't declared until you've selected them AND have clicked 'Done'. You need a `this.player1.clickPrompt('Done')` here to declare the bandit as attacking, after which the test passes. Good to have the test, but it doesn't reveal a problem with the card or effects engine.

PR #725, opened 2017-04-01
add integration test for Jhogo

zacchiro:


zacchiro:
As the strength boosting of Jhogo seems to be broken at the moment, (half of) this test currently doesn't pass.
Note however that the other half—gaining stealth when you control another bloodrider—passes just fine.

test/server/cards/characters/02/02113-jhogo.spec.js
```diff
@@ -0,0 +1,56 @@
...
+            this.player2Object.moveCard(this.chud1, 'dead pile');
+            this.player2Object.moveCard(this.chud2, 'dead pile');
+        });
+
+        describe('when attacking', function() {
+            beforeEach(function() {
+                this.skipActionWindow();
+
+                this.player1.clickPrompt('Military');
+                this.player1.clickCard(this.jhogo);
```
ystros: Same problem as the Wildling Scout test in #723  - the player needs to click 'Done' in order to declare attackers. After adding that button click, the test passes.

PR #726, opened 2017-04-01
Fix Iron Flout Scout label

### LINKED ISSUE #721: Iron Fleet Scout prompt says '+1 strength' even if first player
It does apply +2 correctly - just a typo in the prompt

DukeTax:
Fixes #721

Have no idea where this json stuff is coming from..

Edit: I have probably ran fetchdata.js more recently than you...

### DIFF
--- a/server/game/cards/locations/01/ironfleetscout.js
+++ b/server/game/cards/locations/01/ironfleetscout.js
@@ -3,7 +3,7 @@ const DrawCard = require('../../../drawcard.js');
 class IronFleetScout extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
-            title: 'Kneel this card to give a character +1 STR',
+            title: 'Kneel this card to give a character +STR',
             cost: ability.costs.kneelSelf(),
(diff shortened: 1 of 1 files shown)


zacchiro:
Note that you included changes to the json-data submodule in this PR, you might want to clean it up.

Also, given Scout gives either +1/+2, i'd rather have a label that either mentions both possibilities ("e.g., give +1/+2") or that is generic ("increase STR").

cryogen:
The submodules are out of sync, you'll need to exclude changes to that in your commit.

However, I'm not sure what to do about this, this used to work dynamically telling you the STR used.  Since changing to the new API we seem to have lost that ability.  @ystros is it still possible to have a dynamic title in The New World(tm)?

ystros:
@cryogen I'm not sure we ever had dynamic titles for actions? Certainly never had it for this card at least. It's possible to implement (`title` will need to be able to take a function instead of a string), but it will also require reworking menus, which currently registers its menu item w/ static title when the card is created.

PR #728, opened 2017-04-03
add remove from game cost and port Varys/House of the Undying

zacchiro:


server/game/cards/characters/01/varys.js
```diff
@@ -3,21 +3,23 @@ const _ = require('underscore');
 const DrawCard = require('../../../drawcard.js');
 
 class Varys extends DrawCard {
-    setupCardAbilities() {
+    setupCardAbilities(ability) {
         this.interrupt({
             when: {
-                onPhaseEnded: (event, phase) => phase === 'dominance'
+                onPhaseEnded: (event, phase) => phase === 'dominance',
+                cost: ability.costs.removeSelfFromGame()
```
ystros: Cost should not go inside the `when` clause, it should be at the same level as `when` and `handler`.

PR #733, opened 2017-04-04
Convert most event cards to use `action`.

ystros:


ystros:
The only event cards not converted that have **Action** in their text were the 'select up to X strength' type cards, because they're kind of weird in the first place. Will probably look at cleaning up that type of prompt before converting those.

PR #743, opened 2017-04-08
implement The Eyrie, adding a "cannot be killed" effect to that end

zacchiro:


server/game/player.js
```diff
@@ -792,7 +792,10 @@ class Player extends Spectator {
             return;
         }
 
-        if(!character.dupes.isEmpty() && allowSave) {
+        if(!character.canBeKilled()) {
+            this.game.addMessage('{0} controlled by {1} cannot be killed',
+                                 this, character);
```
ystros: Message args are slightly out of order - this will print 'player1 controlled by Tywin Lannister cannot be killed'

zacchiro:
On Sat, Apr 08, 2017 at 02:51:08PM -0700, Brian Upton wrote:
> Message args are slightly out of order - this will print 'player1
> controlled by Tywin Lannister cannot be killed'

Nice catch! Fixed, thanks.

(even though Tywin controlling a player is not entirely unheard of :-P)

ystros:
CI failure seems to be unrelated (happened in #744 as well, then stopped in PRs after), merging.

PR #746, opened 2017-04-09
Choose stealth targets before challenge initiated.

### LINKED ISSUE #450: Initiating challenges and stealth/reaction timing
At the moment, let's say I initiate an intrigue challenge with Tyrion Lannister; I choose intrigue, then Tyrion as attacker, then get prompted for Tyrions reaction, and finally I can choose a stealth target.

This is slightly off. Correct order would be type of challenge->characters being declared as attackers->possible stealth targets all in one step, then followed by a reaction window where players in player order can react to the initiated challenge.
Ref: http://thronesdb.com/rulesreference#4.2_Active_player_may_initiate_a_challenge

This matter little at the moment, but might become more pressing when a card like Winterfell will be implemented.

ystros:
Per the rules, stealth targets must be chosen prior to any reactions or
interrupts firing for the challenge initiation.

Fixes #450.

### DIFF
--- a/server/game/gamesteps/challenge/challengeflow.js
+++ b/server/game/gamesteps/challenge/challengeflow.js
@@ -15,7 +15,6 @@ class ChallengeFlow extends BaseStep {
             new SimpleStep(this.game, () => this.resetCards()),
             new SimpleStep(this.game, () => this.announceChallenge()),
             new SimpleStep(this.game, () => this.promptForAttackers()),
-            () => new ChooseStealthTargets(this.game, this.challenge, this.challenge.getStealthAttackers()),
             new SimpleStep(this.game, () => this.announceAttackerStrength()),
             new ActionWindow(this.game, 'After attackers declared'),
             new SimpleStep(this.game, () => this.promptForDefenders()),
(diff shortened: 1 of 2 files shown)


cryogen:
Will this still allow challenges to be cancelled by not selecting anyone?  I think yes, but just checking

ystros:
Yes, clicking 'Done' when you haven't selected any attackers still cancels the challenge. Not selecting a stealth target doesn't cancel the challenge, but it didn't before and there are valid use cases for not selecting one (e.g. no valid targets, wanting to force your opponent into opposing the challenge, etc)

cryogen:
👍🏼

On 11 Apr 2017, at 17:34, Brian Upton <notifications@github.com> wrote:

Yes, clicking 'Done' when you haven't selected any attackers still cancels
the challenge. Not selecting a stealth target doesn't cancel the challenge,
but it didn't before and there are valid use cases for not selecting one
(e.g. no valid targets, wanting to force your opponent into opposing the
challenge, etc)

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/cryogen/throneteki/pull/746#issuecomment-293320489>,
or mute
the thread

PR #747, opened 2017-04-09
Give reduction to player that activated Kingsroad.

### LINKED ISSUE #569: Stolen Kingsroad reduces the cost to the original owner
Just a minor thing: if you steal a Kingsroad with Euron and use it next turn, the -3 reduction isn't applied to the current controller, but the original owner of the card.

ystros:
Previously, the Kingsroad was always giving its reduction to the player
set in the `controller` property. However, because the Kingsroad was
sacrificed prior to applying the reduction, the controller was always
the original owner. This lead to a bug where the Kingsroad would apply
its reduction improperly to the owner even when it had been taken
control using cards like Euron Crow's Eye.

Fixes #569.

### DIFF
--- a/server/game/cards/locations/01/thekingsroad.js
+++ b/server/game/cards/locations/01/thekingsroad.js
@@ -13,11 +13,18 @@ class TheKingsroad extends DrawCard {
                 ability.costs.kneelSelf(),
                 ability.costs.sacrificeSelf()
             ],
-            handler: () => {
-                this.game.addMessage('{0} kneels and sacrifices {1} to reduce the cost of the next character by 3', this.controller, this);
+            handler: context => {
+                // Because the Kingsroad ends up in the discard pile of its owner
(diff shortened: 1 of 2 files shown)


ystros:
Easier to review the new test case without whitespace. The old take control tests just got wrapped in a describe and indented. https://github.com/cryogen/throneteki/pull/747/files?w=1

zacchiro:
I've just mimicked this patch in #752 to fix #750 , feedback welcome.

Note that I haven't been this verbose in the comment about why we need to extract current controller from context, because it looks like we're gonna have a bunch of places where that info is gonna be repeated. Maybe we should note it down in implementing-cards, or something.

cryogen:
As I said on the other PR, I'm not super keen on the card implementations having to keep track of this, it feels like something the engine/reactions stuff should keep track of, or be told to keep track of via a flag of some sort.  However it will do for now

PR #752, opened 2017-04-09
make House of the Undying work for the current controller

### LINKED ISSUE #750: stolen House of the Undying works for the wrong player
Just witnessed a match where a stolen House of The Undying, put into play using Euron, when used worked for the wrong player, i.e., the owner rather than the current controller.

zacchiro:
Before this fix House of the Undying always worked for its owner, because
removing it from the game (which happens before the effect) reset the
controller to the owner.

Fixes #750


cryogen:
@ystros  FYI I don't like that we're having to do this

PR #753, opened 2017-04-09
Prevent reactions from firing from wrong location.

### LINKED ISSUE #640: Harrenhal vs Littlefinger
I marshal Littlefinger, opponent (first player) kills him with Harrenhal, the game still asks whether I want to trigger Littlefinger.

ystros:
Previously, if an interrupt or reaction ended up removing another card
from play (e.g. Harrenhal kills Littlefinger when he enters play), then
the player would still be prompted to trigger the removed card's
abilities. Now reactions check that the card is still in the proper area
to be eligible to activate.

Fixes #640.

### DIFF
--- a/server/game/triggeredability.js
+++ b/server/game/triggeredability.js
@@ -64,6 +64,10 @@ class TriggeredAbility extends BaseAbility {
             return false;
         }
 
+        if(this.card.location !== this.location) {
+            return false;
+        }
+
         if(!this.canPayCosts(context) || !this.canResolveTargets(context)) {
             return false;
         }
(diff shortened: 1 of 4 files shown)


cryogen:
Are there any scenarios where it's still legit for a reaction to trigger even though it's location has changed?  Like that new card that triggers from the discard pile, or something similar?

ystros:
Not that I can think of. The Prince's Plan will check that it is still in discard before triggering because the ability location for that one is explicitly set to 'discard pile.' The analogous scenario for that one would be if there were an on win/lose reaction that could shuffle Prince's Plan back into the deck in the same window it would normally execute.

The other edge case is for cards that change position as part of a cost (e.g. sacrifice card to do something, event cards moving from hand to discard). However, `meetRequirements` is only checked prior to trying to resolve the ability. Once it starts being resolved and costs are paid, it isn't checked again. There are tests for other cards (e.g. Dolorous Edd, some event card tests) that should catch any problems should that change.

PR #779, opened 2017-04-18
Implement Ygritte's stealth ability.

ystros:


server/game/cards/characters/06/ygritte.js
```diff
@@ -0,0 +1,20 @@
...
+        // TODO: Cannot be knelt by card effects.
+        this.persistentEffect({
+            condition: () => this.controlsAnotherWildling(),
+            match: this,
+            effects: ability.effects.addKeyword('stealth')
+        });
+    }
+
+    controlsAnotherWildling() {
+        return this.controller.cardsInPlay.any(card => card.hasTrait('Wildling') && card.getType() === 'character' && card !== this);
```
cryogen: I've been moving away from directly operating on cardsInPlay recently in favour of a method on player, can we do that here?

ystros: 👍

server/game/cards/characters/06/ygritte.js
```diff
@@ -0,0 +1,20 @@
...
+        // TODO: Cannot be knelt by card effects.
+        this.persistentEffect({
+            condition: () => this.controlsAnotherWildling(),
+            match: this,
+            effect: ability.effects.addKeyword('stealth')
+        });
+    }
+
+    controlsAnotherWildling() {
+        return this.controller.anyCardsInPlay(card => card !== this && card.getType() === 'character' && card.hasTrait('Wildling'));
```
cryogen: sorry, I was thinking something like findCard() - or should this not scan attachments too?  if so perhaps we need a findCard that skips attachments (kinda like the one youve done)

ystros: Kind of trying to get rid of `findCard` outside using it inside of `Player`, it's only a slight step of encapsulation above searching `cardsInPlay` directly. But I can change it.

Neither `findCard` nor this version of `anyCardsInPlay` searches attachments. I did continue off of this point on another branch to convert the rest of cards to use helper methods (no more direct usages of `.cardsInPlay`) and ended up changing `anyCardsInPlay` to search attachments. Doesn't really matter for Ygritte though.

cryogen: findCard(s) definitely does search attachments

Issue #764, opened 2017-04-12
Rains with Caswell's Keep


Ecth3lion:
When triggering Rains of Castamere, if you have Caswell's Keep marshalled you supposedly get a trigiger after revealing a new plot. However you are not given the option to do so and the trigger is impossible to do manually.

cryogen:
Fixed by #785

PR #785, opened 2017-04-20
Make plot reveal process more consistent

### LINKED ISSUE #758: Old Nan triggers before you see plots
Old Nan's prompt (to give trait or pass) appears after plots are selected, but before you can actually see them.

ystros:
* Replaces the event that cards should interrupt / react on when revealing a plot - `onPlotsRevealed` which has a list of all plots being revealed at the moment. This event wraps the entire plot reveal process, meaning all "when revealed" abilities are resolved by the time the reaction window opens.
* Modifies Rains of Castamere to use the extracted reveal plots class to maintain consistency between the plots it reveals and the normal process.
* Varys's Riddle and Pulling the Strings explicitly don't use the new class, as they only "initiate" the when revealed ability and don't count as revealing the plot they duplicate in terms of reactions.
* Plot resolution order is now handled by the same mechanism that allows first player to choose the order of forced interrupts and reactions.

Fixes #758 and #764.


server/game/basecard.js
```diff
@@ -170,15 +171,14 @@ class BaseCard {
         this.abilities.reactions.push(reaction);
     }
 
-    // TODO: When revealed abilities shouldn't be a synonym for forced reactions
-    //       but it is probably close enough for now.
     whenRevealed(properties) {
```
cryogen: I think I understand why, but I find it a little weird that this is on basecard and not plotcard - AFAIK, draw cards don't/can't have 'when revealed's

server/game/cards/attachments/04/beggarking.js
```diff
@@ -9,10 +9,10 @@ class BeggarKing extends DrawCard {
         });
         this.reaction({
             when: {
-                onPlotRevealCompleted: () => (
-                    _.any(this.game.getPlayers(), player => (
-                        player !== this.controller &&
-                        this.controller.activePlot.getIncome(true) < player.activePlot.getIncome(true)
+                onPlotsRevealed: event => (
```
cryogen: This logic is a bit clumsy now, although it will be nice for multi-player support if we ever do it, it's now checking all plots and ignoring one of them, rather than just getting the other player's plot and comparing that directly, if that makes sense?

server/game/cards/characters/03/oldnan.js
```diff
@@ -6,12 +6,12 @@ class OldNan extends DrawCard {
...
             when: {
-                onPlotFlip: () => true
+                onPlotsRevealed: () => true
             },
             cost: ability.costs.kneelSelf(),
-            handler: () => {
-                var buttons = _.map(this.game.getPlayers(), player => {
-                    return { text: player.selectedPlot.name, method: 'plotSelected', arg: player.selectedPlot.uuid };
+            handler: context => {
+                let buttons = _.map(context.event.plots, plot => {
```
cryogen: whereas here i think it works just fine 👍

server/game/basecard.js
ystros: Correct, I just lumped them all in one spot when building out the ability API. Could certainly move it to `PlotCard`.

server/game/cards/attachments/04/beggarking.js
ystros: The logic is more correct now since it's only looking at the plots revealed instead of the plots for all players (e.g. Rains where only one person is revealing a plot), but I agree it's ugly.

I generally try to avoid code that assumes only 2 players in case we find a good way to support Melee but we can worry about that later. I've updated Beggar King in the interest of clarity.

PR #796, opened 2017-04-23
Implement Knight of the Reach

DukeTax:


server/game/cards/characters/06/knightofthereach.js
```diff
@@ -0,0 +1,37 @@
...
+            },
+            handler: context => {
+                context.target.modifyPower(1);
+                this.game.addMessage('{0} uses {1} to have {2} gain 1 power', 
+                                      this.controller, this, context.target);
+            }
+        });
+    }
+
+    hasSingleParticipatingChar() {
```
cryogen: This is a candidate for moving on to the challenge itself if this starts becoming a common thing

PR #800, opened 2017-04-24
Games List > A few style changes

xabiermartinez:
Minor fixes and contrast changes.

screenshot-localhost-4000-2017-04-24-14-17-42
screenshot-localhost-4000-2017-04-24-14-16-55

### DIFF
--- a/client/GameList.jsx
+++ b/client/GameList.jsx
@@ -72,9 +72,14 @@ class InnerGameList extends React.Component {
 
             return (
                 <div key={ game.id } className='game-row'>
-                    <div><b>[{ game.gameType }] { game.name }</b>{ this.props.isAdmin && this.props.showNodes ? <span className='game-node'>Node: { game.node }</span> : null}</div>
-                    { gameLayout }
-                    <span className='pull-right'>
+                    <div>
+                        <b className='title'>{ game.name }</b>
+                        { this.props.isAdmin && this.props.showNodes ? <span className='game-node'>Node: { game.node }</span> : null}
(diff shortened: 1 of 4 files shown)


cryogen:
Could you include a screenshot or two so I can see what the changes look like please? Thanks!

xabiermartinez:
Sure, I'm at work, at the lunch time I could upload them.

xabiermartinez:
Done!

cryogen:
Hmm.  The game type is in the title on purpose to draw people's attention to it.  Also, the right hand side is pretty crowded already, especially on mobile

xabiermartinez:
I moved game type because where it is has no visibility, even less being of the same colour. Being near join/watch buttons is useful, when you join/watch is when you want to know the game type. 
IMO the right hand side is clear, is the left side the crowded one.

zacchiro:
FWIW, i agree with @xabiermartinez that where it is the game type is not really visible. While it is true that it currently appears *before* anything else, it gets visually lost within titles — that's why the other day I proposed to maybe change its font a little bit.

I don't know if moving the game type to the right is the nest solution (and I've no idea about the crowded aspect, given I always play on a desktop), but *some* way of making the game type stand out  from the title is needed.

cryogen:
Going to leave this open so I don't lose it, I am going to do some of these changes :)

xabiermartinez:
Sure.

cryogen:
This has now been incorporated into a future change, or at least the bits i want to, for now

PR #813, opened 2017-04-27
Simplify prompt buttons representing cards.

ystros:
Previously, when a menu prompt button was associated with a card, there
was a lot of repetition in that the text, argument and card summary were
almost always the same. Now, it is possible to just specify the card and
the other arguments are automatically filled in. If the `text` or `arg`
properties are specified, they will override the default card-generated
values.

Additionally, instead of sending the full card summary, only the
information used for card previews is sent to the client.

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -443,6 +443,15 @@ class BaseCard {
         return false;
     }
 
+    getShortSummary() {
+        return {
+            code: this.cardData.code,
+            label: this.cardData.label,
+            name: this.cardData.name,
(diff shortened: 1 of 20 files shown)


server/game/basecard.js
```diff
@@ -443,6 +443,15 @@ class BaseCard {
         return false;
     }
 
+    getShortSummary() {
```
ystros: If you have a better name for this, please suggest it. It's being used for card preview on the client side, but didn't want to tie the naming to that usage.

PR #815, opened 2017-04-27
Yoren should only activate when marshalled.

ystros:


ystros:
Going to close this one in favor of the real fix applied to all cards - this will still fire inappropriately if Yoren is put into play by a card effect during the marshal phase.

PR #823, opened 2017-04-28
Add support for passworded games

cryogen:


server/pendinggame.js
```diff
@@ -69,33 +71,71 @@ class PendingGame {
...
             return false;
         }
 
-        this.spectators[user.username] = {
-            id: id,
-            name: user.username,
-            emailHash: user.emailHash,
-            user: user
-        };
+        this.addPlayer(id, user);
```
ystros: If I'm reading this correct, this will count spectators toward the player count. So if I start a game and then a spectator watches, no one else can join as a player and I could start the game with that spectator. Even weirder, if another spectator watches, a player COULD join since the failure check is `_.size(this.players) === 2`, at which point you could start the game with 3+ players.

Additionally, you updated the usage of `PendingGame.watch` in the lobby to take a password and callback like `join`, but haven't modified here to implement that.

cryogen: oops, I tried to reduce some duplicated code, but it wasn't as duplicated as I thought it was..

PR #824, opened 2017-04-28
Fix selectable card state when selecting opponent cards.

ystros:
Previously, the `selected` and `selectable` state for cards was stored directly on the card. This made it difficult to hide selection state from the opponent when their cards were selected or selectable (hence the `opponentSelected` flag). In the case of `selectable`, it was impossible to show the selecting player which opponent cards were selectable without also showing the opponent.

Now, the list of selected and selectable cards is stored directly on the player object doing the selection. This made it easy to only show selection state to that player, though it did require passing the full player object in the serializers instead of a boolean flag indicating 'active' player.

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -452,7 +452,8 @@ class BaseCard {
         };
     }
 
-    getSummary(isActivePlayer, hideWhenFaceup) {
+    getSummary(activePlayer, hideWhenFaceup) {
+        let isActivePlayer = activePlayer === this.owner;
         return isActivePlayer || (!this.facedown && !hideWhenFaceup) ? {
             code: this.cardData.code,
             controlled: this.owner !== this.controller,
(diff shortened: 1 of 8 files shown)


server/game/basecard.js
```diff
@@ -452,16 +452,19 @@ class BaseCard {
...
         return isActivePlayer || (!this.facedown && !hideWhenFaceup) ? {
             code: this.cardData.code,
             controlled: this.owner !== this.controller,
             facedown: this.facedown,
             menu: this.getMenu(),
             name: this.cardData.label,
             new: this.new,
-            selected: (isActivePlayer && this.selected) || this.opponentSelected,
-            selectable: (isActivePlayer && this.selectable),
+            // The `this.selected` property here is a hack for plot selection,
```
cryogen: We should probably make this go away in the future

PR #825, opened 2017-04-28
Implement Marriage Pact

DukeTax:
The forced interrupt on Marriage Pact is not working currently which must be the same issue as Viserys Targaryen.


cryogen:
Pretty icky that we have to put that check on wardens et al now, but i dont have any better suggestions!

PR #827, opened 2017-04-29
Implement events from Guarding the Realm

DukeTax:
Both the reactions from discard pile don't seem to be working currently...


ystros:
#836 fixes the problem with reactions from discard pile.

PR #839, opened 2017-05-03
Auto pass windows

cryogen:
Add the ability to auto pass on action windows
default the auto pass on some lesser used windows
implement a menu option for people to set whether to auto pass or not


cryogen:
image

cryogen:
image

cryogen:
image

cryogen:
image

cryogen:
image

zacchiro:
I think that's amazing, thanks for this!

As a UX comment, I suspect reversing the logic would be better. The current way is asking what to *pass*, with checkboxes, which kinda sound like double negation in a way.

How about a "Prompt for action windows", with checkboxes for action windows to show instead?
Or maybe, with even simpler language, "Show action windows" in the dialog box title.

Just my 0.02 EUR, this feature will be amazing anyhow.

client/GameComponents/AutoPassMenu.jsx
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    render() {
+        return (
+            <div className='panel auto-pass'>
+            Auto pass action windows
+                <div className='checkbox'>
+                    <label>
+                        <input type='checkbox' checked={ this.props.options.plot } onChange={ this.onToggle.bind(this, 'plot', this.props.options.plot) } />Plots Revealed
```
ystros: This is nitpicky, but since there are so many of these and they're all the same, it may be more readable to express this as an array and then use a loop to render them. e.g.
```javascript
let actionWindows = [
  { flag: 'plot', text: 'Plot Revealed' },
  { flag: 'draw', text: 'Draw Phase' },
  // ...
];

_.map(actionWindows, actionWindow => (
  <div className='checkbox'>
    <label>
      <input type='checkbox' checked={ this.props.options[actionWindow.flag] } ... />{actionWindow.text}
    </label>
  </div>
));
```

client/GameComponents/MenuPane.jsx
```diff
@@ -33,10 +33,10 @@ class MenuPane extends React.Component {
 
         var buttons = _.map(this.props.buttons, button => {
             var option = (
-                <button key={button.command + buttonIndex.toString()} className='btn btn-primary'
-                    onClick={event => this.onButtonClick(event, button.command, button.arg, button.method)}
-                    onMouseOver={event => this.onMouseOver(event, button.card)} onMouseOut={event => this.onMouseOut(event, button.card)}
-                    disabled={button.disabled}>{button.text}</button>);
+                <button key={ button.command + buttonIndex.toString() } className='btn btn-primary'
```
ystros: Is there an eslint rule we can update to enforce this style change?

client/GameComponents/AutoPassMenu.jsx
cryogen: Agreed

client/GameComponents/MenuPane.jsx
cryogen: No :( and what's more vscode enforces the other way, which is why so much of the code has it that way, but i find it quite unreadable

PR #843, opened 2017-05-03
Add test for phase ending vs at end of phase.

ystros:


ystros:
This will conflict with #839 since it assumes the plot phase window is always prompted. So whichever gets merged first, the other will need to remove the skipActionWindow here to get the test to pass.

PR #847, opened 2017-05-04
Add basic logging on disconnect and reconnect.

ystros:


ystros:
Test failures should be fixed by #846.

PR #853, opened 2017-05-05
Implement Plaza of Pride, using a new discard from hand cost

zacchiro:


server/game/costs.js
```diff
@@ -239,6 +239,43 @@ const Costs = {
...
+                    onCancel: () => {
+                        result.value = false;
+                        result.resolved = true;
+                    }
+                });
+
+                return result;
+            },
+            pay: function(context) {
+                context.player.moveCard(context.discardCostCard, 'discard pile');
```
ystros: This should be done explicitly with `player.discardCard` to ensure the appropriate discard events are raised (e.g. LoCR Cersei should be able to react to you paying this cost).

zacchiro: fixed

PR #854, opened 2017-05-05
Add User profiles

### LINKED ISSUE #75: Suggestion: add connected / online players
I would like to know who is online when I log in. That way, I will know whether or not it is a good idea to start a game.

### LINKED ISSUE #765: Option to unhook Gravatar from our accounts
Forcing Gravatar is slightly annoying if you don't want to change your Gravatar picture just to not show up in Throneteki or use a different email.

cryogen:
Allow change of email address and password
Allow gravatar images to be disabled
Allow default action windows to be set

Fixes #765

### DIFF
--- a/client/Profile.jsx
+++ b/client/Profile.jsx
@@ -180,6 +180,7 @@ class InnerProfile extends React.Component {
 InnerProfile.displayName = 'Profile';
 InnerProfile.propTypes = {
     refreshUser: React.PropTypes.func,
+    socket: React.PropTypes.object,
     user: React.PropTypes.object
 };
 
--- a/client/Profile.jsx
+++ b/client/Profile.jsx
@@ -29,7 +29,8 @@ class InnerProfile extends React.Component {
             loading: false,
             newPassword: '',
             newPasswordAgain: '',
-            promptedActionWindows: this.props.user.promptedActionWindows || this.windowDefaults
+            promptedActionWindows: this.props.user.promptedActionWindows || this.windowDefaults,
+            validation: {}
         };
 
         this.windows = [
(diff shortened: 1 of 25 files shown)


client/Profile.jsx
```diff
@@ -0,0 +1,195 @@
...
+    verifyPassword(isSubmitting) {
+        var validation = this.state.validation;
+
+        delete validation['password'];
+
+        if(!this.state.newPassword && !this.state.newPasswordAgain) {
+            return;
+        }
+
+        if(this.state.newPassword.length < 6) {
```
ystros: Either now or in a separate PR we should consider consolidating these rules (along with email validation) into shared logic so that we remain consistent. This will be the third place we validate password on the client side.

server/game/player.js
```diff
@@ -40,7 +40,7 @@ class Player extends Spectator {
         this.selectedCards = [];
         this.selectableCards = [];
         this.cannotGainChallengeBonus = false;
-        this.promptedActionWindows = {
+        this.promptedActionWindows = user.promptedActionWindows || {
```
ystros: Should we do an explicit `clone` here? As it currently stands, `this.promptedActionWindows` will reference the same object as `user.promptedActionWindows`. This means that as a player checks / unchecks things in a game, it will be modified on their user object as well, thus changing their account level preferences. Worse, even though it's changing their preferences, those changes will be lost the next the user object is loaded from scratch (guessing this only happens on initial log in). So we get this weird behaviour where 1. changes are being made to preferences without the user realizing it and 2. the changes are being reset at intervals not predictable to the user.

client/Profile.jsx
cryogen: Yeah, agreed.  Login and Register both need rewriting though, so I'll do it as part of that

server/game/player.js
cryogen: Surprisingly, it seems to reinitialise the user every game, so we're good here

PR #857, opened 2017-05-06
Fix viserys

### LINKED ISSUE #727: Viserys Targaryen (Core) is not triggering when leaving play
Currently, Viserys is not triggering when he leaves play. He should allow the player to discard an attachment.

This is a more general problem with self-interrupts to `onCardLeftPlay`. The `Player.moveCard` method fires the `onCardLeftPlay` event, but also moves the card prior to the resolution of that event. This means that the card is already out of play before the event is resolved, and thus is no longer listening to the event at all due to `Card.moveTo` unbinding the event listeners.

The solution probably should be that `onCardLeftPlay` should be handled in a similar asynchronous way like `onCharacterKilled` is.

cryogen:
Make discarding cards async so cards that trigger on leaving play trigger again.  Fixes #727

### DIFF
--- a/server/game/cards/characters/04/jaqenhghar.js
+++ b/server/game/cards/characters/04/jaqenhghar.js
@@ -37,7 +37,7 @@ class JaqenHGhar extends DrawCard {
         });
         this.forcedReaction({
             when: {
-                onCardLeftPlay: (event, player, card) => card === this
+                onCardLeftPlay: event => event.card === this
             },
             handler: () => {
                 if(this.selectedCards) {
--- a/server/game/cards/attachments/06/marriagepact.js
+++ b/server/game/cards/attachments/06/marriagepact.js
@@ -8,7 +8,7 @@ class MarriagePact extends DrawCard {
 
         this.forcedInterrupt({
             when: {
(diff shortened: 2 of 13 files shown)


server/game/cards/characters/04/jaqenhghar.js
```diff
@@ -37,7 +37,7 @@ class JaqenHGhar extends DrawCard {
         });
         this.forcedReaction({
             when: {
-                onCardLeftPlay: (event, player, card) => card === this
+                onCardLeftPlay: event => event.card === this
```
ystros: FYI, Jaqen is broken and will still be broken after this PR. The forced reaction will not (and should not) fire because it would be a reaction to itself leaving play - you can only interrupt self-leaving play. It also should not be implemented as an interrupt / reaction at all because it isn't something that can be cancelled (e.g. you can't Treachery getting rid of the tokens). Probably should open up a separate issue about this cause I keep forgetting about it.

ystros:
I'm somewhat scared of this (looks too straightforward from what I was expecting), but if that was all the test fixes you had to make it should be pretty good.

cryogen:
There may yet be some timing issues, but it seemed to not be necessary to make everything be async

Issue #414, opened 2017-02-18
support for immunity to game effects


zacchiro:
[Several cards](https://thronesdb.com/find?q=x%3A%22immune+to%22&sort=name&view=spoiler) support some sort of immunity to some or all game effects. In the current meta we have at least the following occurrences of this:
- immune to opponents' events
- immune to Omen card effects
- immune to opponent's character abilities
- immune to opponents' plot effects
- immune to card effects
- immune to opponents' triggered effects
- immune to events
- cannot be knelt by card effects (Ygritte)
- cannot be killed (The Eyrie)

We need a generic mechanism to:
1) allow cards to declare what they are immune it, possibly temporarily
2) allow game effects (or their enacters) to query target cards for immunity and skip effect application if so

The main difficulty is that immunity seems in most cases to be dependent on who is the origin of the effect (as type of card and possibly its controller). This gets in the way of the current style of coding cards where card implementations directly do calls like `player.killCharacter(card)` without passing around the information about *who* is commanding the kill.
That will need to be reworked, by either passing around "this" as event originator where needed, or by passing through a higher-level of abstraction that will check immunity before actually delivering the effect.

This is going to be ~~a pain~~ fun :-)

ystros:
In terms of *providing* the immunity, we'll need to introduce a new effect (or set of effects). Since immunity is provided by card text, can be provided from one card onto another (e.g. Maester Luwin giving Bran "immunity to plot effects"), and can be blanked like other abilities, it can be activated like any other effect.

We should probably code it such that there's a generic effect that takes a function w/ parameters for the source card and the type of ability it is ('action', 'reaction', etc). With the generic one in place we can write effects that use it for the more common cases.

Some potential examples:

```javascript
// Craster
this.persistentEffect({
    match: this,
    effect: ability.effects.immuneTo((source, type) => source.hasTrait('Omen'))
});

// Maester Luwin
this.persistentEffect({
    match: (card) => card.name === 'Bran Stark',
    effect: ability.effects.immuneToPlotEffects()
});
```

These effects should push the appropriate immunity function onto an array on the card (since you could potentially get multiple immunities). Then whatever method that does the kill (or any other action that is subject to immunity) can exclude any cards who has an immunity function that returns true.

```javascript
class DrawCard {
    isImmune(source, type) {
        return _.any(this.immunities, immunity => immunity(source, type));
    }
}

// The thing doing the kill
if(card.isImmune(this.source, this.type)) {
    return;
}

[...]

PR #859, opened 2017-05-07
Allow actions to be restricted per card.

ystros:
Provides infrastructure for #414 but does not implement immunity.

It is now possible to restrict an action from being taken on a card
based on restrictions placed on the card. As abilities are activated,
they are placed on a stack on `Game` to keep track of the current
ability's source card. When a game action is applied to a card (e.g.
being killed, discarded, knelt, etc), the card checks through the list
of restrictions to see if that particular action coming from the source
card is allowed. This provides the necessary infrastructure to implement
both immunity abilities as well as conditional 'cannot' effects (e.g.
Ygritte).

To make checking these restrictions more convenient, a `applyGameAction`
method is provided on `Game` that will only apply an action to eligible
cards.

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -46,6 +46,7 @@ class BaseCard {
             initiative: true,
             reserve: true
         };
+        this.abilityRestrictions = [];
         this.menu = _([]);
         this.events = new EventRegistrar(this.game, this);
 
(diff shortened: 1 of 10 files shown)


ystros:
May be slightly easier to review without whitespace - a bunch of noise from wrapping things in `applyGameAction`. https://github.com/cryogen/throneteki/pull/859/files?w=1

PR #867, opened 2017-05-10
Make the onCardPlayed event more consistent.

ystros:
* Ensures the onCardPlayed event is raised for events implemented via the legacy `play` method and through the ability API (including actions and reactions/interrupts).
* Ensures that additional event card abilities such as return-to-hand-from-discard do not count as playing an event.
* Removes the now unused onCanceled handler from the event window.

### DIFF
--- a/server/game/triggeredability.js
+++ b/server/game/triggeredability.js
@@ -53,7 +53,7 @@ class TriggeredAbility extends BaseAbility {
     }
 
     meetsRequirements(context) {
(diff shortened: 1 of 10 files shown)


server/game/triggeredability.js
```diff
@@ -97,6 +97,10 @@ class TriggeredAbility extends BaseAbility {
         return false;
     }
 
+    isPlayableEventAbility() {
```
cryogen: bothers me slightly that this is duplicated, but probably not enough to suggest refactoring these two classes to share a common base

PR #869, opened 2017-05-10
implement Balon GreyJoy (CtA)

cryogen:


server/game/cards/characters/04/balongreyjoy.js
```diff
@@ -0,0 +1,35 @@
...
+
+class BalonGreyjoy extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => {
+                var otherPlayer = this.game.getOtherPlayer(this.controller);
+
+                return !otherPlayer ||
+                    (this.game.currentChallenge &&
+                    this.game.currentChallenge.challengeType !== 'military' &&
```
ystros: Balon should only get his effect on military challenges, rather than on all but military challenges.

server/game/cards/characters/04/balongreyjoy.js
```diff
@@ -0,0 +1,35 @@
...
+                ability.effects.doesNotKneelAsAttacker()
+            ]
+        });
+
+        this.action({
+            title: 'Kneel a greyjoy location',
+            cost: ability.costs.kneel(card => card.isFaction('greyjoy') && card.getType() === 'location'),
+            handler: () => {
+                this.untilEndOfChallenge(ability => ({
+                    match: card => card.isLoyal() && card.controller === this.controller,
```
ystros: Probably should check that the card is a character for good measure - this will attempt to boost strength on loyal locations, for instance.

PR #878, opened 2017-05-12
Add 'cannot trigger card abilities'

cryogen:
Implement Catelyn Stark (Core)


server/game/forcedtriggeredability.js
```diff
@@ -28,6 +28,10 @@ class ForcedTriggeredAbility extends TriggeredAbility {
         this.handler = properties.handler;
     }
 
+    isForcedAbility() {
```
cryogen: Open to criticism on this, not sure if it's the direction you'd have gone in

server/game/game.js
```diff
@@ -814,6 +814,7 @@ class Game extends EventEmitter {
     }
 
     continue() {
+        this.effectEngine.reapplyStateDependentEffects();
```
cryogen: So, this one is interesting, when nested steps call continue themselves, effects are not recalculated.

e.g. Core Cat vs Tyrion

Challenge is initiated, stealth targets are chosen, 'onChallenge' event is raised before the game effects are recalculated

Not sure if there's a better way to fix that than this

server/game/gamesteps/challenge/challengeflow.js
```diff
@@ -17,6 +17,7 @@ class ChallengeFlow extends BaseStep {
             new SimpleStep(this.game, () => this.announceChallenge()),
             new SimpleStep(this.game, () => this.promptForAttackers()),
             new SimpleStep(this.game, () => this.announceAttackerStrength()),
+            new SimpleStep(this.game, () => this.chooseStealthTargets()),
```
cryogen: Split this up to get the pipeline to continue and apply effects in between the steps

ystros: Timing on this is now incorrect though - per the rules, the declaration of stealth targets is done at the time of choosing attackers but prior to the challenge officially being initiated. Hence, if I declare a challenge using Core Tyrion, the sequence should be:
```
Pick 'Intrigue' challenge
Choose Tyrion
Choose stealth target for Tyrion
Challenge is declared
Tyrion gains 2 gold
```

Moving it out here will inappropriately change the flow to:
```
Pick 'Intrigue' challenge
Choose Tyrion
Challenge is declared
Tyrion gains 2 gold
Choose stealth target for Tyrion
```

server/game/forcedtriggeredability.js
ystros: I might have gone the other way (`isThisAnAbilityDirectlyTriggeredByPlayerChoice` but with a shorter name) but this works fine. 👍

server/game/game.js
ystros: I'm a bit surprised this was sufficient to fix your test case, but I'm not sure it will work in the general case. What's really needed is to recalculate between each step in the pipeline, included nested pipelines like the individual phases (whereas this just reapplies before the first pipeline step after pausing for user input). Not sure how much overhead that will incur.

server/game/gamesteps/challenge/challengeflow.js
ystros: Per discussion earlier in Discord, misread this and everything is still in the right order. Sorry.

PR #883, opened 2017-05-13
Don't recalculate challenge strength after win.

ystros:
The Haunted Forest's bonus strength was being unapplied after the
challenge winner was determined but before unopposed power was granted.
Thus when strengths were checked for unopposed power, the defender
strength was 0 and power was granted. To prevent this, challenge
strength is no longer recalculated after a winner has been determined,
and the unopposed check looks at winner and loser strength instead of
attacker and defender.

### DIFF
--- a/server/game/challenge.js
+++ b/server/game/challenge.js
@@ -107,6 +107,10 @@ class Challenge {
     }
 
     calculateStrength() {
+        if(this.winnerDetermined) {
+            return;
+        }
+
         this.attackerStrength = this.calculateStrengthFor(this.attackers) + this.attackerStrengthModifier;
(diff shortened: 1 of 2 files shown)


server/game/challenge.js
```diff
@@ -199,7 +203,7 @@ class Challenge {
     }
 
     isUnopposed() {
-        return this.defenderStrength <= 0 && this.attackerStrength > 0;
+        return this.loserStrength <= 0 && this.winnerStrength > 0;
```
cryogen: These don't appear to be defined anywhere? Unless I'm missing something

ystros: They were already defined, we just weren't using them here. See https://github.com/cryogen/throneteki/blob/master/server/game/challenge.js#L144-L160

cryogen: 👍

PR #891, opened 2017-05-15
Prompt for dead pile order when multiple characters killed.

ystros:
* Adds an `ordered` property to selection prompts that will display numbers above the cards selected to indicate the selection order.
* When two or more characters are killed, the player is prompted to select the order to place the characters in the dead pile (top-most card first). If only one card is killed, the player will not receive the prompt. If the player clicks 'Done' after selecting the order for some cards but not all of them, it won't continue. If the player clicks 'Done' without selecting any of the cards, it will place them in the dead pile from left to right automatically (right card being top-most).
screen shot 2017-05-15 at 12 36 11 pm
screen shot 2017-05-15 at 12 37 25 pm
* Because killing characters was getting more complicated, it has been moved to its own game step class and off of `Game`.

### DIFF
--- a/client/GameComponents/Card.jsx
+++ b/client/GameComponents/Card.jsx
@@ -208,6 +208,14 @@ class Card extends React.Component {
         return dupes;
     }
 
+    getCardOrder() {
+        if(!this.props.card.order) {
+            return null;
+        }
+
+        return (<div className='card-order'>{this.props.card.order}</div>);
+    }
+
     showMenu() {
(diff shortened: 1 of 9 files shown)


ystros:
Probably easier to review the Valar test without whitespace - the existing test was wrapped in a describe blocked. https://github.com/cryogen/throneteki/pull/891/files?w=1#diff-c5940cea96f70ce8bdd19d7de301ec9c

PR #899, opened 2017-05-17
implement now my watch begins

cryogen:


server/game/cards/events/07/nowmywatchbegins.js
```diff
@@ -0,0 +1,22 @@
+const DrawCard = require('../../../drawcard.js');
+
+class NowMyWatchBegins extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardDiscarded: event => this.controller !== event.player &&
```
ystros: This is not quite the correct event. The card text reads 'After [card] is **placed** in an opponent's discard pile'.  Queenscrown, for example, explicitly does not discard cards but places them in discard, as does the upcoming [Wheels Within Wheels](https://thronesdb.com/card/06100). You can use the `onCardPlaced` event instead.

PR #910, opened 2017-05-18
Fix doubling of effects from setup attachments.

### LINKED ISSUE #833: Some attachments placed in setup double their effect
When some attachments are placed during setup, it grants whatever its effect twice.

For example, if a Valyrian Steel Dagger is attached to a character during setup, that character will get +4 STR during intrigue challenges. If it is attached any other time, they will get the proper +2 STR bonus.

This may be a result of the `onCardEntersPlay` event being fired twice for attachments during setup - once when placed, then again when attached. However, if that were the only cause I would expect all attachments in setup to have the same problem and they don't seem to.

ystros:
Attachments placed during setup were having their effects doubled in at
least some cases, because the effects were registered with the engine
twice - once when placed facedown, then again when attached. Now,
effects are only registered once for attachments when they're actually
attached.

Fixes #833.

### DIFF
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -447,8 +447,12 @@ class Player extends Spectator {
             this.removeCardFromPile(card);
             dupeCard.addDuplicate(card);
         } else {
+            // Attachments placed in setup should not be considered to be 'played',
+            // as it will cause then to double their effects when attached later.
+            let isSetupAttachment = playingType === 'setup' && card.getType() === 'attachment';
+
             card.facedown = this.game.currentPhase === 'setup';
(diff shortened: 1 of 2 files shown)


server/game/player.js
```diff
@@ -447,8 +447,12 @@ class Player extends Spectator {
             this.removeCardFromPile(card);
             dupeCard.addDuplicate(card);
         } else {
+            // Attachments placed in setup should not be considered to be 'played',
```
cryogen: This whole function should go on the 'to be refactored' list!

PR #912, opened 2017-05-19
Add a 'does not contribute strength to challenge' effect

ystros:
* Adds an effect that prevents strength from being added to a challenge.
* Converts Core Balon, The Kraken's Grasp, and Raiding Longship to use the effect.
* Implements Fishwhiskers.


cryogen:
:+1: :+1:

PR #921, opened 2017-05-19
Implement Healing Expertise

DukeTax:


server/game/cards/events/03/healingexpertise.js
```diff
@@ -0,0 +1,31 @@
...
+            },
+            location: 'hand',
+            cost: ability.costs.kneel(card => card.hasTrait('Maester') && card.getType() === 'character'),
+            target: {
+                activePromptTitle: 'Select character to save',
+                cardCondition: (card, context) => context.event.cards.includes(card) && !card.hasTrait('Army') && card.controller === this.controller
+            },
+            handler: context => {
+                context.event.saveCard(context.target);                
+                this.game.addMessage('{0} plays {1} to kneel {2} to save {3}',
```
zacchiro: nitpick on the game message, "plays {1}  to kneel {2} **and** save {3}"

DukeTax: It is on purpose, to distinguish between cost and effect, although it doesn't really look the best...

zacchiro:
On Sat, May 20, 2017 at 06:04:43AM -0700, DukeTax wrote:
> +                this.game.addMessage('{0} plays {1} to kneel {2} to save {3}', 
> It is on purpose, to distinguish between cost and effect, although it doesn't really look the best...

Yeah, I suspected that. For what is worth, given any player can always
see the actual card text within Throneteki, for me the messages should
be in general be more informative (and hence should be "nice") for the
players than normative. If that makes sense...

Thanks for implementing this! :-)

PR #932, opened 2017-05-21
Implement Lady Sansas Rose

DukeTax:
This is a bit of a messy implementation. If I've missed any easy opportunities to simplify it, please let me know. :)

### DIFF
--- a/server/game/cards/events/02/ladysansasrose.js
+++ b/server/game/cards/events/02/ladysansasrose.js
@@ -1,0 +1,53 @@
+const DrawCard = require('../../../drawcard.js');
+
+class LadySansasRose extends DrawCard {
+    setupCardAbilities() {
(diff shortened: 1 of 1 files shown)


server/game/cards/events/02/ladysansasrose.js
```diff
@@ -0,0 +1,53 @@
...
+            return (this.game.currentChallenge.isParticipating(card) &&
+                    card.hasTrait('Knight') && 
+                    card.getType() === 'character');
+        });
+
+        return !!cards.length;
+    }
+
+    hasLadyCharacter() {
+        let cards = this.controller.filterCardsInPlay(card => {
```
ystros: For checks like this you can just return the result of `this.controller.anyCardsInPlay(card => ...)` instead.

PR #936, opened 2017-05-22
Implement 'Max X per period' ability restrictions.

ystros:
* Adds `max` property to implement maximums per card title. Uses the same objects as the `limit` property.
* Uses the new property on all event cards that have been implemented via the ability API.
* Does NOT add the restriction to event cards implemented via the legacy `canPlay` / `play` mechanism. This should be done when those cards are converted.
* As a UI nicety, if the player has multiple max-restricted cards in hand, the prompt they see will only have one button.

### DIFF
--- a/server/game/baseability.js
+++ b/server/game/baseability.js
@@ -174,6 +174,10 @@ class BaseAbility {
     isCardAbility() {
         return true;
     }
+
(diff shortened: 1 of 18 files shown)


cryogen:
:+1:

PR #944, opened 2017-05-23
Implement Daring Rescue

DukeTax:


server/game/cards/events/05/daringrescue.js
```diff
@@ -0,0 +1,44 @@
...
+                    source: this,
+                    cardCondition: card => (
+                        card.location === 'play area' && 
+                        card.hasTrait('Knight') &&
+                        card.getType() === 'character' && 
+                        card.controller === this.controller),
+                    onSelect: (p, card) => {
+                        card.modifyPower(1);
+                        this.game.addMessage('{0} plays {1} to return {2} to its owner\'s hand then have {3} gain 1 power', 
+                                              this.controller, this, context.target, card);
```
zacchiro: As i understand the card, you might use it to return the sole Knight character you've in play. In that case, the event is playable, but you'll have no Knight to gain power in the second step. I think this card implementation would hang at that point, forcing the user to /cancel-prompt, isn't it?

Aside from that, the message should probably be made smarter to 1) differentiate the two cases, and 2) make sure it get posted when you return the last remaining Knight to hand.

DukeTax: Well, the promptForSelect method always has a 'Done' option in case there's no legal target, although that does mean no message at all will be displayed. So you're right the message needs to be built up in multiple steps I think...

PR #947, opened 2017-05-24
Implement The Pack Survives

DukeTax:


cryogen:
How does having 2 interrupts on the same card work, @ystros ?

ystros:
Each interrupt will register itself with the current prompt, with the `title` property added to the card's name, i.e.
```
[ The Pack Survives - Sac Direwolf ]
[ The Pack Survives - Kneel chars ]
[ Pass ]
```

It will only prompt with the versions that the cost can be paid for. Since each handler cancels the event, when you choose one the other should go away.

It's a bit weird looking, still might prefer an 'X or Y' cost mechanism, but this should work™️ .

cryogen:
Oh.  Ew.  I'm not sure I like that.  Especially with the standard of title's I've been approving lately because I thought they weren't displayed anywhere

ystros:
The titles on Action events I'm pretty sure wouldn't be displayed, but yeah, that's how it works on interrupts / reactions. It's kind of a vestigial mechanism - it was once used to help identify the targets during save interrupts but that was phased out in favor of selecting the cards directly. Pretty sure this if merged would be the only place actually using it now.

cryogen:
@DukeTax The danger here is that the buttons can end up really wordy and may end up getting truncated.  I think it would be preferable if first you chose the card to trigger, then get an option as to which cost to pay

DukeTax:
I tested this one quite extensively. Buttons all looked fine and it's the best we can do right now. Also easy enough to update once OR costs are implemented. Don't see any issue. :)

cryogen:
I think what I'm saying is I'd rather the 'or' costs were in before this went in.  Failing that, as a bare minimum, 'sac' has to become 'sacrifice'

cryogen:
and chars => characters

PR #948, opened 2017-05-24
Allow abilities to trigger when they enter play.

### LINKED ISSUE #245: Wardens of the West + Rains of Castamere
If Wardens of the West is played through triggering of Rains of Castamere, its reaction does not fire currently because the event has passed. 

More generally, if a card enters play as a result of a reaction or interrupt, it should be eligible for the same reaction trigger. This is probably most applicable to "when you win/lose a challenge" cards but there may be other situations it occurs.

ystros:
Previously, if an ability triggering brought a card into play or into
hand that should be playable, the new card wouldn't trigger because the
appropriate event listeners weren't bound when the original event
occurred. Now, when a card is moved to a location it would listen to
events, it registers its abilities with the Game object, which will add
the abilities to any prompts that are already open. This allows
interactions such as triggering Rains of Castamere to play Wardens of
the West, or allowing an event card drawn mid-reaction to be playable if
that ability window is still open.

Fixes #245.

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -286,23 +286,28 @@ class BaseCard {
     }
 
     moveTo(targetLocation) {
-        if(LocationsWithEventHandling.includes(targetLocation) && !LocationsWithEventHandling.includes(this.location)) {
+        let originalLocation = this.location;
+
+        this.location = targetLocation;
+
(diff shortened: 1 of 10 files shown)


ystros:
Probably easier to review the nested ability test change without whitespace (existing test got wrapped in a describe): https://github.com/cryogen/throneteki/pull/948/files?w=1#diff-c05ebc50c70af2b6a62fa26051c529b3

PR #959, opened 2017-05-26
Apply effects after the controller is set.

### LINKED ISSUE #711: after controller change, Devoted Bloodrider gives +1 to the wrong bloodriders

### LINKED ISSUE #845: Strength not recalculated when control is modified
Situation: Opponent had Ser Mallister in play with The Wall, boosting his strength to 5.  I played Ward on him to take control, and his strength was still set to 5 and had to be manually adjusted.

### LINKED ISSUE #906: Great Kraken for UO and when giving control
I took my opponents Great Kraken while using Euron Greyjoy and the game indicated  HIS balon still had stealth (mine too though). 

Another thing that happened is my opponent was controlling Great Kraken, I attack with something, he defends, i remove mine from the challenge and he got a Great Kraken proc, claiming it was unopposed (still not sure about ruling but i m pretty sure this is not how it is intended to work)

### LINKED ISSUE #924: euron and control change
hi, there are some may issues with control change interaction like when i pick up an iron mine with euron i cannot save my characters but i can save the rival characters.
And other cases like steal great kraken, steal the wall etc..

ystros:
Previously, if a card was put into play under another player's control
(e.g. through Euron's ability), effects were being applied before the
control change occurred. Thus, the effects would be applied to the
opponent and not the player taking control.

Fixes #711 
Fixes #845 
Fixes #906 
Fixes #924

### DIFF
--- a/server/game/effectengine.js
+++ b/server/game/effectengine.js
@@ -55,8 +55,16 @@ class EffectEngine {
                 // the new controller from scratch.
                 effect.cancel();
                 effect.addTargets(this.getTargets());
+            } else if(effect.duration === 'persistent' && effect.hasTarget(card) && !effect.isValidTarget(card)) {
+                // Evict the card from any effects applied on it that are no
+                // longer valid under the new controller.
+                effect.removeTarget(card);
             }
         });
+
+        // Reapply all relevant persistent effects given the card's new
+        // controller.
+        this.addTargetForPersistentEffects(card, 'play area');
     }
 
     addTargetForPersistentEffects(card, targetLocation) {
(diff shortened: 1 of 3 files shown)


DukeTax:
Just had an opponent take control of my Drogon, which then gave my Daenerys Renown. Would this be fixed through this PR as well?

ystros:
Yes, that sort of problem should be fixed by this. The problem is it was applying the effects before control was transferred, so Drogon and Dany had the same controller at the time.

PR #960, opened 2017-05-26
Implement House Frey cards from The Fall of Astapor

DukeTax:


server/game/cards/locations/06/thetwins.js
```diff
@@ -0,0 +1,29 @@
...
+                this.game.currentChallenge.attackingPlayer === this.controller && 
+                this.controller.getNumberOfChallengesInitiated() === 3 &&
+                this.hasAttackingFrey()),
+            match: card => card === this.controller.activePlot,
+            effect: ability.effects.modifyClaim(1)
+        });
+    }
+
+    hasAttackingFrey() {
+        let cards = this.controller.filterCardsInPlay(card => {
```
ystros: If you're just checking that there are any eligible cards, you can use `anyCardsInPlay` instead.

DukeTax: Thanks. Old habits die hard... :)

PR #961, opened 2017-05-26
Fix Ours is the Fury/Mother of Dragons to target characters that are allowed to participate in challenges

DukeTax:
Restrict Ours is the Fury/Mother of Dragons to target characters that are allowed to participate in challenges

### DIFF
--- a/server/game/cards/events/01/oursisthefury.js
+++ b/server/game/cards/events/01/oursisthefury.js
@@ -7,7 +7,7 @@ class OursIsTheFury extends DrawCard {
             condition: () => this.game.currentChallenge && this.game.currentChallenge.defendingPlayer === this.controller,
             target: {
                 activePromptTitle: 'Select character',
-                cardCondition: card => card.kneeled && card.controller === this.controller && card.isFaction('baratheon')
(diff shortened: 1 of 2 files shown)


server/game/cards/events/01/oursisthefury.js
```diff
@@ -7,7 +7,7 @@ class OursIsTheFury extends DrawCard {
             condition: () => this.game.currentChallenge && this.game.currentChallenge.defendingPlayer === this.controller,
             target: {
                 activePromptTitle: 'Select character',
-                cardCondition: card => card.kneeled && card.controller === this.controller && card.isFaction('baratheon')
+                cardCondition: card => card.kneeled && card.controller === this.controller && card.isFaction('baratheon') && card.canParticipateInChallenge()
```
ystros: Just FYI, there's a property in `target` that you can specify the intended game action, which will check whether the target is eligible for that action. So in this case, you could have done:
```
target: {
    // ...
    gameAction: 'participateInChallenge'
}
```

I don't think there's a real difference vs what you've done, just a shorthand, so this is fine.

PR #965, opened 2017-05-27
Implement Duel

DukeTax:


server/game/cards/plots/06/duel.js
```diff
@@ -0,0 +1,65 @@
...
+                    source: this,
+                    cardCondition: card => !card.hasTrait('Army') && card.getType() === 'character' && card.getCost() >= 6,
+                    onSelect: (player, cards) => this.targetsSelected(player, cards)
+                });
+            }
+        });
+    }
+
+    targetsSelected(player, cards) {
+        this.targets = cards;
```
ystros: Select card prompts work by selecting 'up to 2 cards' by default. So you'll need to check that exactly 2 cards were selected here, and return false if they didn't.

server/game/cards/plots/06/duel.js
```diff
@@ -0,0 +1,65 @@
...
+        this.game.addMessage('{0} has chosen {1} as the targets for {2}', player, cards, this);
+
+        this.game.promptWithMenu(player, this, {
+            activePrompt: {
+                menuTitle: 'Choose character to kneel',
+                buttons: [
+                    { text: cards[0].name, method: 'resolve', arg: [0,1]},
```
ystros: This is pretty clever use of arg. If we're going to do a menu here, you should use `card: card` so that you get a card preview when you hover over the button. I get the feeling that @cryogen would prefer a card selection prompt here though.

DukeTax:
Thanks for the feedback. I quite like the menu over a card selection prompt here when you have to choose one out of two, but we'll see what cryo says. :)

I also added a cancel button to the menu. Since it's not really hooked into the engine it's probably a good idea to leave users with escape options I think.

server/game/cards/plots/06/duel.js
```diff
@@ -0,0 +1,76 @@
...
+
+        this.targets = cards;
+        this.game.addMessage('{0} has chosen {1} as the targets for {2}', player, cards, this);
+
+        this.game.promptWithMenu(player, this, {
+            activePrompt: {
+                menuTitle: 'Choose character to kneel',
+                buttons: [
+                    { text: cards[0], method: 'resolve', arg: [0,1]},
+                    { text: cards[1], method: 'resolve', arg: [1,0]},
```
ystros: Not quite right - `card: cards[0]` and `card: cards[1]`, not `text`.

ystros:
@cryogen This one's good to go, but I've held off merging in case you have a preference about how the kneel character should be selected (right now via menu, could be done via a card select instead). Current flow is:

1. Select the 2 characters
2. Choose which character to kneel (via menu)
3. Other character is killed automatically.

This general flow is needed because the wording of the card does not guarantee a kill - you can select an unkillable character in step 1, then choose to kneel the only killable one in step 2. That's why it's structured as select 2 characters, then select which of those 2 to kneel. Only question is whether step 2 should be a menu or a card select (out of the 2 previously selected cards).

cryogen:
A card select prompt is probably a better, and more consistent experience - although you could argue there's already precedent that saves prompt via buttons...

PR #969, opened 2017-05-28
Implement Winterfell Crypt

DukeTax:


server/game/cards/locations/02/winterfellcrypt.js
```diff
@@ -0,0 +1,55 @@
...
+            }
+        });
+    }
+
+    uniqueStarkCharacterSacrificedOrKilled(event, card) {
+        return (
+            this.controller === card.controller &&
+            card.isUnique() &&
+            card.isFaction('stark') &&
+            card.getType() === 'character'
```
ystros: Winterfell Crypt also requires that the kill / sacrifice happened in specifically the challenge phase.

PR #971, opened 2017-05-29
Improve messaging for The Dornishman's Wife

DukeTax:
Instead of building up a long string, pushing "hits" to an array allows existing methods to take care of proper punctuation


ystros:
👍

PR #973, opened 2017-05-30
Implement Trial by Combat

samuellinde:


server/game/cards/events/02/trialbycombat.js
```diff
@@ -0,0 +1,36 @@
...
+                    challenge.challengeType === 'intrigue'
+                )
+            },
+            handler: context => {
+                let opponent = this.game.getOtherPlayer(this.controller);
+
+                context.skipHandler();
+
+                this.game.addMessage('{0} uses {1} to have {2} apply military claim instead of intrigue claim', this.controller, this, opponent);
+
```
cryogen: If you do:
this.game.addMessage('{0} uses {1} to have {2} apply {3} claim instead of {4} claim', this.controller, this, opponent, 'military', 'intrigue');

instead you'll get pretty pictures in chat

cryogen:
There's some formatting fixes needed to get the linting checks to pass

server/game/cards/events/02/trialbycombat.js
```diff
@@ -0,0 +1,36 @@
+const DrawCard = require('../../../drawcard.js');
+const ApplyClaim = require('../../../gamesteps/challenge/applyclaim.js');
+
+class TrialByCombat extends DrawCard {
+    setupCardAbilities() {
+        this.interrupt({
+            when: {
+                onClaimApplied: (event, challenge) => (
+                    challenge.winner === this.controller &&
+                    challenge.attackingPlayer === this.controller &&
```
ystros: The winner / attacking player restrictions aren't technically in the card text, but this is probably a reasonable change from a playability stand point (most of the time you wouldn't want to trigger on yourself, but there are edge cases). We can revisit it if we ever implement Melee or if there are complaints.

ystros:
Glad to see this one implemented. 👍

server/game/cards/events/02/trialbycombat.js
samuellinde: True – I guess it's a trade-off, having the interrupt on both players intrigue challenge claim would probably cause more annoyance.

PR #974, opened 2017-05-30
Implement Scaling the Wall

samuellinde:


server/game/cards/events/07/scalingthewall.js
```diff
@@ -0,0 +1,50 @@
...
+        this.reaction({
+            when: {
+                afterChallenge: (event, challenge) => (
+                    challenge.winner === this.controller &&
+                    challenge.attackingPlayer === this.controller &&
+                    this.hasAttackingWildling()
+                )
+            },
+            handler: () => {
+                this.game.promptForSelect(this.controller, {
```
DukeTax: Instead of a manual prompt, passing a target property into the reaction is less work and makes sure the reaction doesn't prompt when there are no legal targets. Something like this (it's explained in more depth in the card implementation docs):

            target: {
                activePromptTitle: 'Select location',
                cardCondition: card => card.getType() === 'location' && !card.isLimited();
            }

server/game/cards/events/07/scalingthewall.js
```diff
@@ -0,0 +1,50 @@
...
+                    activePromptTitle: 'Select location',
+                    source: this,
+                    onSelect: (player, card) => this.onCardSelected(player, card)
+                });
+            }
+        });
+    }
+
+    hasAttackingWildling() {
+        let cards = this.controller.filterCardsInPlay(card => {
```
DukeTax: This is more easily checked by doing:

    hasAttackingWildling() {
        return this.controller.anyCardsInPlay(card => this.game.currentChallenge.isAttacking(card) &&
                                                      card.hasTrait('Wildling') && 
                                                      card.getType() === 'character');
    }

server/game/cards/events/07/scalingthewall.js
```diff
@@ -0,0 +1,50 @@
...
+                afterChallenge: (event, challenge) => (
+                    challenge.winner === this.controller &&
+                    challenge.attackingPlayer === this.controller &&
+                    this.hasAttackingWildling()
+                )
+            },
+            handler: () => {
+                this.game.promptForSelect(this.controller, {
+                    cardCondition: card => this.cardCondition(card),
+                    activePromptTitle: 'Select location',
```
ystros: Instead of a manual `promptForSelect` here, it should be possible to use the `target` property. E.g.
```javascript
this.reaction({
    when {
        // ...
    },
    target: {
        cardCondition: card => this.cardCondition(card),
        activePromptTitle: 'Select location'
    },
    handler: context => {
        // the chosen card will be stored on context.target
    }
});
```

See [documentation](https://github.com/cryogen/throneteki/blob/master/docs/implementing-cards.md#choosing--targeting-cards) for more details (the examples given are for actions, but it works the same with interrupts and reactions)

server/game/cards/events/07/scalingthewall.js
ystros: Should be simpler to return the result of `this.controller.anyCardsInPlay` instead of filtering cards and then checking length (slowly trying to get rid of usages like this)

server/game/cards/events/02/trialbycombat.js
```diff
@@ -0,0 +1,36 @@
+const DrawCard = require('../../../drawcard.js');
+const ApplyClaim = require('../../../gamesteps/challenge/applyclaim.js');
+
+class TrialByCombat extends DrawCard {
```
ystros: Looks like you've inadvertently included commits from your previous PR. We typically use a feature-branch + squash-merge flow.

cryogen:
You'll need to clean this up before I can merge it, it has other people's commits in it.  My advice is to restore master back to original (using the guide I linked to you in discord), create a new branch, and then cherry pick commit 96a5868 onto that branch and then re-create the pull request against that branch.  Let me know in discord if none of this makes sense

PR #978, opened 2017-05-30
Implement Spears of the Merling King

DukeTax:


server/game/cards/locations/06/spearsofthemerlingking.js
```diff
@@ -0,0 +1,28 @@
+const DrawCard = require('../../../drawcard.js');
+
+class SpearsOfTheMerlingKing extends DrawCard {
+    setupCardAbilities(ability) {
+        this.interrupt({
+            when: {
+                onCharacterKilled: event => {
+                    if(event.card.controller === this.controller) {
+                        this.returnCard = event.card;
```
ystros: While I don't mind this pattern on non-merged events (where you'd need to do nasty things like `context.event.params[2]`), it feels unnecessary on merged events where you can just do `context.event.card`.

PR #991, opened 2017-05-31
Implement Tyrion's Chain

samuellinde:


server/game/cards/events/04/tyrionschain.js
```diff
@@ -0,0 +1,64 @@
...
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                afterChallenge: (event, challenge) => (
+                    challenge.winner === this.controller &&
+                    this.hasParticipatingUniqueLannister()
+                )
+            },
+            max: ability.limit.perPhase(1),
+            condition: () => this.game.anyPlotHasTrait('War'),
```
DukeTax: I think the condition property doesn't work on reactions/interrupts, just actions. So you can move this check to the event listener for the desired result.

PR #997, opened 2017-06-01
Implement Black Betha.

jbrz:
I think the strength calculation conditions could be a little cleaner, but otherwise a pretty straight-forward card to implement, nothing really new.

### DIFF
--- a/server/game/cards/locations/07/blackbetha.js
+++ b/server/game/cards/locations/07/blackbetha.js
@@ -8,6 +8,8 @@ class BlackBetha extends DrawCard {
         });
         this.action({
             title: 'Kneel to give attacking character +X STR',
+            condition: () => this.game.currentChallenge && this.calculateStrength() >= 1,
+            cost: ability.costs.kneelSelf(),
             target: {
                 activePromptTitle: 'Select character to gain STR',
                 cardCondition: card => this.game.currentChallenge.isAttacking(card)
--- a/server/game/cards/locations/07/blackbetha.js
+++ b/server/game/cards/locations/07/blackbetha.js
@@ -1,0 +1,57 @@
+const DrawCard = require('../../../drawcard.js');
+
+class BlackBetha extends DrawCard {
+    //Just used Iron Fleet Scout as template.  Overly verbose, but easy to understand.  Could be refactored into something more concise (a la The Valyrian).
+
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: card => card.name === 'Ser Davos Seaworth',
+            effect: ability.effects.addKeyword('renown')
+        });
+        this.action({
+            title: 'Kneel to give attacking character +X STR',
+            cost: ability.costs.kneelSelf(),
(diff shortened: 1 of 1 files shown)


server/game/cards/locations/07/blackbetha.js
```diff
@@ -0,0 +1,57 @@
...
+        this.action({
+            title: 'Kneel to give attacking character +X STR',
+            cost: ability.costs.kneelSelf(),
+            condition: () => this.game.currentChallenge && this.game.currentChallenge.attackers.length >= 1,
+            method: 'kneel'
+        });
+    }
+
+    kneel(player) {
+        this.game.promptForSelect(player, {
```
DukeTax: A bunch of older cards use these manual prompts still, but we've since switched to using a target property inside the action to handle cases like these, which would look something like this:

```javascript
            target: {
                activePromptTitle: 'Select character to gain STR',
                cardCondition: card => this.returnCondition(card)
            },
```
(See the card implementation docs for more information.)
This prevents the ability from initiating if there's no legal target, meaning you can get rid of the current condition check, and instead check if the defending player has at least 1 kneeling character (so the effect will actually change the game state, another requirement for it to be able to initiate).

server/game/cards/locations/07/blackbetha.js
```diff
@@ -0,0 +1,57 @@
...
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: card => card.name === 'Ser Davos Seaworth',
+            effect: ability.effects.addKeyword('renown')
+        });
+        this.action({
+            title: 'Kneel to give attacking character +X STR',
+            cost: ability.costs.kneelSelf(),
+            condition: () => this.game.currentChallenge && this.game.currentChallenge.attackers.length >= 1,
+            method: 'kneel'
```
DukeTax: I don't think this matters much, but I believe using a handler instead of a method is preferred for actions now.

server/game/cards/locations/07/blackbetha.js
```diff
@@ -1,53 +1,35 @@
...
 class BlackBetha extends DrawCard {
-    //Just used Iron Fleet Scout as template.  Overly verbose, but easy to understand.  Could be refactored into something more concise (a la The Valyrian).
-
     setupCardAbilities(ability) {
         this.persistentEffect({
             match: card => card.name === 'Ser Davos Seaworth',
             effect: ability.effects.addKeyword('renown')
         });
         this.action({
             title: 'Kneel to give attacking character +X STR',
```
DukeTax: Now you just need a cost (so that it actually kneels), and a condition (so that it can only initiate if it actually has the ability to change the game state. Something like this should work:

```javascript
            title: 'Kneel to give attacking character +X STR',
            condition: () => this.game.currentChallenge && this.calculateStrength() >= 1,
            target: {
```

jbrz:
My mistake, was in a rush this morning and wanted to get this pushed before I left for work, ended up clipping both of those (present in the previous commit haha).

PR #998, opened 2017-06-02
Convert legacy event cards to ability API.

ystros:


server/game/cards/events/01/olennascunning.js
```diff
@@ -1,35 +1,31 @@
...
-
-        this.game.promptWithMenu(this.game.currentChallenge.loser, this, {
-            activePrompt: {
-                menuTitle: 'Select a card type',
-                buttons: buttons
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                afterChallenge: (event, challenge) => (
+                    ['intrigue', 'power'].includes(challenge.challengeType) &&
```
cryogen: I dislike the way this reads, but not enough to be too bothered about it

ystros: I generally prefer this pattern to doing an explicit or since it's much more compact, but understand the readability pushback. I've been toying with adding an isMatch function to challenge so we can do things like this instead:
```javascript
challenge.isMatch({
    challengeType: ['intrigue', 'power'],
    winner: this.controller
})
```
which would be both more readable and more compact than what we're currently doing. But there are subtleties that have held me back from doing it, and it seems like by the time I get around to it we'll have pretty much all cards implemented. :woman_shrugging:

PR #999, opened 2017-06-02
Implement You Murdered Her Children.

ystros:


server/game/drawcard.js
```diff
@@ -137,12 +138,26 @@ class DrawCard extends BaseCard {
         });
     }
 
+    modifyStrengthMultiplier(amount, applying = true) {
```
cryogen: It sucks that this is necessary

PR #1012, opened 2017-06-04
Implement There Are No Men Like Me.

jbrz:


server/game/cards/events/07/therearenomenlikeme.js
```diff
@@ -0,0 +1,32 @@
...
+                        this.game.currentChallenge &&
+                        this.game.currentChallenge.challengeType === 'military'
+                    ),
+                    effect: [
+                        ability.effects.doesNotKneelAsAttacker(),
+                        ability.effects.doesNotKneelAsDefender()
+                    ]
+                }));
+
+                this.game.addMessage('{0} uses {1} to keep {2} standing when participating in military challenges this phase.', context.player, this, context.target);
```
DukeTax: This is super minor, but the character can still be knelt by card effects when participating during military challenges. Maybe something like "{0} uses {1} to have {2} not kneel when declared as an attacker or defender in military challenges until the end of the phase".

Also, commits from previous PR's seem to be included here.

jbrz:
Yeah whoops. Not sure why the previous PRs are there...I'll fix that and clarify the game message (good shout on that), cheers.

PR #1015, opened 2017-06-05
Implement Abandoned Stronghold

DukeTax:


server/game/cards/locations/07/abandonedstronghold.js
```diff
@@ -0,0 +1,40 @@
...
+                }));
+
+                this.game.addMessage('{0} kneels {1} to give {2} +{3} STR until the end of the challenge', 
+                                      context.player, this, context.target, strBoost);
+            }
+        });
+    }
+
+    getNumberOfBuilders() {
+        let cards = this.controller.filterCardsInPlay(card => {
```
ystros: FYI, you could just return `this.controller.getNumberOfCardsInPlay(card => ...)` instead, but this is less awkward than the `any` checks, so I'll let it through for now.

PR #1016, opened 2017-06-05
Implement I Shall Win No Glory

DukeTax:


ystros:
Kind of iffy on this one. I'm a bit surprised it doesn't allow you to retrigger the ability if you can pay the cost multiple times (e.g. 3 Rangers + 3 Builders) since no cancel is involved, but I tested it and it works correctly.

Between this and the previously implemented Pack Survives, it feels like we should have an 'or' or 'choose' cost. e.g.
```javascript
this.reaction({
    // ...
    cost: ability.costs.choose({
        'Kneel Builders': ability.costs.kneelMultiple(3, card => card.hasTrait('Builder') && card.getType() === 'character'),
        'Kneel Rangers': ability.costs.kneelMultiple(3, card => card.hasTrait('Ranger') && card.getType() === 'character'),
        'Kneel Stewards': ability.costs.kneelMultiple(3, card => card.hasTrait('Steward') && card.getType() === 'character')
    }),
    // ...
});
```

If the player can pay multiple costs, they're prompted with an additional prompt asking which cost they want to pay (unfortunately an additional click, but the buttons will have shorter / more focused text). If they can pay only a single cost, it automatically chooses it without a prompt.

Thoughts?

DukeTax:
I know you're asking cryogen, but I think that sounds good. Now that we can choose the order of triggers, each trigger only comprising 1 button is the most 'clean'. (I have a similar issue with the choices on Great Kraken for example, but that's another discussion.)

ystros:
#1034 will allow us to do the cost as I suggested above.

PR #1017, opened 2017-06-05
Implement Catapult on the Wall.

jbrz:
As noted (at least as far as I could figure out), there's not really a clean way to include kneeling the attached character as part of the cost of using the attachment action.  `ability.costs` needs another / updated kneel method to allow for this.  Certainly something I could do / we might consider, as I'm sure more costs like this are coming, but as I'm probably the least familiar with the code and gameflow, I figured I would just do the quick and dirty implementation.

### DIFF
--- a/server/game/cards/attachments/07/catapultonthewall.js
+++ b/server/game/cards/attachments/07/catapultonthewall.js
@@ -1,0 +1,40 @@
+const DrawCard = require('../../../drawcard.js');
+
+class CatapultOnTheWall extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kneel Catapult and attached character',
+            condition: () => this.game.currentChallenge,
+
+            // This is not a nice interaction for users, but `ability.costs` would need to be updated with a different method to allow something like attachment.parent.kneel() (I think)
+            cost: [
+                ability.costs.kneelSelf(),
+                ability.costs.kneel(card => card === this.parent)
+            ],
+            target: {
+                activePromptTitle: 'Select character to kill',
+                cardCondition: card => this.game.currentChallenge.isAttacking(card)
+            },
+            handler: context => {
+                context.target.owner.killCharacter(context.target);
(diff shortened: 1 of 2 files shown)


server/game/cards/attachments/07/catapultonthewall.js
```diff
@@ -0,0 +1,40 @@
...
+class CatapultOnTheWall extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kneel Catapult and attached character',
+            condition: () => this.game.currentChallenge,
+
+            // This is not a nice interaction for users, but `ability.costs` would need to be updated with a different method to allow something like attachment.parent.kneel() (I think)
+            cost: [
+                ability.costs.kneelSelf(),
+                ability.costs.kneel(card => card === this.parent)
```
ystros: If you pull in this commit, you should be able to get rid of the weird interaction:
https://github.com/ystros/throneteki/commit/f998dd75cc01a80e40338dc692c24a5296cbd4a4

server/game/cards/attachments/07/catapultonthewall.js
```diff
@@ -0,0 +1,40 @@
...
+            ],
+            target: {
+                activePromptTitle: 'Select character to kill',
+                cardCondition: card => this.game.currentChallenge.isAttacking(card)
```
ystros: The card text also requires that the card have strength 4 or less.

jbrz: oh right, whoops.  Got so caught up with trying to figure out the kneel that I forgot about making sure the card was correct, apologies.

server/game/cards/attachments/07/catapultonthewall.js
ystros: I apologize for not being clearer, you should also switch this to `ability.costs.kneelParent()` now that you've pulled in the cost implementation.

jbrz: Ah nope, that is clear in the code you pushed, I again overlooked it.

ystros:
👍

PR #1022, opened 2017-06-06
Implement Wheels Within Wheels

DukeTax:
Took a stab at this weird card, couple of notes:

* Cards are dumped in the discard pile before being "officially" revealed through a chat message. This is not ideal, though to get the ever-decreasing prompt we want for effects like these I needed to move them somewhere, and discard pile makes the most sense here to me.

* The prompt does not show duplicates. That is preferable for a card like Summons, a bit less so for this one. If you hit a duplicate, you won't know it at first, though you find out after selecting them (since the card will just keep showing in the prompt until all duplicates have been moved to the discard pile). Not a huge deal though.

* I'm quite sure that if you would manage to hit 10 events, it won't resolve correctly since you won't get the chance to choose Done. Increasing numToSelect to 11 does not solve this, as that triggers a whole new search of 10 more cards. This is an extremely unlikely occurrence, and it should work correctly for all other use cases.

* It is quite verbose, but I like the extra checks to minimize clicks and get the messages just right.

### DIFF
--- a/server/game/cards/plots/06/wheelswithinwheels.js
+++ b/server/game/cards/plots/06/wheelswithinwheels.js
@@ -1,0 +1,85 @@
+const _ = require('underscore');
+
+const PlotCard = require('../../../plotcard.js');
+
+class WheelsWithinWheels extends PlotCard {
+    setupCardAbilities() {
+        this.whenRevealed({
+            handler: () => {
+                this.game.promptForDeckSearch(this.controller, {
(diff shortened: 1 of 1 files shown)


cryogen:
I can't help but think there's a better way to do this, this card makes my head hurt

ystros:
I second the head hurting feeling, but I don't have a better solution, so merging.

PR #1028, opened 2017-06-07
Implement The White Shadows

DukeTax:


server/game/cards/plots/07/thewhiteshadows.js
```diff
@@ -0,0 +1,28 @@
...
+            handler: context => {
+                this.controller.putIntoPlay(context.event.card);
+                this.game.addMessage('{0} uses {1} to put {2} into play under their control, blank', 
+                                      this.controller, this, context.event.card);
+
+                this.untilEndOfPhase(ability => ({
+                    match: context.event.card,
+                    effect: [
+                        ability.effects.blank,
+                        ability.effects.moveToDeadPileIfStillInPlay()
```
ystros: This should probably be moved to a separate `atEndOfPhase` effect to match the card's text (not sure if there are scenarios where it could make a difference since the card is also blank, but there may be)

DukeTax: Hmm not sure. The character is not blank until the end of the phase, it's blank as long as it is in play. Wouldn't linking the blanking to the moving to dead pile match the card text the most that way?

ystros: Ah, yes, you are correct in that they should be linked. I think the effect duration should still be `atEndOfPhase` though, instead of `untilEndOfPhase` to match card wording. It reads a bit weird saying 'at end of phase the card is blank' but I believe it should work properly - the card will be blank *until* the 'at end of phase' event fires.

DukeTax: I sorta misread your first post; atEndOfPhase makes more sense here. :)

PR #1032, opened 2017-06-09
Implement Vaes Dothrak

DukeTax:


server/game/costs.js
```diff
@@ -358,20 +358,26 @@ const Costs = {
         };
     },
     /**
-     * Cost that requires discarding a card from hand.
+     * Cost that requires discarding a card from hand matching the passed
+     * condition predicate function.
      */
-    discardFromHand: function() {
+    discardFromHand: function(condition) {
```
ystros: It would be preferable to default the condition here than update the existing usages. i.e.

```javascript
discardFromHand: function(condition = () => true) {
```

PR #1034, opened 2017-06-10
Implement `choose` cost.

ystros:
Implements a choose cost that allows the player to choose which cost
they want to pay on a card. Useful for cards like The Pack Survives and
I Shall Win No Glory which give players an 'or' choice. If only one cost
is able to be paid, it will pick that cost automatically. If multiple
costs are payable, the player will be prompted to choose which they wish
to pay.


cryogen:
👍

PR #1036, opened 2017-06-10
Lady Sansa's Rose: pass challenge object around to avoid it's null

zacchiro:
Fixes THRONETEKI-VN


cryogen:
As discussed on discord i dont think this is the correct fix here, this should be fixed by a later PR

zacchiro:
ack

PR #1043, opened 2017-06-12
Implement Spearmaiden

DukeTax:


server/game/gamesteps/challenge/fulfillmilitaryclaim.js
```diff
@@ -27,11 +36,27 @@ class FulfillMilitaryClaim extends BaseStep {
...
+        return this.forcedClaim.includes(card);
+    }
+
     fulfillClaim(p, cards) {
         if(!_.isArray(cards)) {
             cards = [cards];
         }
 
+        if(this.forcedClaim.length < this.claim) {
+            cards.push(this.forcedClaim);
```
ystros: This pushes the `forcedClaim` array onto the `cards` array, leading `cards` to look like:

```
[ character1, [character2] ]
```

This works fine when outputting the message, but when looping through characters to kill, it sees one card and one array. I'm a bit surprised it doesn't just outright crash.

You should either loop through each card in `forcedClaim` and push them individually onto `cards`, or just set `cards` to the concatted result of it and `forcedClaim`.

server/game/gamesteps/challenge/fulfillmilitaryclaim.js
```diff
@@ -10,15 +10,24 @@ class FulfillMilitaryClaim extends BaseStep {
     }
 
     continue() {
-        var promptMessage = 'Select ' + this.claim + ' ' + (this.claim > 1 ? 'characters' : 'character') + ' to fulfill military claim';
+        this.forcedClaim = _.filter(this.player.mustChooseAsClaim, card => card.controller === this.player && card.location === 'play area');
+        
+        let claimToSelect = this.claim;
+        
+        if(this.forcedClaim.length < this.claim) {
+            claimToSelect = this.claim - this.forcedClaim.length;
```
ystros: If there is 0 claim left to select, it should kill the characters immediately. Right now it will prompt the player to select 0 characters.

It would also probably be useful to add a message saying that they were forced to choose certain characters so that both players can follow what's going on.

server/game/gamesteps/challenge/fulfillmilitaryclaim.js
```diff
@@ -10,15 +10,24 @@ class FulfillMilitaryClaim extends BaseStep {
...
+        }
+
+        var promptMessage = 'Select ' + claimToSelect + ' ' + (claimToSelect > 1 ? 'characters' : 'character') + ' to fulfill military claim';
```
ystros: Since this variable is being moved and others are being declared in this method, `let` instead of `var`.

server/game/gamesteps/challenge/fulfillmilitaryclaim.js
DukeTax: I'm specifically checking for '<' instead of '<=', so claim to select will always be >0. This does mean that in the most common use case (1 Spearmaiden trigger and 1 claim) the player still gets a normal claim prompt, he'll just have a single option to choose. I could be convinced to skip the prompt whenever this.forcedClaim.length === this.claim, it's basically choosing less clicks versus maintaining clarity of what's going on. <@cryogen 

Another message is useful in any case though.

server/game/gamesteps/challenge/fulfillmilitaryclaim.js
DukeTax: Ugh, noob mistake. Thanks.

ystros:
Could you get the tests for `FulfillMilitaryClaim` passing, even if you don't add tests specific for forced claim? Should be as simple as setting `mustChooseAsClaim` to an empty array on the player spy object.

PR #1044, opened 2017-06-12
Implement Stonesnake

DukeTax:
Opted to only potentially prompt for keywords that do something.

### DIFF
--- a/server/game/cards/characters/03/stonesnake.js
+++ b/server/game/cards/characters/03/stonesnake.js
@@ -2,6 +2,8 @@ const _ = require('underscore');
 
 const DrawCard = require('../../../drawcard.js');
 
+const keywords = ['Insight', 'Intimidate', 'Pillage', 'Renown'];
+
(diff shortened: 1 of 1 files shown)


server/game/cards/characters/03/stonesnake.js
```diff
@@ -0,0 +1,64 @@
...
+        }));
+
+        this.game.addMessage('{0} uses {1} to have {1} gain {2} until the end of the phase', 
+                              this.controller, this, keyword);
+
+        return true;
+    }
+
+    hasNoCopyableKeyword(card) {
+        if(!card.hasKeyword('insight') &&
```
ystros: Some slight repetition between here and the `keywords` const in the handler. May be worth pulling that const outside the class, then using it in both places. You could then replace this method with:
```javascript
hasNoCopyableKeyword(card) {
    return !_.any(keywords, keyword => card.hasKeyword(keyword));
}
```

PR #1046, opened 2017-06-13
add effect cannotBeDiscarded and use it to immunize The God's Eye

zacchiro:
Note that this assumes that all possibilities of discarding The God's Eye from
play are via the engine effect. As far as I understand this should be the case
right now, but @ystros might want to double-check that.

### DIFF
--- a/server/game/cards/locations/04/thegodseye.js
+++ b/server/game/cards/locations/04/thegodseye.js
@@ -1,8 +1,12 @@
 const DrawCard = require('../../../drawcard.js');
 
 class TheGodsEye extends DrawCard {
-    setupCardAbilities() {
-        // TODO: Cannot be discarded.
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: this,
+            effect: ability.effects.cannotBeDiscarded()
+        });
+
         this.plotModifiers({
             reserve: 1,
             gold: 1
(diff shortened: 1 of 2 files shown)


server/game/cards/locations/04/thegodseye.js
```diff
@@ -1,8 +1,12 @@
 const DrawCard = require('../../../drawcard.js');
 
 class TheGodsEye extends DrawCard {
-    setupCardAbilities() {
-        // TODO: Cannot be discarded.
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: this,
+            effect: ability.effects.cannotBeDiscarded(context => context.stage === 'effect')
```
ystros: The card text indicates that it simply cannot be discarded from play, not that it cannot be discarded by card effects. Should this be just `cannotBeDiscarded()`?

zacchiro:
On Tue, Jun 13, 2017 at 10:58:02AM -0700, Brian Upton wrote:
> +            effect: ability.effects.cannotBeDiscarded(context => context.stage === 'effect')
> 
> The card text indicates that it simply cannot be discarded from play,
> not that it cannot be discarded by card effects. Should this be just
> `cannotBeDiscarded()`?

Yes, totally :-)
Fixed.

Thanks!

PR #1053, opened 2017-06-14
Implement Northern Refugee.

ystros:
* Implements the ability to have an effect be active from 'any' location. I wanted to support other specific locations (e.g. while a card is in discard or dead pile), but there are no such cards right now and the change required was exponentially more complex. So for now, it'll just throw an exception if you do an effect with an unsupported location.
* Implements an effect to reduce the cost of the current card.
* Implements Northern Refugee's self-reduction ability.


server/game/basecard.js
```diff
@@ -184,7 +184,19 @@ class BaseCard {
...
+        const allowedLocations = ['active plot', 'agenda', 'any', 'play area'];
+        const defaultLocationForType = {
+            agenda: 'agenda',
+            plot: 'active plot'
+        };
+
+        let location = properties.location || defaultLocationForType[this.getType()] || 'play area';
+
+        if(!allowedLocations.includes(location)) {
+            throw new Error(`'${location}' is not a supported effect location.`);
```
cryogen: Huh.  Did not know this was a thing

server/game/gamesteps/keywordwindow.js
```diff
@@ -0,0 +1,53 @@
+const _ = require('underscore');
```
cryogen: This is cool, but did you mean to include it in this PR?

cryogen:
Sorry for taking ages to review this!

server/game/gamesteps/keywordwindow.js
ystros: Whoops, yeah, unrelated WIP. Removed.

server/game/basecard.js
ystros: Yup. I think this is the first place in the server code base but I used it for error messages in the test framework. They're handy.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

cryogen: :+1: 

Yeah, C# has those, quite useful indeed.

I'm sure you do already but be mindful of https://kangax.github.io/compat-table/es6/

Anything which is in node 6.5 or higher is fine for the server/tests.

Client can use anything in the bable column because it's transpiled

Issue #1048, opened 2017-06-13
Annals of Castle Black


uBaHoB:
Events can be used from discard as action without problem also when discarded for claim they go out of game which is perfect.

The problem lies with the reactions. It appears the game doesn't detects the events in discard pile. I had to move them to hand so I can trigger them.

Cheers,
uBaH

PR #1055, opened 2017-06-14
Implement onChallengeInitiated event

### LINKED ISSUE #1047: Initiate challenge timing issue
When initiating a challenge, a couple of different steps happen at once:

RRG:

> To initiate a challenge, the active player does the following. The game treats these criteria as being performed atomically (all at once), and they must all be completed together in order to initiate a challenge.

>* Announces the type of challenge ([military], [intrigue], or [power]) being initiated.
>* Announces the opponent against whom the challenge is being initiated.
>* Announces which characters (under his or her control) are being declared as attackers, and kneels them simultaneously. In order to be eligible as an attacker, a character must have a challenge icon that corresponds with the type of challenge being initiated, and the character must be standing. At least one character must be declared as an attacker to initiate a challenge. Any character with the stealth keyword that has been declared as an attacker also chooses its stealth targets at this time if its controller desires to use the stealth keyword.

To make sure abilities get triggered at the right time, the following events need to be raised simultaneously:

`onChallenge`
`onAttackersDeclared`
`onBypassedByStealth`

DukeTax:
Fixes #1047

### DIFF
--- a/server/game/cards/characters/01/jonsnow.js
+++ b/server/game/cards/characters/01/jonsnow.js
@@ -9,7 +9,8 @@ class JonSnow extends DrawCard {
         this.registerEvents(['onChallengeInitiated']);
     }
 
-    onChallengeInitiated(event, challenge) {
+    onChallengeInitiated(event) {
+        let challenge = event.challenge;
         if(this.kneeled || this.controller !== challenge.attackingPlayer) {
             return;
         }
--- a/server/game/cards/characters/01/jonsnow.js
+++ b/server/game/cards/characters/01/jonsnow.js
@@ -6,10 +6,10 @@ class JonSnow extends DrawCard {
     constructor(owner, cardData) {
         super(owner, cardData);
 
-        this.registerEvents(['onAttackersDeclared']);
+        this.registerEvents(['onChallengeInitiated']);
     }
 
-    onAttackersDeclared(event, challenge) {
+    onChallengeInitiated(event, challenge) {
         if(this.kneeled || this.controller !== challenge.attackingPlayer) {
             return;
         }
--- a/server/game/cards/characters/02/seralliserthorne.js
+++ b/server/game/cards/characters/02/seralliserthorne.js
@@ -11,7 +11,7 @@ class SerAlliserThorne extends DrawCard {
         this.reaction({
             location: 'hand',
             when: {
-                onChallenge: (event, challenge) => challenge.challengeType === 'military'
+                onChallengeInitiated: event => event.challenge.challengeType === 'military' && event.challenge.defendingPlayer === this.controller
             },
             cost: [
                 ability.costs.kneelFactionCard(),
--- a/server/game/cards/characters/03/wolvesofthenorth.js
+++ b/server/game/cards/characters/03/wolvesofthenorth.js
@@ -4,22 +4,19 @@ class WolvesOfTheNorth extends DrawCard {
     setupCardAbilities() {
         this.reaction({
             when: {
-                onBypassedByStealth: (event, challenge, source, target) => {
-                    if(source !== this) {
-                        return false;
-                    }
-
-                    this.bypassed = target;
-
-                    return true;
-                }
+                onChallengeInitiated: event => event.challenge.isStealthSource(this)
             },
-            handler: () => {
-                this.game.addMessage('{0} uses {1} to give {2} a penalty to STR', this.controller, this, this.bypassed);
+            handler: context => {
+                let target = context.event.challenge.getStealthTargetFor(this);
+                let strDecrease = -this.controller.getNumberOfCardsInPlay(c => c.hasTrait('Direwolf'));
+
                 this.untilEndOfPhase(ability => ({
-                    match: this.bypassed,
-                    effect: ability.effects.dynamicStrength(() => -this.controller.getNumberOfCardsInPlay(c => c.hasTrait('Direwolf')))
+                    match: target,
+                    effect: ability.effects.modifyStrength(strDecrease)
                 }));
+
+                this.game.addMessage('{0} uses {1} to give {2} {3}STR until the end of the phase', 
+                                      this.controller, this, target, strDecrease);
             }
         });
     }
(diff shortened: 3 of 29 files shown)


server/game/cards/characters/01/jonsnow.js
```diff
@@ -6,10 +6,10 @@ class JonSnow extends DrawCard {
     constructor(owner, cardData) {
         super(owner, cardData);
 
-        this.registerEvents(['onAttackersDeclared']);
+        this.registerEvents(['onChallengeInitiated']);
     }
 
-    onAttackersDeclared(event, challenge) {
+    onChallengeInitiated(event, challenge) {
```
ystros: Since this is now a merged event, `challenge` doesn't exist as a second param, so it should be removed. Need to update usages from `challenge` to `event.challenge` in the method (or simply define `let challenge = event.challenge` as the first line in the method to reduce change noise)

server/game/cards/characters/02/seralliserthorne.js
```diff
@@ -11,7 +11,7 @@ class SerAlliserThorne extends DrawCard {
         this.reaction({
             location: 'hand',
             when: {
-                onChallenge: (event, challenge) => challenge.challengeType === 'military'
+                onChallengeInitiated: event => event.challenge.challengeType === 'military' && event.challenge.defendingPlayer === this.controller
```
ystros: Good catch. 👍

server/game/cards/characters/03/wolvesofthenorth.js
```diff
@@ -4,22 +4,19 @@ class WolvesOfTheNorth extends DrawCard {
...
-                this.game.addMessage('{0} uses {1} to give {2} a penalty to STR', this.controller, this, this.bypassed);
+            handler: context => {
+                let target = context.event.challenge.getStealthTargetFor(this);
+                let strDecrease = -this.controller.getNumberOfCardsInPlay(c => c.hasTrait('Direwolf'));
+
                 this.untilEndOfPhase(ability => ({
-                    match: this.bypassed,
-                    effect: ability.effects.dynamicStrength(() => -this.controller.getNumberOfCardsInPlay(c => c.hasTrait('Direwolf')))
+                    match: target,
+                    effect: ability.effects.modifyStrength(strDecrease)
```
ystros: Also good catch - value should be fixed at the time of the ability, not dynamic. 👍

server/game/cards/characters/07/maryaseaworth.js
```diff
@@ -4,16 +4,20 @@ class MaryaSeaworth extends DrawCard {
     setupCardAbilities(ability) {
         this.reaction({
             when: {
-                onBypassedByStealth: (event, challenge, source, target) => {
-                    this.bypassed = target;
-                    return true;
+                onChallengeInitiated: event => {
```
ystros: This implementation is no longer correct - previously, if 2 characters were bypassed by stealth, 2 events would be fired and Marya would trigger twice. However, now she will only trigger once since onChallengeInitiated is only fired once. This... makes things very difficult, given that she would need to pay 1 gold for each character, and the limit is now associated with challenges being initiated, not characters bypassed.

server/game/gamesteps/challenge/challengeflow.js
```diff
@@ -67,17 +67,16 @@ class ChallengeFlow extends BaseStep {
...
-            this.game.raiseEvent('onAttackersDeclared', this.challenge);
-        });
     }
 
     announceAttackerStrength() {
         // Explicitly recalculate strength in case an effect has modified character strength.
         this.challenge.calculateStrength();
-        this.game.addMessage('{0} has initiated a {1} challenge with strength {2}', this.challenge.attackingPlayer, this.challenge.challengeType, this.challenge.attackerStrength);
+
+        this.game.raiseMergedEvent('onChallengeInitiated', { challenge: this.challenge }, () => {
```
ystros: Per the Lord of the Crossing test failures, initiating the challenge and then recalculating / announcing the strength of the challenge need to be 2 separate steps, like they were before.

ystros:
Unfortunately, the more I think about it the more I think combining the events is the wrong approach. It would have worked if not for Marya, but even then it has increased the complexity of some of the card implementations (if only slightly). My fault for suggesting we go down this path. 🙁 

The better solution would be to have a special event window where multiple events can appear within the same prompt. So a prompt that groups `onChallengeInitiated`, `onAttackersDeclared` and individual `onBypassedByStealth` interrupts / reactions. This is a lot more complex than this change, but the complexity will be inside the engine.

I don't think we should close this PR quite yet, though, as it provides a good base to build off of. It:
* Renames the ambiguous `onChallenge` event to `onChallengeInitiated`.
* Converts events to the new merged style.
* Records stealth choices, which is necessary if the event isn't fired immediately.
* Fixes several card implementation bugs.

DukeTax:
Agreed and no problem, hard to predict these things. :) I do like how the stealth cards look now, although that's more because of raising a merged event now, not because of combining the previously separate events.

Am I correct in assuming this is something you'll be working further on from now?

ystros:
Yeah, I'll work on it (unless you want to take a crack at this horribleness). But I'm probably going to tackle #1048 first since that card is 50% broken at the moment.

ystros:
As an update, I have this working on this branch: https://github.com/cryogen/throneteki/compare/master...ystros:simultaneous-challenge-events?expand=1

I'm waiting to PR it because:
1. naming is starting to get really bad. There is now both a 'simultaneous' event window and a 'combined' event window, which do somewhat different things. I'll either rename them to something more distinct, combine them, or introduce them as different event types instead of windows prior to PR.
2. I'd rather cryo spend time reviewing the easier and more critical bug fixes + stuff related to the new pack when he gets back than the in-depth review needed for this change.

ystros:
#1146 implements just the merged-event conversion, rename and misc bug fixes. This should make it easier to review the bigger change when it comes (less noise since the cards will already be in the correct form).

Closing this now to make it clear that this isn't the final PR that should be reviewed.

PR #1058, opened 2017-06-16
Implement Summer (WotW)

DukeTax:
The 'attach to a different character' handler is a little bit awkward, but I didn't want to just invoke player.attach() since that raises onCardEntersPlay, or player.putIntoPlay(), which would prompt for a parent where the old parent would be made a legal choice (and invokes player.attach() itself anyway).

### DIFF
--- a/server/game/cards/attachments/07/summer.js
+++ b/server/game/cards/attachments/07/summer.js
@@ -1,0 +1,58 @@
+const DrawCard = require('../../../drawcard.js');
+
+class Summer extends DrawCard {
+
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Have parent participate',
+            condition: () => this.game.currentChallenge && this.game.currentChallenge.challengeType === 'military',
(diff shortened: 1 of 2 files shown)


server/game/cards/attachments/07/summer.js
```diff
@@ -0,0 +1,58 @@
...
+                cardCondition: card => (
+                    card.location === 'play area' &&
+                    card !== this.parent &&
+                    this.canAttach(this.controller, card))
+            },
+            handler: context => {
+                this.controller.removeCardFromPile(this);
+                this.parent = context.target;
+                this.moveTo('play area');
+                context.target.attachments.push(this);
```
ystros: I think a better solution would be to modify `Player.attach` to just not raise 'onCardEntersPlay' if the card is already in play, rather than replicate the code here.

DukeTax: That... makes a lot more sense. :)

PR #1060, opened 2017-06-16
Implement Aeron Damphair (TIMC)

DukeTax:


server/game/event.js
```diff
@@ -27,8 +27,10 @@ class Event {
...
         }
 
-        card.markAsSaved();
-        this.cards = _.reject(this.cards, c => c === card);
+        card.game.raiseMergedEvent('onCharacterSaved', { card: card }, () => {
+            card.markAsSaved();
+            this.cards = _.reject(this.cards, c => c === card);
+        });
 
         if(_.isEmpty(this.cards)) {
```
ystros: This `if` needs to be inside the event handler, because removing the card is no longer immediate.

server/game/event.js
```diff
@@ -27,8 +27,10 @@ class Event {
             return;
         }
 
-        card.markAsSaved();
-        this.cards = _.reject(this.cards, c => c === card);
+        card.game.raiseMergedEvent('onCharacterSaved', { card: card }, () => {
```
ystros: I would rename this to just `onCardSaved` because we should also have the event fire when saving locations / attachments (even if it won't be used yet). There's little value having two separate events.

You also need to look at raising this when a dupe is used to save a card. There also may be other event handlers that do not use `Event.saveCard` - look for any handlers for onCharacterKilled that do just an `event.cancel()` instead (there may or may not be any)

server/game/cards/characters/04/aerondamphair.js
```diff
@@ -0,0 +1,26 @@
+const DrawCard = require('../../../drawcard.js');
+
+class AeronDamphair extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCharacterSaved: () => true
```
ystros: Per my event rename suggestion, you should check that the card being saved is a character.

DukeTax:
Overdue update: for some reason this makes dupes not work anymore vs Valar. I can't really figure out why and it's been a while so I'm closing this for now.

PR #1063, opened 2017-06-17
require Stealth and Intimidate targets to be in play

zacchiro:


zacchiro:
This was originally a fix only for this bug:

screenshot from 2017-06-17 11-09-41

where you can see Intimidate allowing to select characters from the discard and dead piles.

But then I looked back at 3d0f64b500c72f0fb96a2ebc754e652bb1d20059, where DukeTax fixed the issue of prompting for stealth when there are no target in play. That was a required fix. But still, when there are valid stealth targets in play, it looks like the current code will allow to select among both in play and out of play targets. This PR should also fix that issue.

PR #1064, opened 2017-06-17
improve and uniform game messages for all tutoring cards

zacchiro:
Before this change, when no card was found (or chosen) by the user, tutoring
completed with messages "FOO does not use BAR to ...", which is misleading
given the player did use the card, but did not find (or choose) any card. With
this change this is replaced by message like "FOO uses BAR, but does not ...".

Also, this change uniform tutoring so that there is a single message about
tutoring outcome, rather than more than once for some weird cards.

Finally, this change also drops the explicit use of the verb "reveal", which
was redundant given that tutoring success messages always name cards,
"revealing" them to the opponent already.


ystros:
👍

PR #1068, opened 2017-06-18
Implement Burning on the Sand

### LINKED ISSUE #1062: Burning on the Sand
The reaction on this card is not prompted by automation.

DukeTax:
* Fixes #1062 
* This has some wonky interactions with the claim modification buttons in the client, as illustrated by the pasted game chat below. Nothing game-breaking though.

```
user1 is initiating a  challenge
user1 has initiated a  challenge with strength 2
DukeTax does not defend the challenge
user1 uses Winter Is Coming to raise the claim value on their revealed plot card by 1 until the end of the challenge
user1 won a  challenge 2 vs 0
DukeTax plays Burning on the Sand to set the claim value on user1's revealed plot card to 0 until the end of the challenge
//Here I press the + Claim button three times. Claim remains 0 on the indicator, though chat messages don't correspond
user1 sets claim to 2 (+1)
user1 sets claim to 3 (+1)
user1 sets claim to 4 (+1)
user1 has gained 1 power from an unopposed challenge
The claim value for  is 0
//After the challenge the claim indicator rose to 4
```

### DIFF
--- a/server/game/cards/events/04/burningonthesand.js
+++ b/server/game/cards/events/04/burningonthesand.js
@@ -1,0 +1,26 @@
+const DrawCard = require('../../../drawcard.js');
+
+class BurningOnTheSand extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                afterChallenge: (event, challenge) => challenge.loser === this.controller && challenge.isUnopposed()
+            },
+            handler: () => {
+                this.untilEndOfChallenge(ability => ({
+                    match: card => card === card.controller.activePlot,
+                    targetController: 'opponent',
+                    effect: ability.effects.setClaim(0)
+                }));
+
+                let opponent = this.game.getOtherPlayer(this.controller);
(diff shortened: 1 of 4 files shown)


ystros:
I think the wonkiness when manually modifying claim is fine, but I'll defer to @cryogen. I'm not sure how you'd make it work otherwise TBH because 'set' is supposed to ignore all modifiers, which is how the manual stuff all works.

cryogen:
changeStat tries to be efficient and just indexes the player object to find the 'stat' that has changed.  It should probably be changed to be slightly more long winded and call the appropriate getFoo() functions.

I'm happy to have that done as a separate task, though.

cryogen:
Though, having reflected on it for more than 5 minutes, the buttons should override card behaviour, in case of bugs/misfeatures.  So technically this PR is wrong to that behaviour.  Hmm.

DukeTax:
This is a super quick solution.

cryogen:
That will work for now, but I think there's a bigger piece of work here to tidy it up

PR #1069, opened 2017-06-19
do not prompt for Paid Off and Brothel Madame if gold is 0

zacchiro:


server/game/cards/attachments/02/paidoff.js
```diff
@@ -8,10 +8,14 @@ class PaidOff extends DrawCard {
             },
             handler: () => {
                 this.parent.controller.kneelCard(this.parent);
-
                 this.game.addMessage('{0} uses {1} to kneel {2}', this.controller, this, this.parent);
 
-                this.game.promptWithMenu(this.parent.controller, this, {
+                let player = this.parent.controller;
+                if(player.gold < 1) {
+                    this.cancel(player);
```
ystros: Need a return in this `if`, otherwise it'll message saying they didn't pay then prompt them anyway.

zacchiro: Nice catch, thanks!
Added.

PR #1071, opened 2017-06-19
Fix Old Wyk

DukeTax:
* Previously, Old Wyk made use of a lasting effect until the end of the challenge to resolve its effect. This was incorrect as its ability should resolve after the winner is determined, but before any reactions, similar to Dolorous Edd or Spearmaiden.
* Checking if there's any Drowned God characters in the dead pile before prompting to trigger the reaction prevents triggering it without any change in game state.

### DIFF
--- a/server/game/cards/locations/05/oldwyk.js
+++ b/server/game/cards/locations/05/oldwyk.js
@@ -21,7 +21,7 @@ class OldWyk extends DrawCard {
                 this.game.addMessage('{0} kneels {1} to put {2} into play from their dead pile as an attacker', 
                                       this.controller, this, card);
 
-                this.game.once('afterChallenge:interrupt', (event, challenge) => this.resolveAfterChallenge(challenge, card));
+                this.game.once('afterChallenge', (event, challenge) => this.resolveAfterChallenge(challenge, card));
             }
         });
     }
--- a/server/game/cards/locations/05/oldwyk.js
+++ b/server/game/cards/locations/05/oldwyk.js
@@ -7,29 +7,41 @@ class OldWyk extends DrawCard {
         this.reaction({
             when: {
                 onChallenge: (event, challenge) =>
-                    challenge.attackingPlayer === this.controller
-                    && challenge.challengeType === 'power'
+                    challenge.attackingPlayer === this.controller &&
+                    challenge.challengeType === 'power' &&
+                    this.anyDrownedGodInDeadPile()
             },
             cost: ability.costs.kneelSelf(),
             handler: () => {
-                var card = _.last(this.controller.deadPile.filter(c => c.hasTrait('Drowned God')));
-
-                if(!card) {
-                    this.game.addMessage('{0} finds no character in their dead pile to put into play with {1}', this.controller, this);
-                    return;
-                }
+                let card = _.last(this.controller.deadPile.filter(c => c.hasTrait('Drowned God')));
 
                 this.controller.putIntoPlay(card);
                 this.game.currentChallenge.addAttacker(card);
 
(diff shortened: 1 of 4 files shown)


server/game/cards/locations/05/oldwyk.js
```diff
@@ -7,30 +7,42 @@ class OldWyk extends DrawCard {
...
 
-                this.game.addMessage('{0} uses {1} to put into play {2} as an attacker from their dead pile', this.controller, this, card);
+                this.game.addMessage('{0} kneels {1} to put {2} into play from their dead pile as an attacker', 
+                                      this.controller, this, card);
 
-                this.untilEndOfChallenge(ability => ({
-                    match: card,
-                    effect: ability.effects.returnToHandOrDeckBottom()
-                }));
+                this.game.once('afterChallenge:interrupt', (event, challenge) => this.resolveAfterChallenge(challenge, card));
```
ystros: The ThronesDB page is a bit unclear whether this should also happen before forced interrupts and "would" cancellation interrupts. My guess is the timing should be bullet 1 on the rules for "Triggering Condition." https://thronesdb.com/rulesreference#Triggering_Condition

In which case, this should listen to 'afterChallenge:cancelinterrupt' which will execute *just* before prompts for cancellation interrupts fire. But I'm not 100% sure.

server/game/effects.js
```diff
@@ -787,33 +787,6 @@ const Effects = {
                 player.mustChooseAsClaim = _.reject(player.mustChooseAsClaim, c => c === card);
             }
         };
-    },
-    /**
-     * Effects specifically for Old Wyk.
-     */
-    returnToHandOrDeckBottom: function() {
```
ystros: Good riddance. 👍

server/game/cards/locations/05/oldwyk.js
DukeTax: > 2. This is purely hypothetical, but during the D of DUCK the following abilities trigger; what is the proper order?

The Red Wedding-Lord of the Crossing-Old Wyk-Spearmaiden-Superior Claim.

> 2) 
Interrupts - The Red Wedding
Delayed effects - Lord of the Crossing / Old Wyk / Spearmaiden (first player decides)
Reactions: Superior Claim

I used afterChallenge:interrupt to mirror Spearmaiden/Dolorous Edd, but if afterChallenge:cancelinterrupt is correct I'll change them all.

ystros: Hmm. That seems in contradiction to the FAQ for Old Wyk: http://thronesdb.com/card/05028
> The "If you win the challenge..." effect resolves before any Interrupts or Reactions to winning/losing the challenge can be played.

If it happens between Interrupts and Reactions as you stated, then Old Wyk was correct before (the effect should be reprocessed just before Reactions). If the FAQ is correct, it's a question of whether it should happen before all Interrupts (including forced and cancellation interrupts) or just before player triggered interrupts.

ystros: (I'm getting the sense that the ThronesDB FAQ may just be wrong here, since it definitely fits the definition of a Delayed Effect, which should happen at step 5 of the triggered condition resolution, just before Reactions but after Interrupts)

DukeTax: Yeah, it's definitely wrong, so 'afterChallenge:interrup' is right?

ystros:
Just afterChallenge. afterChallenge:interrupt would fire before interrupts
are prompted. Plain afterChallenge will fire after interrupts have resolved
but before forced reactions. But again, that should have been what it was
already doing before your change.

DukeTax:
But the way it was implemented, untilEndOfChallenge lasts until after keywords have been resolved.

PR #1077, opened 2017-06-22
Trigger reaction events for Annals of Castle Black.

### LINKED ISSUE #1048: Annals of Castle Black

ystros:
Previously, Annals of Castle Black would not allow event interrupts /
reactions to trigger from discard. This is because event listeners were
only ever bound if the card was in the expected location - in the case
of event cards, in hand.

Now, interrupts / reactions listen to events in all open information
locations plus while in hand. Since the `location` property restricts
whether the ability actually fires, abilities should not activate while
in inappropriate locations even if it listens to events.

Fixes #1048.

### DIFF
--- a/server/game/triggeredability.js
+++ b/server/game/triggeredability.js
@@ -109,7 +109,11 @@ class TriggeredAbility extends BaseAbility {
     }
 
     isEventListeningLocation(location) {
-        return this.location === location;
+        // Reactions / interrupts need to listen for events in all open
+        // information locations plus while in hand. The location property of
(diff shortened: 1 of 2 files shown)


ystros:
This PR scares me and I feel like it should have broken something, but no tests broke and I couldn't trigger any funkiness through manual testing either. We'll see? 🙃  🙃  🙃

PR #1078, opened 2017-06-22
implement City Watch

zacchiro:


zacchiro:
This is the straightforward implémentation of City Watch, based on what I've seen in Northern Refugee, but **it doesn't work yet**. Am I using "ambush" wrong in this context, @ystros ?

PR #1079, opened 2017-06-22
implement Grizzled Miner

zacchiro:


server/game/cards/characters/07/grizzledminer.js
```diff
@@ -0,0 +1,26 @@
...
+            effect: ability.effects.reduceSelfCost('marshal', this.getDiscount())
+        });
+    }
+
+    getDiscount() {
+        let cards = this.controller.filterCardsInPlay(
+            card => card.isFaction('thenightswatch')
+                && (card.getType() === 'attachment' || card.getType() === 'location'));
+
+        return Math.min(cards.length, this.getCost() - 1);
```
ystros: Much as I'd love it to be this simple, the minimum of 1 applies to the entire card, not just the discount. So, for example, if I used 2 Kingsroads to reduce the cost of Grizzled Miner by a total of 6, I would still have to pay 1 gold.

http://thronesdb.com/card/07006

Thus, this needs to be a new effect that sets a minimum cost on a card and check it during cost reduction calculations in `Player.getReducedCost`. Should be fairly straightforward, actually - the return value `Math.max` check should just use the minimum instead of just 0.

server/game/cards/characters/07/grizzledminer.js
```diff
@@ -0,0 +1,26 @@
+const DrawCard = require('../../../drawcard.js');
+
+class GrizzledMiner extends DrawCard {
+
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            location: 'any',
+            targetType: 'player',
+            targetController: 'current',
+            effect: ability.effects.reduceSelfCost('marshal', this.getDiscount())
```
ystros: Should be `() => this.getDiscount()` for dynamic costs. This will set the cost at the time of creating the card (which would be 0 if it weren't crashing)

server/game/cards/characters/07/grizzledminer.js
zacchiro: OK, this should now (at last!) be done in the last update I've just pushed to this PR.

PR #1080, opened 2017-06-23
don't trigger Sannis (TIMC) if there are no valid targets

zacchiro:
also, when actually triggering him, do not allow to select standing characters

### DIFF
--- a/server/game/cards/characters/04/stannisbaratheon.js
+++ b/server/game/cards/characters/04/stannisbaratheon.js
@@ -14,20 +14,26 @@ class StannisBaratheon extends DrawCard {
 
         this.reaction({
             when: {
-                onDominanceDetermined: (event, winner) => this.controller === winner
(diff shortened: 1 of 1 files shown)


zacchiro:
Actually, i might need a rule check on this. I was thinking that Stannis cannot pick standing characters, as the game state won't change, and hence they are invalid targets. But in fact I just realized that they might kneel *after* they have been chosen, e.g., Tickler. So there is a game state change, which will be remembered until the next standing phase. If this is the case, feel free to reject this PR.

DukeTax:
> I just realized that they might kneel after they have been chosen, e.g., Tickler. So there is a game state change, which will be remembered until the next standing phase

Yeah, this should be correct. Basically the application of the effect itself is a change in game state, whether it matters in the end or not.

ystros:
Yup, what @DukeTax said.

PR #1081, opened 2017-06-23
Update plot resets

DukeTax:
* Skips the prompt for Marched to the Wall, Wildfire Assault and Political Disaster if there's nothing to select.
* Fixes Political Disaster not displaying the FP's choices before prompting the second player to choose.
* Improves messaging. Most notably, while the word 'save' informally covers such effects, it also has a specific meaning in the game making it unsuitable here.

### DIFF
--- a/server/game/cards/plots/02/politicaldisaster.js
+++ b/server/game/cards/plots/02/politicaldisaster.js
@@ -14,6 +14,11 @@ class PoliticalDisaster extends PlotCard {
     }
 
     onSelect(player, cards) {
+        if(_.isEmpty(cards)) {
+            this.game.addMessage('{0} does not choose any locations for {1}', player, this);
+        } else {
+            this.game.addMessage('{0} chooses {1} for {2}', player, cards, this);
+        }
         this.selections.push({ player: player, cards: cards });
         this.proceedToNextStep();
         return true;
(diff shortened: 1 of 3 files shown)


server/game/cards/plots/02/politicaldisaster.js
```diff
@@ -26,17 +31,17 @@ class PoliticalDisaster extends PlotCard {
...
+            let player = selection.player;
+            let toDiscard = _.difference(player.filterCardsInPlay(card => card.getType() === 'location'), selection.cards);
 
             _.each(toDiscard, card => {
                 player.discardCard(card);
             });
 
             if(_.isEmpty(toDiscard)) {
-                this.game.addMessage('{0} does not discard any locations with {1}', player, this);
+                this.game.addMessage('{0} does not have any locations discarded for {1}', player, this);
```
ystros: @cryogen may have opinions about the use of 'killed' and 'discarded' in this PR vs the previous wording. Otherwise looks good to me.

DukeTax: Sure. Personally I'd like to move away from the "[user] uses [card] to do [things]" template for effects where the [user] did not have any choice in the matter.

cryogen:
I see where you're coming from but I'm gonna leave this one until I can think about how to better word these kinds of things

server/game/cards/plots/02/politicaldisaster.js
zacchiro: In the same vein of these changes, updating Confiscation so that it doesn't prompt if there are no eligible targets would be Good™ too.
(Or, actually, shouldn't we just change the general card selector so that it doesn't prompt if there are no eligible targets for cardCondition ?)

DukeTax: Confiscation already has the feature by virtue of it using the target API (which can't be used on these cards because of its alternating nature).

Second point sounds.... good. :)

PR #1087, opened 2017-06-23
Implement Saving The Kingdom

DukeTax:


server/game/cards/events/06/savingthekingdom.js
```diff
@@ -0,0 +1,40 @@
...
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Make character unable to stand',
+            phase: 'challenge',
+            target: {
+                activePromptTitle: 'Select a character',
+                cardCondition: card => card.location === 'play area' && card.getType() === 'character' && card.getStrength() <= 3
+            },
+            handler: context => {
+                this.untilEndOfRound(ability => ({
```
ystros: The card text specifies that it kneels the card as well. So you'll need to kneel the card, plus add `gameAction: 'kneel'` to the target properties.

DukeTax: About the gameAction: that would prevent the whole effect right? While (simularly to Waking the Dragon), if card has multiple effects, you try to resolve as many of them so the stand prevention should still happen...

ystros: Oh, interesting. In that case, yes, you were correct to leave off the `gameAction`, as that will make you unable to target a character that couldn't be kneeled.

PR #1088, opened 2017-06-23
Implement Tywin's Stratagem

DukeTax:


server/game/cards/events/06/tywinsstratagem.js
```diff
@@ -0,0 +1,55 @@
...
+            cost: ability.costs.payGold(1),
+            handler: () => {
+                this.game.addMessage('{0} pays 1 gold to move {1} back to their hand', this.controller, this);
+                this.controller.moveCard(this, 'hand');
+            }
+        });
+    }
+
+    cardCondition(card) {
+        return card.location === 'play area' && card.getType() === 'character' && card.getStrength() <= 2;
```
ystros: Should be checking printed cost is 2 or lower, not strength.

PR #1093, opened 2017-06-24
Implement Beric Dondarrion

DukeTax:


client/About.jsx
```diff
@@ -49,7 +49,7 @@ class About extends React.Component {
                     <li>/power x - Sets the power of a card to x</li>
                     <li>/strength x - Sets the strength of a card to x</li>
                     <li>/take-icon x - Take an x icon from a character; where 'x' is as per '/give-icon'</li>
-                    <li>/token t x - Set the token count of a card of type 't' to 'x'. Currently used token types are: 'betrayal', 'gold', 'poison', 'vengeance'.</li>
+                    <li>/token t x - Set the token count of a card of type 't' to 'x'. Currently used token types are: 'betrayal', 'gold', 'poison', 'vengeance', 'ear', 'venom', 'kiss', 'valarmorghulis' and 'stand'.</li>
```
ystros: Good catch but could we keep them in alphabetical order?

server/game/cards/characters/06/bericdondarrion.js
```diff
@@ -0,0 +1,36 @@
+const DrawCard = require('../../../drawcard.js');
+
+class BericDondarrion extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: this,
+            effect: ability.effects.dynamicStrength(() => this.tokens['kiss'])
+        });
+
+        this.forcedReaction({
```
ystros: Could you add a TODO comment here saying that we need a way for it not to be cancellable by things like Treachery? It'll be easy to forget otherwise. Or, alternatively, implement an ability flag that will prevent it from being cancellable.

PR #1094, opened 2017-06-24
Implement The Brotherhood Without Banners

DukeTax:


ystros:
This looks fine for the backend implementation, we should also update deck validation to implement the 'cannot include loyal characters' restriction. Deck validations are done in `/client/deck-validator.js`.

(it's kind of a mess overall but should be easy to add a new rule. Check the Fealty restrictions around line 158, BwB's should be similar, though keep in mind plot cards also need to be check for loyalty)

DukeTax:
It only disallows loyal characters though. :)

ystros:
Whoops, so it does. 😃

PR #1095, opened 2017-06-24
Implement Unbridled Generosity

DukeTax:


server/game/cards/events/06/unbridledgenerosity.js
```diff
@@ -0,0 +1,35 @@
+const _ = require('underscore');
+
+const DrawCard = require('../../../drawcard.js');
+
+class UnbridledGenerosity extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Put gold on cards',
+            handler: () => {
+                this.game.promptForSelect(this.controller, {
```
ystros: Should use target API instead (the card text does say 'choose').

PR #1099, opened 2017-06-26
Implement Podrick Payne & Rakharo

DukeTax:
Similarly to Stark Harrenhal, I took the liberty to have Podrick trigger solely on characters you control.

### DIFF
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -210,6 +210,19 @@ const Costs = {
         };
     },
     /**
+     * Cost that will put into play the card that initiated the ability.
+     */
+    putSelfIntoPlay: function() {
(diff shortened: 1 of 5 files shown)


server/game/costs.js
```diff
@@ -210,6 +210,19 @@ const Costs = {
         };
     },
     /**
+     * Cost that will put into play the card that initiated the ability.
+     */
+    putSelfIntoPlay: function() {
+        return {
+            canPay: function() {
+                return true;
```
ystros: You should check `canPutIntoPlay` to handle cases where the character may be in the dead pile / you've take control of a copy of the same card from an opponent / etc.

PR #1103, opened 2017-06-29
Implement Oldtown Informer

DukeTax:


server/game/cards/events/06/oldtowninformer.js
```diff
@@ -0,0 +1,38 @@
+const DrawCard = require('../../../drawcard.js');
+
+class OldtownInformer extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardEntersPlay: () => this.game.currentPhase = 'challenge' && this.tokens['gold'] >= 1
```
ystros: Currently this will set the current phase to 'challenge' every single time a card enters play. 😬 

`=== 'challenge'` instead.

DukeTax: Oh wow, that's bad. :P

PR #1105, opened 2017-06-30
Properly remove / apply effects during reattach.

### LINKED ISSUE #1097: Arya's Gift
Hi~ When i shift Milk of the Poppy from 1 character to another using Arya's Gift. The milk effect will still stay on the first character.

ystros:
Previously, when an attachment was re-attached from one card to another,
it would leave its effect on the original parent in addition to applying
it to the new parent. Now, effects are removed immediately when the
parent changes, and effects are applied to the new parent asynchronously
after the move has completed.

Fixes #1097

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -316,10 +316,12 @@ class BaseCard {
         this.tokens = {};
     }
 
-    moveTo(targetLocation) {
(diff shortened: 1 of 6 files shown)


cryogen:
How the hell did those debug messages get in? haha

PR #1110, opened 2017-07-04
Implement promptForIcon gamestep

DukeTax:


server/game/cards/characters/02/nymeriasand.js
```diff
@@ -1,69 +1,35 @@
...
+            limit: ability.limit.perPhase(1),
+            phase: 'challenge',
+            target: {
+                activePromptTitle: 'Select a character',
+                cardCondition: card => card.location === 'play area' && card.controller !== this.controller &&
+                                       card.getType() === 'character'
             },
-            source: this
+            handler: context => {
+                this.game.promptForIcon(this.controller, context);
```
ystros: I'm not sure I like how this is used. While putting it onto the context object isn't a bad idea, the fact that it's async but doesn't appear async is a problem. It would be very easy to forget to do the `queueSimpleStep`s you've added.

Couple options: we could make this a flag on abilities, so that you're not calling it directly. By the time it reaches the handler, it will have the icon on the context object like you've done.
```javascript
this.action({
    // ...
    chooseIcon: true,
    handler: context => {
        // do something with context.icon
    }
});
```

Or, to make the async nature more clear, the prompt should take a callback instead of the context object.
```javascript
this.game.promptForIcon(this.controller, icon => {
    // do something with icon
});
```

DukeTax: The ability engine is still a bit of a maze to me, so went with the callback, though I might revisit this in the future.

PR #1111, opened 2017-07-05
Convert 'cannot stand during standing' to true cannot effect

DukeTax:
Previously, effects that made cards unable to stand during the standing phase were implemented by setting and checking a flag during the mandatory standing in the standing phase. This worked fine for the majority of use cases, but it still wrongly allowed cards to be stood during the standing phase by for example a card like Magister Illyrio. This PR fixes that by porting these implementations to true cannot effects.

Needs #1109


ystros:
Looks good! As you mention, this should be merged once #1109 is merged.

PR #1114, opened 2017-07-05
Implement remaining cards from The Red Wedding

DukeTax:
- Dacey Mormont ~~won't work currently as a result of strength X cards now being defined as 'null' instead of '0' in the JSON data (Beric Dondarrion has the same issue)~~ needs #1115.

- Improved Fortifications needs #1108 to function.

- Let me know if you want me to convert the trigger of Breaker of Chains to the upcoming onChallengeInitiated event.

### DIFF
--- a/server/game/cards/events/06/plunder.js
+++ b/server/game/cards/events/06/plunder.js
@@ -1,0 +1,36 @@
+const DrawCard = require('../../../drawcard.js');
+
+class Plunder extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Gain gold',
+            condition: () => this.getGold() >= 1,
+            cost: ability.costs.kneelFactionCard(),
+            handler: () => {
+                let gold = this.getGold();
+                this.game.addGold(this.controller, gold);
+                this.game.addMessage('{0} uses {1} and kneels their faction card to gain {2} gold', this.controller, this, gold);
+            }
+        });
+    }
+
+    getGold() {
+        let opponent = this.game.getOtherPlayer(this.controller);
+
+        if(!opponent) {
+            return;
+        }
+
+        return opponent.allCards.reduce((num, card) => {
+            if(card.location === 'discard pile' && (card.getType() === 'location' || card.getType() === 'attachment')) {
+                return num + 1;
+            }
+
+            return num;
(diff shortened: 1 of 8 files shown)


server/game/cards/events/06/plunder.js
```diff
@@ -0,0 +1,36 @@
...
+                this.game.addMessage('{0} uses {1} and kneels their faction card to gain {2} gold', this.controller, this, gold);
+            }
+        });
+    }
+
+    getGold() {
+        let opponent = this.game.getOtherPlayer(this.controller);
+
+        if(!opponent) {
+            return;
```
ystros: This should return 0 for single player games. Otherwise, you'll end up adding `undefined` gold.

server/game/cards/locations/06/arborvineyard.js
```diff
@@ -0,0 +1,29 @@
...
+                let gold = this.moreSummerThanWinterPlotsRevealed() ? 2 : 1;
+                this.game.addGold(context.player, gold);
+                this.game.addMessage('{0} kneels {1} to gain {2} gold', context.player, this, gold);
+            }
+        });
+    }
+
+    moreSummerThanWinterPlotsRevealed() {
+        let summerPlots = _.filter(this.game.getPlayers(), player => player.activePlot && player.activePlot.hasTrait('Summer'));
+        let winterPlots = _.filter(this.game.getPlayers(), player => player.activePlot && player.activePlot.hasTrait('Winter'));
```
ystros: Use `this.game.getNumberOfPlotsWithTrait` instead.

server/game/cards/events/06/plunder.js
DukeTax: Doesn't matter here with the condition check right?

ystros: You're correct, I missed the condition. Still glad you changed it though, it was kind of odd regardless.

ystros:
I'm going to go ahead and merge despite needing #1108 and #1115. This PR doesn't introduce any new bugs, just reveals existing ones fixed by those PRs. Contrast with the PR changing cannot-stand-during-stand-phase which would break those effects entirely if merged without the corresponding fix to effect recalculation.

PR #1116, opened 2017-07-06
Handle layout for additional piles client side.

ystros:
Removes the presentational concerns of 'area' and 'title' for Scheme and
Out of Game piles from the backend and moves them to the client side.


cryogen:
oh good, this bothered me when it first went in :) 👍

ystros:
Yeah, it's been on my todo list since then. 😰

PR #1125, opened 2017-07-07
Implement Craster's resurrect ability.

ystros:
* Keeps track of the current round of the game.
* Keeps track of when characters are killed in a card event log (can be extended in the future should we need to track discards, return to hand, etc for effects similar to Craster).
* Implements Craster's ability.

I would have preferred to track character's killed directly on Craster, but he can't listen to events when he's not in play. Thus, if characters are killed during the same phase prior to him entering play (see test w/ Valar vs Called Into Service), it wouldn't register those characters as having been killed when tracking on the card directly.

### DIFF
--- a/server/game/cardeventlog.js
+++ b/server/game/cardeventlog.js
@@ -1,0 +1,32 @@
+const _ = require('underscore');
+
+const EventRegistrar = require('./eventregistrar.js');
+
+class CardEventLog {
+    constructor(game) {
+        this.game = game;
+        this.events = new EventRegistrar(game, this);
+
+        this.log = [];
+
+        this.events.register(['onCharacterKilled']);
+    }
+
+    any(predicate) {
+        return _.any(this.log, predicate);
+    }
+
+    filterCards(predicate) {
+        return _.pluck(_.filter(this.log, predicate), 'card');
+    }
+
+    register(type, card) {
+        this.log.push({ type: type, card: card, round: this.game.round, phase: this.game.currentPhase });
+    }
+
+    onCharacterKilled(event) {
+        this.register('kill', event.card);
+    }
+}
+
+module.exports = CardEventLog;
--- a/server/game/cards/characters/04/craster.js
+++ b/server/game/cards/characters/04/craster.js
@@ -1,3 +1,5 @@
+const _ = require('underscore');
+
 const DrawCard = require('../../../drawcard.js');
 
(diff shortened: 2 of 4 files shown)


ystros:
@cryogen - Actually, it may be possible to create something like the card log that's specific to Craster and that lives on his card object instead of game (won't be restricted by the normal event listening locations if it's a separate object). Let me know if you prefer that. It'd be less flexible for future cards and more specific to Craster, but it wouldn't track stuff that doesn't need to be tracked if Craster isn't in the deck.

zacchiro:
> Keeps track of the current round of the game.
> Keeps track of when characters are killed in a card event log (can be extended in the future should we need to track discards, return to hand, etc for effects similar to Craster).

As you know, big +1 on my part on the event log idea, it will come in handy for a lot more in the future I think.

But then, why make it opt-in, with the need to individually register the events we want to keep track of? 

Just go ahead and make it register *any* event whatsoever. If we leave it opt-in it will *never* be complete, and its usefulness when complete is much higher (e.g., for debugging purposes, as it will be very easy to make it persistent, allowing post-mortem inspection of it).

ystros:
Auto registering each event would be more useful long term but it would also balloon the memory footprint for each game significantly for not a lot of use. That's why I may it specifically whitelisted events. It would also change the lookup from this:

```javascript
let characters = this.game.filterCardsInLog(log => (
    log.type === 'kill' &&
    log.round === this.game.round &&
    log.phase === this.game.currentPhase &&
    log.card.location === 'dead pile'
 ));
```

to this:
```javascript
// Need to look up by event because not every event has a card associated with it
// and those that do it isn't always called 'card' (e.g. onBypassedByStealth has
// two different cards associated with it - source and target)
let events = this.game.filterEventsInLog(log => ( 
    log.event.name === 'onCharacterKilled' &&
    log.abilityType === 'reaction' && // needed because interrupts would be recorded, but could be cancelled.
    log.round === this.game.round &&
    log.phase === this.game.currentPhase &&
    log.event.card.location === 'dead pile'
 ));
 let characters = _.pluck(events, 'card');
```

I got the distinct impression that @cryogen didn't like this approach overall from Discord, though, so this discussion may be moot.

cryogen:
It definitely set off my "wtf" sensors, but I've not had chance to properly think it over yet :)

PR #1126, opened 2017-07-07
Implement Pay The Iron Price

DukeTax:


server/game/gamesteps/attachmentprompt.js
```diff
@@ -11,7 +11,7 @@ class AttachmentPrompt extends UiPrompt {
     continue() {
         this.game.promptForSelect(this.player, {
             activePromptTitle: 'Select target for attachment',
-            cardCondition: card => this.player.canAttach(this.attachmentCard.uuid, card),
+            cardCondition: card => this.attachmentCard.owner.canAttach(this.attachmentCard.uuid, card),
```
ystros: ... I think I see why this was necessary but I don't really like it. But I guess we can deal with it separately, as the whole attachment system is still a mess.

PR #1127, opened 2017-07-07
Saving the Kingdom: mention kneel in card title and log message

zacchiro:


DukeTax:
Aah. I had fixed the message #1111 but not the label.  (y)

PR #1136, opened 2017-07-08
Implement first spoiled batch from House of Thorns

DukeTax:


test/server/card/drawcard.modifyPower.spec.js
```diff
@@ -6,7 +6,7 @@ const DrawCard = require('../../../server/game/drawcard.js');
 describe('DrawCard', function () {
     beforeEach(function () {
         this.testCard = { code: '111', label: 'test 1(some pack)', name: 'test 1' };
-        this.gameSpy = jasmine.createSpyObj('game', ['raiseEvent', 'checkWinCondition']);
+        this.gameSpy = jasmine.createSpyObj('game', ['raiseEvent', 'checkWinCondition', 'applyGameAction']);
```
ystros: Adding the following should get tests passing again:
```javascript
this.gameSpy.applyGameAction.and.callFake((action, card, callback) => callback(card));
```

server/game/cards/characters/09/macetyrell.js
```diff
@@ -0,0 +1,43 @@
...
+                this.game.addMessage('{0} pays 1 gold to gain 1 power on {1}', this.controller, this);
+            }
+        });
+
+        this.action({
+            title: 'Remove character from game',
+            cost: ability.costs.kneelFactionCard(),
+            target: {
+                activePromptTitle: 'Select a character',
+                cardCondition: card => card.location === 'play area' && card.controller === this.controller &&
```
ystros: Technically, the card doesn't say that it has to be a Tyrell character you control, only that it's another Tyrell character. Unless I'm misreading or missing a rule, it seems like you could use the ability offensively by removing your opponent's heavy hitting Tyrell character from the game at the start of the challenges phase.

(if that's correct, it means we need a saveable version of "remove from game" in case such a card has a duplicate on it)

DukeTax: He says "you control" on the card, just the data entry is incomplete.

ystros: I swear I didn't see the 'you control' when looking at the card image, but you're right.

PR #1137, opened 2017-07-09
Refactor the deck builder to make better use of redux, and to only send

cryogen:
necessary data to and from the server.  Also add notification of invalid and
not released decks to opponents in the game creation stage


server/repositories/cardRepository.js
```diff
@@ -0,0 +1,43 @@
+const _ = require('underscore');
+
+const logger = require('../log.js');
+const BaseRepository = require('./baseRepository.js');
+
+class DeckRepository extends BaseRepository {
+    getCards(brief, callback) {
```
ystros: This is probably very nitpicky, but I feel like this would be better served as two separate methods rather than passing a boolean flag whose meaning is unclear unless you look at the method itself.

(spoilers, I'm probably going to separate `getStrength(bool)` into `getPrintedStrength()` and `getStrength()` in the near future)

server/lobby.js
```diff
@@ -446,9 +449,36 @@ class Lobby {
...
+                    _.each(deck.drawCards, draw => {
+                        draw.card = cards[draw.card.code];
+                    });
+
+                    if(deck.agenda) {
+                        deck.agenda = cards[deck.agenda.code];
+                    }
+
+                    let validation = validateDeck(deck, packs);
+                    deck.status = validation.status;
```
ystros: Again on the nitpicky side, but it would be nice if this method read more like:
```javascript
this.loadDeck(deckId, deck => { // or some method on deckRepository
    game.selectDeck(socket.user.username, deck);
    this.sendGameState(game);
});
```

Instead of the 3 nested callbacks it is currently:
```javascript
this.deckRepository.getById(deckId, deck => {
    this.cardRepo.getCards(cards => {
        this.cardRepo.getPacks(packs => {
            // do a bunch of stuff preparing the deck
            // FINALLY actually select the deck and send data
        });
    });
});

server/repositories/cardRepository.js
cryogen: Since i've already forgotten what that flag is for, you might have a point :)

cryogen: I might prefer named options, rather than changing the method name, though, mainly because they do similar things and it's nice to be able to default behaviour without having yet another method they both call

server/lobby.js
cryogen: As discussed I'm gonna leave this be until I do the promises stuff

PR #1142, opened 2017-07-11
Fix adding banner cards for Alliance in builder.

ystros:
Fixes THRONETEKI-Y2
Fixes THRONETEKI-YE
Fixes THRONETEKI-YM
Fixes THRONETEKI-YN


ystros:
Eh, hold on, this fixes the error but leaves other major problems.

PR #1144, opened 2017-07-11
Refactor deck validation

ystros:


client/deck-validator.js
```diff
@@ -67,174 +36,208 @@ function isCardInReleasedPack(packs, card) {
...
+            }
+        ]
+    },
+    // Kings of Summer
+    '04037': {
+        cannotInclude: card => card.type_code === 'plot' && hasTrait(card, 'Winter'),
+        rules: [
+            {
+                message: 'Kings of Summer cannot include Winter plot cards',
+                condition: deck => !_.any(deck.plotCards, cardQuantity => hasTrait(cardQuantity.card, 'Winter'))
```
ystros: Rules like these weren't strictly necessary, as the deck would be ruled invalid from the `cannotInclude` function. However, it's nice to have the reason the deck is invalid highly visible. Either way, the specific plot will be labelled as not allowed.

PR #1152, opened 2017-07-11
Use printed cost for Hand's Judgment to cancel.

### LINKED ISSUE #1076: canceling a 1st Hand's Judgment with a 2nd one should always cost 0...
... rather than costing the price of the 1st HJ.

ystros:
Fixes #1076.

### DIFF
--- a/server/game/cards/events/01/thehandsjudgment.js
+++ b/server/game/cards/events/01/thehandsjudgment.js
@@ -28,7 +28,7 @@ class TheHandsJudgment extends DrawCard {
             return super.getCost();
         }
 
-        return this.eventToInterrupt.cardData.cost || 0;
+        return this.eventToInterrupt.getPrintedCost();
     }
 }
 
--- a/server/game/cards/events/01/thehandsjudgment.js
+++ b/server/game/cards/events/01/thehandsjudgment.js
@@ -28,7 +28,7 @@ class TheHandsJudgment extends DrawCard {
             return super.getCost();
         }
 
-        return this.eventToInterrupt.getCost();
+        return this.eventToInterrupt.cardData.cost || 0;
     }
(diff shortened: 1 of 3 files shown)


server/game/cards/events/01/thehandsjudgment.js
```diff
@@ -28,7 +28,7 @@ class TheHandsJudgment extends DrawCard {
             return super.getCost();
         }
 
-        return this.eventToInterrupt.getCost();
+        return this.eventToInterrupt.cardData.cost || 0;
```
cryogen: This is literally what getCost() does, except the || 0 bit, so do we not need to fix it in DrawCard instead of here?

ystros: The problem is Hand's Judgement overrides `getCost`. So:
* I play Hear Me Roar. HearMeRoar.getCost() === DrawCard.getCost() === cardData.cost === 1.
* Opponent uses Hand's Judgment to cancel. HandsJudgement.getCost() === the played event's cost === 1.
* I use Hand's Judgment to cancel my opponent's Hand's Judgment. HandsJudgment.getCost() === the result of my opponent's HandsJudgment.getCost() === the originally played event's cost === 1.

Which is incorrect because cancelling Judgment with another Judgment should cost 0 since it's the printed cost of the event. Hence, the major fix here isn't the || 0 but explicitly grabbing the printed cost, since getCost can be overridden (and has been for Judgment).

The || 0 was a superfluous safety measure, but probably should exist in DrawCard as well.

cryogen: Oh. Ew. Which idiot did that 🙃

In that case we probably need a getPrintedCost as for strength

PR #1168, opened 2017-07-14
Implement The Iron Bank Will Have Its Due

DukeTax:


server/game/cards/events/06/theironbankwillhaveitsdue.js
```diff
@@ -0,0 +1,33 @@
+const DrawCard = require('../../../drawcard.js');
+
+class TheIronBankWillHaveItsDue extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Return character to gain gold',
+            condition: () => !this.game.currentPhase === 'taxation',
```
ystros: Should be !==.

`!this.game.currentPhase === 'taxation'` is `(!this.game.currentPhase) === 'taxation'`, which boils down to `false === 'taxation'`.

PR #1171, opened 2017-07-15
Implement Devout Freerider

DukeTax:
Not entirely sure if this is the desired way, but..... it works I guess. I don't really like the additional checks on other cards to prevent trigger prompts, but it's only a handful (and even then it's 95% Tyrion Lannister (Core)).

Other slight issue: there are a bunch of cards that transfer gold implemented by invoking game.addGold() twice, instead of just game.transferGold(), which could be prevented from resolving correctly by Devout Freerider. It's on the list for when I go through the entire cardpool.

### DIFF
--- a/server/game/cards/characters/01/oldforesthunter.js
+++ b/server/game/cards/characters/01/oldforesthunter.js
@@ -4,29 +4,15 @@ class OldForestHunter extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
             title: 'Discard a card to gain 1 gold',
-            method: 'discard',
-            limit: ability.limit.perPhase(1)
+            condition: () => !this.controller.cannotGainGold,
+            cost: ability.costs.discardFromHand(),
+            limit: ability.limit.perPhase(1),
+            handler: context => {
+                this.game.addGold(this.controller, 1);
+                this.game.addMessage('{0} uses {1} and discards {2} from their hand to gain 1 gold', 
+                                      this.controller, this, context.discardCostCard);
+            }
         });
-    }
-
-    discard(player) {
-        this.game.promptForSelect(player, {
-            activePromptTitle: 'Select a card to discard',
-            source: this,
-            cardCondition: card => card.location === 'hand',
-            onSelect: (p, card) => this.onCardSelected(p, card)
-        });
-
-        return true;
-    }
-
-    onCardSelected(player, card) {
-        this.controller.discardCard(card);
-        this.game.addGold(this.controller, 1);
-
-        this.game.addMessage('{0} uses {1} to discard {2} and gain 1 gold', player, this, card);
-
-        return true;
     }
 }
 
--- a/server/game/cards/characters/01/tyrionlannister.js
+++ b/server/game/cards/characters/01/tyrionlannister.js
@@ -4,12 +4,12 @@ class TyrionLannister extends DrawCard {
     setupCardAbilities(ability) {
         this.reaction({
             when: {
-                onChallengeInitiated: event => event.challenge.challengeType === 'intrigue'
+                onChallengeInitiated: event => event.challenge.challengeType === 'intrigue' && !this.controller.cannotGainGold
             },
             limit: ability.limit.perRound(2),
             handler: () => {
                 this.game.addGold(this.controller, 2);
-                this.game.addMessage('{0} uses {1} to gain 2 gold as an intrigue challenge has been declared', this.controller, this);
+                this.game.addMessage('{0} uses {1} to gain 2 gold', this.controller, this);
             }
         });
     }
--- a/server/game/cards/events/02/indoransname.js
+++ b/server/game/cards/events/02/indoransname.js
@@ -4,12 +4,12 @@ class InDoransName extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
             title: 'Kneel faction card to gain gold',
-            condition: () => this.controller.getNumberOfUsedPlots() >= 1,
+            condition: () => this.controller.getNumberOfUsedPlots() >= 1 && !this.controller.cannotGainGold,
             cost: ability.costs.kneelFactionCard(),
             handler: context => {
                 let gold = this.controller.getNumberOfUsedPlots();
                 this.game.addGold(this.controller, gold);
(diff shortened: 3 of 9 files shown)


server/game/cards/characters/01/tyrionlannister.js
```diff
@@ -4,12 +4,12 @@ class TyrionLannister extends DrawCard {
     setupCardAbilities(ability) {
         this.reaction({
             when: {
-                onChallengeInitiated: event => event.challenge.challengeType === 'intrigue'
+                onChallengeInitiated: event => event.challenge.challengeType === 'intrigue' && !this.controller.cannotGainGold
             },
             limit: ability.limit.perRound(2),
             handler: () => {
                 this.game.addGold(this.controller, 2);
```
zacchiro: same issue here

server/game/cards/characters/01/oldforesthunter.js
```diff
@@ -4,30 +4,16 @@ class OldForestHunter extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
             title: 'Discard a card to gain 1 gold',
-            method: 'discard',
-            limit: ability.limit.perPhase(1)
+            condition: () => !this.controller.cannotGainGold,
+            cost: ability.costs.discardFromHand(),
+            limit: ability.limit.perPhase(1),
+            handler: context => {
+                this.game.addGold(this.controller, 1);
```
zacchiro: this should probably be moved after the messagae below, as otherwise the log will see first the message about "cannot gain gold", and later then one about gaining gold, which would be confusing, I guess.

Off the top of my head, logging first and then doing the action is also the most common message logging patterns in card implementations (at least in recent card implementations, that is).

server/game/cards/events/02/indoransname.js
```diff
@@ -4,12 +4,12 @@ class InDoransName extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
             title: 'Kneel faction card to gain gold',
-            condition: () => this.controller.getNumberOfUsedPlots() >= 1,
+            condition: () => this.controller.getNumberOfUsedPlots() >= 1 && !this.controller.cannotGainGold,
             cost: ability.costs.kneelFactionCard(),
             handler: context => {
                 let gold = this.controller.getNumberOfUsedPlots();
                 this.game.addGold(this.controller, gold);
```
zacchiro: ditto

server/game/cards/characters/01/oldforesthunter.js
DukeTax: Agreed, it is probably preferable to see for example 'DukeTax uses Tyrion Lannister to gain 2 gold' followed by 'DukeTax cannot gain gold', instead of the other way around. 

In these three instances it doesn't matter though since these cannot prompt anyway if you cannot gain gold (and they make up probably >95% of uses cases for gaining gold during the challenge phase).

ystros:
Hmm. I hate having to add these checks as well. You've hit some of the major cases, but here are more. I don't know how I feel about adding the check to all of these though. Some of these are more useful for gaining gold for ambushes than others, which I feel is the biggest intention of Devout Freerider.

* Lions of the Rock
* Highgarden Minstrel
* Tithe
* Plunder
* Tower of the Sun
* Golden Tooth
* The Hightower
* To The Rose Banner (has added complication in that if you gained no gold, the 'Then' clause shouldn't trigger
* Crossroads Sellsword (doesn't truly matter since he only triggers when he's killed, just wouldn't prompt you at all if you cannot gain gold)
* GoH Ser Davos (though maybe skip this one? Since you have a choice to either gain gold or draw and we don't have a mechanism to prevent a specific choice)
* SeasonedWoodsman (similar problem as w/ Ser Davos)
* The Dornishman's Wife (can't trigger if you only meet the 'gain gold' condition)
* Many cards that gain gold during non-challenge phases (particularly marshalling) that would need to be changed if another card with the effect were released for other phases.

server/game/cards/characters/01/oldforesthunter.js
ystros: Re: conflicting messages, I've thought about converting some methods to promises or promise-like syntax so that in cases where an action doesn't succeed (card didn't get put into play as expected, gold wasn't gained in this case), the message doesn't go out. Something like:

```javascript
this.game.addGold(this.controller, 1).then(gold => {
   this.game.addMessage('{0} uses {1} to gain {2} gold', this.controller, this, gold);
})
```

I haven't done it because 1. getting promises to work properly in this context will require rewriting most of the pipeline code, 2. it's kinda ugly (though can perhaps make it prettier), and 3. I'd want to do it in a way that supports 'Then' abilities, which also require the pre-then to have succeeded before proceeding.

cryogen: Am I missing something here or is this not just as simple as having addGold return false and checking that?

ystros: Would work in this case since things are synchronous (at least for now). Doesn't work in the general case (async things like discarding cards, killing characters, leaving play, etc).

In any case, adding the cannot-gain-gold check to the triggering condition for appropriate cards sidesteps this problem, acting as an upfront if for the ability.

DukeTax:
Cards like The Hightower and The Iron Will Have Its Due are even more problematic, since gaining gold isn't their only effect, and they would need to be able to trigger even if you can't gain gold. They would ideally then have their messages broken up and checked whether certain effects resolved before displaying those messages.

So yeah, this is quite a list, though a lot of these cards are even less played than the Devout Freerider itself...

ystros:
I would say let's explicitly check for Lions of the Rock, since it's *explicitly* about gaining gold during the challenges phase and only gaining gold, then say "good enough." Maybe Plunder as well since you did the somewhat similar In Doran's Name? I'm not sure either of these are really played, the rest can have funky messages and we'll fix them if people complain / are confused.

server/game/cards/characters/01/oldforesthunter.js
cryogen: I was more thinking for this specific case

cryogen:
Long term I see this working more elegantly by having cards apply actions, much like costs and targets, something like action: actions.gainGold(2), it would be much easier to apply cannots like this with that.  I'm sure I wont end up doing the work and @ystros has better ideas though :)

PR #1173, opened 2017-07-16
News admin

cryogen:


client/ReduxActions/news.js
```diff
@@ -23,3 +24,44 @@ export function receiveNews(news) {
...
+
+            if(options && options.limit) {
+                optionsStr += 'limit=' + options.limit;
+            }
+
+            if(optionsStr !== '') {
+                optionsStr = '?' + optionsStr;
+            }
+
+            return $.ajax('/api/news/' + optionsStr, { cache: false });
```
ystros: Assuming this is actually jQuery and not a lookalike interface that may not support everything, you should be able to specify the query params directly rather than constructing it as a string.

```javascript
let data = {};
if(options && options.limit) {
    data.limit = options.limit;
}
$.ajax('/api/news/', { data: data, cache: false });
```

https://api.jquery.com/jquery.ajax/

PR #1175, opened 2017-07-17
Implement Stannis's Cavalry & Northern Encampment

DukeTax:


server/game/cards/locations/06/northernencampment.js
```diff
@@ -0,0 +1,38 @@
+const DrawCard = require('../../../drawcard.js');
+
+class DominanceTracker {
+    constructor(game, card) {
+        this.hasWonDominanceThisRound = undefined;
+        game.on('onDominanceDetermined', (event, winner) => this.hasWonDominanceThisRound = card.controller === winner ? true : false);
```
ystros: Extremely minor thing, but `card.controller === winner` will return a boolean directly. No need to do the `? true : false`

DukeTax: I love minor tips like this, thx! :)

cryogen: Just be careful, javascript is weird. :)

PR #1179, opened 2017-07-17
Pass the deck name when saving it

cryogen:
Pass the deck name when saving it as stats use it.  Patch existing decks which don't have a faction name with value - will require fixing in stats too later


server/game/player.js
```diff
@@ -934,6 +934,7 @@ class Player extends Spectator {
         this.deck.selected = true;
 
         this.faction.cardData = deck.faction;
+        this.faction.cardDat.name = deck.faction.name;
```
ystros: Typo

PR #1182, opened 2017-07-18
Fix uniqueness checks for take control.

### LINKED ISSUE #1170: Players can still play copies of uniques if their opponent controls them
If an opponent takes control of a player's unique card, that player cannot play additional copies of that unique card. However, it is currently possible for the owner to play additional copies of the card which the opponent has taken.

Creates kind of an awkward situation since this rule isn't well known amongst casual players.

ystros:
The checks applied to putting unique cards into play were not being
checked on take control. Thus, players were able to take control of
cards in their dead pile. Additionally, checks were not being made that
an opponent had taken control of your card, allowing players to put out
a new copy of a unique card already stolen by the opponent.

Fixes #1170.

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -694,6 +694,10 @@ class Game extends EventEmitter {
             return;
         }
 
+        if(!newController.canPutIntoPlay(card)) {
+            return;
+        }
+
         this.applyGameAction('takeControl', card, card => {
             oldController.removeCardFromPile(card);
             oldController.allCards = _(oldController.allCards.reject(c => c === card));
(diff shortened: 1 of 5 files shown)


ystros:
I killed the unit tests for `canPutIntoPlay` because:
1. the cases are more or less covered by the fuller integration tests already
2. in order to keep them as more-or-less pure unit tests, I would have to extract a bunch of methods (basically each variable in the new implementation) into separate methods on player, and I didn't want to pollute Player even more.

If you want, I can add them back in and make them slightly more integrationy (have to put card objects in the right locations instead of stubbing out methods), but it didn't seem worth it.

PR #1184, opened 2017-07-18
Random bunch of fixes and converts

DukeTax:


server/game/cards/locations/03/towerofthehand.js
```diff
@@ -10,32 +10,24 @@ class TowerOfTheHand extends DrawCard {
                 ability.costs.kneelSelf(),
                 ability.costs.returnToHand(card => this.isParticipatingLannister(card))
             ],
+            target: {
+                activePromptTitle: 'Select a character',
+                cardCondition: (card, context) => card.location === 'play area' && card.getType() === 'character' && card.controller !== this.controller &&
+                                                  card.getPrintedCost() < context.costs.returnedToHandCard.getPrintedCost()
```
ystros: Referring to the cost in the target here may not actually work (even though I'd like it to). Did you test this one locally?

DukeTax: I just now tested it, and it works fine. :)

PR #1186, opened 2017-07-18
Implement The Queen of Thorns (House of Thorns) & Desmera Redwyne

DukeTax:


server/game/cards/characters/09/desmeraredwyne.js
```diff
@@ -0,0 +1,36 @@
+const DrawCard = require('../../../drawcard.js');
+
+class DesmeraRedwyne extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onRemovedFromChallenge: () => true
```
ystros: To ensure that the player gets a useful prompt, could you add a title for this event to the top of triggeredabilitywindowtitles.js? Most of the time the player will be doing the removing and know from context which character is being removed, but it would be nice to have that info in the prompt title.

PR #1189, opened 2017-07-19
Allow player to choose pillage order

### LINKED ISSUE #909: Order of Pillage and indicator of which character is doing the Pillaging
If possible could you add a prompt when there are multiple Pillage characters in a challenge for the order of their Pillage triggers. This matters when you have characters that trigger off of the cards pillaged and you have some way to stack ("The Bear and the Maiden Fair") or filter (Tywin Lannister (LoCR version)) the opponent's deck. An example of this would be:

You win a challenge with Black Winds Crew (Pillage, but no special abilities), Gregor's Mauraders (Pillage, ability that triggers off pillaging an Event), and Ser Gregor (Pillage, ability that triggers off pillaging a character) with Tywin Lannister (LoCR version) (When a card is discarded from opponent's deck, look at the top 2 and choose which to Pillage) on the board. In this situation it is important to be able to trigger the Black Winds Crew first so that when you look at the top two cards you can decide who to pillage with next in case you see a character you want to kill or an event.

Thanks!

ystros:
* Processes keywords in sequence instead of processing all keywords for a given card before moving to the next card.
* If there is a single card with keyword Pillage, pillage will automatically occur as before.
* If there are multiple cards with keyword Pillage, the player will be prompted to select the order to process pillage. If they click 'Done' without selecting the order, then pillage will occur in the default order (the order the characters were declared as attackers). Otherwise, pillage will be processed in the order they select.

Fixes #909.

### DIFF
--- a/server/game/gamesteps/challenge/challengeflow.js
+++ b/server/game/gamesteps/challenge/challengeflow.js
@@ -5,7 +5,7 @@ const SimpleStep = require('../simplestep.js');
 const ChooseStealthTargets = require('./choosestealthtargets.js');
 const ApplyClaim = require('./applyclaim.js');
 const ActionWindow = require('../actionwindow.js');
-const GameKeywords = require('../../gamekeywords.js');
+const KeywordWindow = require('../keywordwindow.js');
 
 class ChallengeFlow extends BaseStep {
     constructor(game, challenge) {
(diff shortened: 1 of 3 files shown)


ystros:
Rebased against master locally, this passes the new eslint rules.

PR #1193, opened 2017-07-19
Prompt the defending player to apply claim

DukeTax:
As discussed in chat, this PR prompts the defending player to apply claim, instead of the attacker. By having the waiting prompt be identical to that of triggered abilities, the attacker will not be able to tell whether the defender gets prompted to play Vengeance for Elia immediately after applying claim, thereby preventing leaking of crucial game state.

### DIFF
--- a/server/game/gamesteps/challenge/challengeflow.js
+++ b/server/game/gamesteps/challenge/challengeflow.js
@@ -221,20 +221,20 @@ class ChallengeFlow extends BaseStep {
         }
 
         this.challenge.claim = this.challenge.getClaim();
-        this.game.promptWithMenu(this.challenge.winner, this, {
+        this.game.promptWithMenu(this.challenge.loser, this, {
             activePrompt: {
-                menuTitle: 'Perform before claim actions',
+                menuTitle: 'Claim will be applied',
                 buttons: [
-                    { text: 'Apply Claim', method: 'applyClaim' },
-                    { text: 'Continue', method: 'cancelClaim' }
+                    { text: 'Continue', method: 'applyClaim' },
+                    { text: 'Cancel', method: 'cancelClaim' }
                 ]
             },
-            waitingPromptTitle: 'Waiting for opponent to apply claim'
+            waitingPromptTitle: 'Waiting for opponents'
         });
     }
 
     applyClaim(player) {
-        if(player !== this.challenge.winner) {
+        if(player !== this.challenge.loser) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


DukeTax:
@cryogen Let me know if there's any chance of this getting merged, before I convert all those tests. :)

server/game/gamesteps/challenge/challengeflow.js
```diff
@@ -221,20 +221,20 @@ class ChallengeFlow extends BaseStep {
...
-        this.game.promptWithMenu(this.challenge.winner, this, {
+        this.game.promptWithMenu(this.challenge.loser, this, {
             activePrompt: {
-                menuTitle: 'Perform before claim actions',
+                menuTitle: 'Claim will be applied',
                 buttons: [
-                    { text: 'Apply Claim', method: 'applyClaim' },
-                    { text: 'Continue', method: 'cancelClaim' }
+                    { text: 'Continue', method: 'applyClaim' },
+                    { text: 'Cancel', method: 'cancelClaim' }
```
ystros: I understand the information hiding arguments for prompting the losing player instead of the winner, but I'm not sure allowing them to cancel claim application is a good idea. Seems like a good way to spawn even more arguments (trolls just refusing to allow claim to be applied). It also could cause confusion - if you want to play Vengeance for Elia, for instance, you should click 'Continue', but one might think you should click 'Cancel' because you're replacing claim (in which case Vengeance won't fire because claim wasn't applied).

I'd propose that there should be only one button - one that applies claim. And perhaps as a future PR, make this prompt optional for decks that don't care about information hiding during claim apply. I defer to @cryogen though.

ystros:
Re: fixing tests, assuming cryo is okay with this change, try to update them to work similar to `selectFirstPlayer`. E.g.
```javascript
// normal challenge declaration
this.applyClaim();
// rest of test
```

Look at test/helpers/gameflowwrapper.js, add the function there, then add the method name to ProxiedGameFlowWrapperMethods at the top of test/helpers/integrationhelper.js. That should make any future changes much easier / less painful.

server/game/gamesteps/challenge/challengeflow.js
DukeTax: I agree with everything here. I basically kept the cancel-option to keep this PR as lean as possible (so as to not detract from my intentions) and because some people mentioned they like it for botched challenges. The latter is probably better fixed by the attackers manually setting their claim to 0, something we can easily communicate to the users via the news. I'll gladly get rid of it if cryogen gives the green light for this PR.

DukeTax:
With the new timer window this PR has no use, so closing.

PR #1202, opened 2017-07-20
Implement Maester at the Rock

DukeTax:


server/game/cards/characters/08/maesterattherock.js
```diff
@@ -0,0 +1,20 @@
+const DrawCard = require('../../../drawcard.js');
+
+class MaesterAtTheRock extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onCardPlayed: event => event.card.isFaction('lannister') && event.card.controller === this.controller
```
ystros: We should explicitly check that the location of the card is the discard pile. This ability specifically can't be triggered if, for instance, the event was played during Annals of Castle Black and thus removed from the game.

http://thronesdb.com/card/08010

PR #1205, opened 2017-07-20
Disable agenda popup if there are no cards

ystros:
Fixes THRONETEKI-112


ystros:
This actually should fix itself once you deploy the corresponding server side changes (comes from `bannerCards` being missing since server side serialization hasn't been deployed) but it may be a good safety check regardless.

PR #1211, opened 2017-07-21
Better crash recovery

ystros:
* When the resolution of ability throws an exception anywhere (paying costs, targeting, in the handler, in prompts nested under the handler, etc), the error will be reported and the ability will be cancelled.
* When a prompt checks an ability for eligibility and the eligibility check throws an exception, the prompt will report the error and remove that ability from the prompt.

There may be cleanup paths I'm missing, particularly around ability resolution, but having the game in a slightly weird state but playable is preferable to total crash, which can happen currently.

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -73,6 +73,10 @@ class Game extends EventEmitter {
         this.pushAbilityContext('framework', null, 'framework');
     }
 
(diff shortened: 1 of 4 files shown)


ystros:
Rebased against master locally, this passes the new eslint rules.

Issue #837, opened 2017-05-02
attachment-induced traits, affiliation, etc. should remain long enough to trigger kill reactions


ystros:
**Scenario:** A character has Ward attached, gaining the Stark affiliation. That character is killed. Reactions to a Stark character being killed such as WotN Catelyn Stark should be able to trigger.

Currently, however, such reactions do not trigger because the Ward attachment is discarded when the character is killed but before the reaction triggers, making the game inappropriately think it wasn't a Stark character that was killed.

PR #1214, opened 2017-07-21
Implement Ashemark

DukeTax:


server/game/cards/locations/08/ashemark.js
```diff
@@ -0,0 +1,35 @@
+const _ = require('underscore');
+
+const DrawCard = require('../../../drawcard.js');
+
+class Ashemark extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onPhaseStarted: () => {
+                    this.gold = this.tokens['gold'];
```
ystros: I understand why but 😬

DukeTax: Might be a good idea to store such info on context for self-sacrificing cards?

ystros: Something like that. It's very similar to #837, just with sacrifice instead of kill.

PR #1215, opened 2017-07-21
Implement Queen's Men

DukeTax:
@ystros: As we talked about in chat, showing the hand through a prompt instead of a chat message would help with an eventual melee implementation. Not sure if I like how this turned out though. Clicking a character card has the same effect as clicking Done, which seems weird but I don't know a way around this. And if you decide to trigger then then-effect, the cost-effect flow is reversed from the usual. On the other hand, this way does take relatively few clicks.

Another possibility: reveal the hand via chat message. Then, if you have a standing non-Bara character, prompt whether to kneel one, and if so, only prompt non-character cards from the opponents hand to be discarded. Showing the hand through a chat message can't be used in melee obviously, unless we would implement player specific chat messages (and the ability to whisper in melee would be cool aside from this). Aside from that, this is much how we would do other cards, with costs paid before deciding on effects, but takes either 0,1 or 3 clicks to resolve after triggering. The current implementation takes 1 or 2 clicks to resolve after triggering.

Finally, is there a way to pass a card object through a menuPrompt instead of a cardId as I do now?

### DIFF
--- a/server/game/cards/characters/08/queensmen.js
+++ b/server/game/cards/characters/08/queensmen.js
@@ -59,7 +59,7 @@ class QueensMen extends DrawCard {
         toKneel.controller.kneelCard(toKneel);
         toDiscard.owner.discardCard(toDiscard);
         this.game.addMessage('{0} then kneels {1} to discard {2} from {3}\'s hand',
-                              this.controller, toKneel, toDiscard, toDiscard.owner);
+            this.controller, toKneel, toDiscard, toDiscard.owner);
 
         return true;
     }
--- a/server/game/cards/characters/08/queensmen.js
+++ b/server/game/cards/characters/08/queensmen.js
@@ -1,0 +1,70 @@
+const DrawCard = require('../../../drawcard.js');
+
+class QueensMen extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardEntersPlay: event => event.card === this && event.playingType === 'marshal'
+            },
+            handler: () => {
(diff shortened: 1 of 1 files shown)


DukeTax:
Sorry about this long ass rant for 1 shitty card.

ystros:
I think implementation wise this is fine for now. It certainly is awkward, but most of the other solutions involve more clicks or a new prompt type.

I'd like to experiment eventually with a prompt that shows the cards directly (for things like this and search prompts) and allows them to be selected like any other card. That would make a prompt like this a bit less awkward. But I know there's been resistance to this type of prompt. 3-ish minute mockup:
mockup

For the menu prompt, there isn't a way to get the object other than how you've done it. We should put together a new 'card menu' prompt at some point that does the card look up for you (and could be then converted to the style prompt above if we ever decide to go in that direction).

server/game/cards/characters/08/queensmen.js
```diff
@@ -0,0 +1,70 @@
...
+        let opponent = this.game.getOtherPlayer(this.controller);
+        this.game.addMessage('{0} uses {1} to look at {2}\'s hand', this.controller, this, opponent);
+        return true;
+    }
+
+    kneelToDiscard(p, toKneel, toDiscard) {
+        toKneel.controller.kneelCard(toKneel);
+        toDiscard.owner.discardCard(toDiscard);
+        this.game.addMessage('{0} then kneels {1} to discard {2} from {3}\'s hand',
+                              this.controller, toKneel, toDiscard, toDiscard.owner);
```
ystros: This will fail the new eslint rules if merged as-is.

PR #1218, opened 2017-07-22
Convert remaining icon cards to cost/target/icon API

DukeTax:
A slight issue: the waiting prompt whenever you're selecting an icon will say "Waiting on opponent to use undefined".


ystros:
For the prompt titles, if all instances of choosing an icon are generated by card abilities (I think they are), then you could pass in the source card to the prompt. E.g.
```javascript
handler: () => {
    this.game.promptForIcon(this.controller, this, icon => {
        // ...
    });
}
```

That's not super consistent with the rest of our prompts, which look more like:
```javascript
handler: () => {
    this.game.promptForIcon(this.controller, {
        source: this,
        onSelect: icon => {
            // ...
        }
    });
}
```

But I'd only do the latter if there are more properties that we should be passing through (e.g. activePromptTitle) but aren't currently.

PR #1219, opened 2017-07-22
Remove "winner determined" action window

ystros:
The "winner determined" action window was put into place to allow a
pause to play in-hand event cards. Now that nearly all events in hand
prompt the player directly, the window is no longer needed.
Additionally, the presence of the window leads newer players to believe
they can ambush cards or play Action events in this pause, which is not
allowed by the rules.


ystros:
This will conflict client-side with the upcoming timed windows. Happy to wait until that's merged and resolve conflicts myself, though the changes are fairly minor overall.

PR #1221, opened 2017-07-23
Add validation rules around draft cards

ystros:
* Allows Draft agendas to be selected.
* Adds validation rules for the Draft agendas
* Indicates whether a deck contains Draft cards if it is otherwise valid.


cryogen:

cryogen:
Add to whitelist

PR #1227, opened 2017-07-23
Convert and fix The Queen of Thorns (Core)

### LINKED ISSUE #1224: Queen of Thorns lets you drop 7 costers into play
As title really. Opponent was able to use Queen of Thorns to put Renly into play, but Queen has a restriction of 6 cost or lower.

DukeTax:
* Convert to use target API and only allow printed cost 6 or lower characters to be put into play.

* Fixes #1224.

### DIFF
--- a/server/game/cards/characters/01/thequeenofthorns.js
+++ b/server/game/cards/characters/01/thequeenofthorns.js
@@ -6,23 +6,16 @@ class TheQueenOfThorns extends DrawCard {
             when: {
                 afterChallenge: (event, challenge) => challenge.winner === this.controller && challenge.isParticipating(this) && challenge.challengeType === 'intrigue'
             },
-            handler: () => {
-                this.game.promptForSelect(this.controller, {
-                    activePromptTitle: 'Select a character to put in play',
-                    source: this,
-                    cardCondition: card => card.location === 'hand' && card.getCost() && card.getType() === 'character' && card.isFaction('tyrell'),
-                    onSelect: (p, card) => this.onCardSelected(p, card)
-                });
+            target: {
+                activePromptTitle: 'Select a character to put into play',
+                cardCondition: card => card.location === 'hand' && card.controller === this.controller &&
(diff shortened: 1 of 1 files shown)


cryogen:

server/game/cards/characters/01/thequeenofthorns.js
```diff
@@ -6,24 +6,17 @@ class TheQueenOfThorns extends DrawCard {
...
-            handler: () => {
-                this.game.promptForSelect(this.controller, {
-                    activePromptTitle: 'Select a character to put in play',
-                    source: this,
-                    cardCondition: card => card.location === 'hand' && card.getCost() && card.getType() === 'character' && card.isFaction('tyrell'),
-                    onSelect: (p, card) => this.onCardSelected(p, card)
-                });
+            target: {
+                activePromptTitle: 'Select a character to put into play',
+                cardCondition: card => card.location === 'hand' && card.controller === this.controller &&
```
zacchiro: "card.controller === this.controller" shouldn't be needed, given the card is in hand, right?
I guess there could be bugs making that assumption false. But, in that case too, I'd rather not have the check because it will hide the bugs.

cryogen:
Add to whitelist

server/game/cards/characters/01/thequeenofthorns.js
DukeTax: I'm not 100% sure, but wouldn't that allow you to target a card in an opponents hand (provided you happen to click one that pass the other conditions)? Or is that the opponent's hand not a legal target area anyway?

ystros: There **may** be something client side that disallows clicking on facedown / hidden / opponent hand cards, but I'm not sure where it happens. We've also recently talked about being able to click opponent's hand cards as an alternate UI for things like His Viper Eyes and other "look at hand" abilities (e.g. make those cards not facedown and then do a normal card select prompt).

I don't think the controller check hurts anything here, but I'm not sure we've been consistent about it either.

PR #1228, opened 2017-07-23
Allow Lady-In-Waiting to only target cards that can be duped (owned a…

### LINKED ISSUE #1225: Lady in Waiting can dupe opponent's cards
I don't know if it's because it's programmed similar to Bodyguard, but when using Lady in Waiting as a dupe, it allows you to dupe your opponent's ladies.

However although it's legal to Bodyguard opponent's cards, you cannot dupe a character you do not own, so Lady in Waiting is illegal to use here. Also illegal to use it to dupe ladies you have stolen control of (as you don't own them).

DukeTax:
* Fixes #1225.


cryogen:

PR #1230, opened 2017-07-23
implement Aeron Damphair (TIMC)

zacchiro:
to that end, introduce a new onCardSaved event, raised when a card (of any
kind) is saved (for any reason: kill, discard, etc.)

### DIFF
--- a/server/game/cards/characters/04/aerondamphair.js
+++ b/server/game/cards/characters/04/aerondamphair.js
@@ -1,0 +1,27 @@
+const DrawCard = require('../../../drawcard.js');
+
+class AeronDamphair extends DrawCard {
+
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardSaved: event => event.card.getType() === 'character'
+            },
+            handler: context => {
+                let card = context.event.card;
+                if(card.kneeled) {
+                    card.controller.standCard(card);
+                    this.game.addMessage('{0} uses {1} to stand {2}', this.controller, this, card);
+                } else {
+                    card.controller.kneelCard(card);
+                    this.game.addMessage('{0} uses {1} to kneel {2}', this.controller, this, card);
+                }
+            }
+        });
+    }
+
+}
+
+AeronDamphair.code = '04071';
+
+module.exports = AeronDamphair;
--- a/server/game/event.js
+++ b/server/game/event.js
@@ -29,6 +29,7 @@ class Event {
 
         card.markAsSaved();
         this.cards = _.reject(this.cards, c => c === card);
+        card.game.raiseMergedEvent('onCardSaved', { card: card });
 
         if(_.isEmpty(this.cards)) {
             this.cancel();
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -848,6 +848,7 @@ class Player extends Spectator {
         if(!card.dupes.isEmpty() && allowSave) {
             if(this.removeDuplicate(card)) {
                 this.game.addMessage('{0} discards a duplicate to save {1}', this, card);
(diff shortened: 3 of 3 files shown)


zacchiro:
this is a first attempt at implementing Aeron (TIMC). A couple of caveats, RFC:

1. I'm not sure I captured all possible cases of save, double-checking is needed
2. I haven't found a way to add the name of the character saved to the dialog labels, because the "context" parameter is received too late. Any tip about how to fix this?

DukeTax:
I tried to implement Damphair a little while back but couldn't resolve a certain issue (https://github.com/cryogen/throneteki/pull/1060). I think adding a title to gamesteps/triggeredabilitywindowtitles.js should work in making clear which character was saved.

server/game/cards/characters/04/aerondamphair.js
```diff
@@ -0,0 +1,29 @@
+const DrawCard = require('../../../drawcard.js');
+
+class AeronDamphair extends DrawCard {
+
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardSaved: event => event.card.getType() === 'character'
+            },
+            choices: {
```
DukeTax: Using choices here is not strictly correct. While the card seems to offer a choice, the fact that abilities can only be triggered if you would change the game state dictates that after triggering you have to either kneel the card (if standing) or stand it (if kneeling)

zacchiro: ah, good point. And it also simplifies things, which is good :)

I briefly looked at the other PR (thanks for the pointer!) and AFAICT this PR already implements the event naming/granularity that @ystros wanted, as well as the duplicate saving part.

I'll check for other corner cases mentioned there before updating this PR again.

zacchiro:
So, here's my status report about this PR re the previous iteration (#1060) by @DukeTax and feedback there by @ystros.

- event naming/granularity should be as desired (onCardSaved)
- automatic saves using duplicates are taken into account
- I've reviewed all other uses of onCharacterKilled and found no occurrence where manual kill cancel is going on
- I've tried Valar v. dupes, and dupes did automatically save, with Aeron triggering each time

I'm only puzzled by the fact that in event.js @DukeTax moved the save code to the callback of onCardSaved, while I just fire the event before calling markAsSaved(). It's likely I'm missing something there. But moving the save code to the callback break a lot of test that IMHO shouldn't break here. I appeal to the Drowned God for further guidance (and review).

DukeTax:
Moving the save code to the callback only matters for interrupts I think, of which there aren't any of this kind (and good chance there will never be) but I think I wanted to future proof it, although it shouldn't matter. But if you do it like this, the event should be raised after the save code. As it is a reaction the save should be fully resolved before Aeron initiates. Rest looks good. :)

ystros:
Yeah, if there's ever an interrupt ability for cards being saved, we'll have to revisit this and wrap the handler with the event.

server/game/event.js
```diff
@@ -27,6 +27,7 @@ class Event {
             return;
         }
 
+        card.game.raiseMergedEvent('onCardSaved', { card: card });
```
ystros: Per @DukeTax's comment, this should be moved after the card has been marked as saved and removed from the card list. It doesn't *strictly* matter since the event is fired asynchronously and the mark-as-saved is synchronous currently. However, if mark-as-saved is ever changed to be asynchronous, it could introduce a subtle bug (reactions firing before the thing actually happens).

zacchiro:
Thanks to both of you for the reviews! I've now updated this to fire the event after the card is marked as saved, as requested.
Let me know if anything else is missing.

PR #1233, opened 2017-07-24
Update card implementations

DukeTax:
* Convert card implementations to cost/target API's.
* Simplify event property functions by fetching the card in question from the context object.
* Standardise chat messages and action titles.

### DIFF
--- a/server/game/cards/characters/04/eliasand.js
+++ b/server/game/cards/characters/04/eliasand.js
@@ -7,25 +7,19 @@ class EliaSand extends DrawCard {
                 afterChallenge: (event, challenge) => this.controller === challenge.loser
             },
             limit: ability.limit.perPhase(2),
-            handler: () => {
-                this.game.promptForSelect(this.controller, {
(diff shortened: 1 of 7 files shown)


server/game/cards/characters/04/eliasand.js
```diff
@@ -7,26 +7,20 @@ class EliaSand extends DrawCard {
...
             limit: ability.limit.perPhase(2),
-            handler: () => {
-                this.game.promptForSelect(this.controller, {
-                    activePromptTitle: 'Select character',
-                    source: this,
-                    cardCondition: card => card.location === 'play area' && card.getType() === 'character',
-                    onSelect: (p, card) => this.onCardSelected(p, card)
-                });
+            target: {
+                activePromptTitle: 'Select a character to gain stealth',
```
cryogen: I think we've been standardising on 'select a character' for stuff like this to not have the prompt get super long or wordy and keep it consistent

PR #1235, opened 2017-07-24
Calculate Crossing STR based on challenge number

### LINKED ISSUE #1169: Second LOTCrossing challenge temporarily gives -1 STR
There is a small window after declaring attackers and before the challenge counter gets updated. In that window the character gets Crossing's -1 STR and could then die for example to Blood of the Dragon.

First challenge with Tyrion, second with Tumblestone Knight:

```
DukeTax is initiating a  challenge
DukeTax uses Tyrion Lannister to gain 2 gold as an intrigue challenge has been declared
DukeTax has initiated a  challenge with strength 2
does not defend the challenge
DukeTax won a  challenge 2 vs 0
DukeTax has gained 1 power from an unopposed challenge
DukeTax is initiating a  challenge
Tumblestone Knight is killed as its STR is 0
DukeTax kills Tumblestone Knight
DukeTax has initiated a  challenge with strength 0
```

Edit: probably temporarily gives +2 STR on the 4th challenge as well.

ystros:
Previously, Lord of the Crossing's STR bonuses were calculated based on
the number of challenges initiated by the player. This lead to a bug
where after attackers were picked but before the challenge was
initiated, the attackers would get the -1 STR penalty. If the opponent
had Blood of the Dragon at the time, this could kill the attackers.

Now, the challenge number is calculated up front and remains constant for
the entire challenge, preventing such bugs.

Fixes #1169

### DIFF
--- a/server/game/cards/agendas/thelordofthecrossing.js
+++ b/server/game/cards/agendas/thelordofthecrossing.js
@@ -17,7 +17,7 @@ class TheLordOfTheCrossing extends AgendaCard {
     }
 
     challengeBonus() {
-        var numChallenges = this.controller.getNumberOfChallengesInitiated();
+        let numChallenges = this.game.currentChallenge.number;
         if(numChallenges === 1) {
             return -1;
         }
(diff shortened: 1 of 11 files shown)


mplain:
If the challenge is cancelled (no attackers are declared), it doesn't count, right?

ystros:
Correct. The challenge count isn't incremented for the player if no attackers are declared, so choosing a challenge type, clicking 'Done' without choosing an attacker, then choosing a challenge type again still counts as challenge #1.

PR #1236, opened 2017-07-25
Recalculate effects when traits change

### LINKED ISSUE #1027: Knighted Arianne doesn't get renown from A Tourney For The King
Just seen in a game i was spectating:

- Arianne Martel has Knighted as attachment
- plot is A Tourney For The King
- Arianne attacks and wins in a power challenge
- but didn't get power from renown

I see no obvious bugs in the implementation of either Knighted or A Tourny For The King, so — if reproducible — it might be a weird (timing?) interaction among adding the trait and looking it up from a plot?

ystros:
Previously, if an effect were already in play that was limited to cards
with a certain trait, and a card already in play gained or lost that
trait, the effect would not be applied or removed from that card. For
example, characters that gained the 'Knight' trait while Tourney for the
King was revealed did not gain the renown keyword.

Now, when a card gains or loses a trait, effects for that card are
recalculated in a similar manner to the way they are recalculated for
faction changes.

Fixes #1027.

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -421,7 +421,7 @@ class BaseCard {
     }
 
     addTrait(trait) {
-        var lowerCaseTrait = trait.toLowerCase();
+        let lowerCaseTrait = trait.toLowerCase();
 
         if(!lowerCaseTrait || lowerCaseTrait === '') {
             return;
@@ -432,6 +432,8 @@ class BaseCard {
         } else {
             this.traits[lowerCaseTrait]++;
         }
+
+        this.game.raiseMergedEvent('onCardTraitChanged', { card: this });
     }
 
     addFaction(faction) {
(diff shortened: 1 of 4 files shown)


mplain:
Note that this should apply only to constant abilities, but not to lasting effects, e.g. [Bless Him With Salt](http://thronesdb.com/card/04092) remains in effect even if [Drowned God's Blessing](http://thronesdb.com/card/02112) is discarded.

ystros:
@mplain - the code specifically checks that the effect has duration 'persistent' (what we ended up calling constant abilities internally), so I believe we're okay re: lasting effects.

There was a bug at one point where the strength provided by Abandoned Stronghold was not properly fixed upon resolution, but it has been previously corrected.

Thanks for the rules heads up!

PR #1238, opened 2017-07-25
Reword action window prompts

DukeTax:
* Remove reactions from the title and use singular 'action' to better match their use.
* Pass instead of Done in the button matches the implementation of the reaction and interrupt windows and mirrors the language used in the RRG.

### DIFF
--- a/server/game/gamesteps/actionwindow.js
+++ b/server/game/gamesteps/actionwindow.js
@@ -22,9 +22,9 @@ class ActionWindow extends PlayerOrderPrompt {
 
     activePrompt() {
         return {
-            menuTitle: 'Any actions or reactions?',
+            menuTitle: 'Initiate an action, or pass',
             buttons: [
-                { text: 'Done' }
+                { text: 'Pass' }
             ],
             promptTitle: this.title
         };
--- a/server/game/gamesteps/actionwindow.js
+++ b/server/game/gamesteps/actionwindow.js
@@ -22,7 +22,7 @@ class ActionWindow extends PlayerOrderPrompt {
 
     activePrompt() {
         return {
-            menuTitle: 'Initiate an action, or pass',
+            menuTitle: 'Initiate an action',
             buttons: [
                 { text: 'Pass' }
             ],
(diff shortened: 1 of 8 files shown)


server/game/gamesteps/actionwindow.js
```diff
@@ -22,9 +22,9 @@ class ActionWindow extends PlayerOrderPrompt {
 
     activePrompt() {
         return {
-            menuTitle: 'Any actions or reactions?',
+            menuTitle: 'Initiate an action, or pass',
             buttons: [
-                { text: 'Done' }
+                { text: 'Pass' }
```
zacchiro: This made a whole lot of integration tests explode, as they're expecting to find "Done" in the dialog window.
I guess you were waiting for feedback on the proposal before fixing those, but just wanted to note down here that, if accepted, they'll need to be fixed before this can go in.

DukeTax: Yep. Depending on how much work fixing those tests would be, I might change it back to "Done". It was mostly about not asking for reactions in action windows.

server/game/gamesteps/actionwindow.js
```diff
@@ -22,9 +22,9 @@ class ActionWindow extends PlayerOrderPrompt {
 
     activePrompt() {
         return {
-            menuTitle: 'Any actions or reactions?',
+            menuTitle: 'Initiate an action, or pass',
```
ystros: I don't think we need to explicitly call out "or pass" here. I'd either title it just "Initiate an action" or "Any actions?", not particular on which.

server/game/gamesteps/actionwindow.js
ystros: I like that it explicitly says Pass now. Changing `GameFlowWrapper.skipActionWindow` should hit 99% of the failures, there may be a couple that need to be updated by hand.

server/game/gamesteps/actionwindow.js
DukeTax: I like the latter for Reaction/Interrupt windows where multiple abilities could be listed. For actions windows the former has my preference, as the latter could possibly be construed as permission to initiate multiple actions in a row.

PR #1243, opened 2017-07-25
card messages: switch to present tense

zacchiro:
as per log message conventions

### DIFF
--- a/server/game/cards/plots/06/duel.js
+++ b/server/game/cards/plots/06/duel.js
@@ -37,7 +37,7 @@ class Duel extends PlotCard {
         }
 
         this.targets = cards;
-        this.game.addMessage('{0} has chosen {1} as the targets for {2}', player, cards, this);
+        this.game.addMessage('{0} chooses {1} as the targets for {2}', player, cards, this);
 
         this.game.promptWithMenu(player, this, {
             activePrompt: {
--- a/server/game/cards/events/06/thenorthremembers.js
+++ b/server/game/cards/events/06/thenorthremembers.js
@@ -50,13 +50,13 @@ class TheNorthRemembers extends DrawCard {
 
     onCardSelected(player, card) {
         this.selections.push({ player: player, card: card });
-        this.game.addMessage('{0} has selected {1} to sacrifice', player, card);
+        this.game.addMessage('{0} selects {1} to sacrifice', player, card);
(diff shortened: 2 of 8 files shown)


zacchiro:
Here are a few additional occurrences of present perfect in card log messages which I haven't included in this PR as I'm not sure if they should be converted or not.

- characters/02/brothelmadame.js:        this.game.addMessage('{0} has now paid {1} and can initiate {2} challenges this phase', event.source, event.target, 'military');
- plots/01/wildfireassault.js:                this.game.addMessage('{0} has {1} killed because of {2}', player, toKill, this);
- plots/02/politicaldisaster.js:                this.game.addMessage('{0} has {1} discarded for {2}', player, toDiscard, this);

Suggestions, especially from native speakers, are very welcome.

server/game/cards/plots/06/duel.js
```diff
@@ -37,7 +37,7 @@ class Duel extends PlotCard {
         }
 
         this.targets = cards;
-        this.game.addMessage('{0} has chosen {1} as the targets for {2}', player, cards, this);
+        this.game.addMessage('{0} choses {1} as the targets for {2}', player, cards, this);
```
DukeTax: Typo. :) Rest is (y).

zacchiro: good catch! fixed

ystros:
👍

ystros:
Oops, merged slightly too quickly without seeing your question. For the ones you asked about, I'd suggest:
characters/02/brothelmadame.js: this.game.addMessage('{0} pays {1} and can now initiate {2} challenges this phase', event.source, event.target, 'military');

Wildfire and Disaster are kind of awkward no matter how I try to phrase it.
"{0} kills/discards {1} because of {2}"
"{0} kills/discards {1} for {2}"
"{0} kills/discards {1} from {2}"

From an American standpoint, all of these are understandable. I believe "because of" or "for" are the most correct. We should be consistent in which we use though.

zacchiro:
np, i'll submit a separate PR for the remaining ones. Thanks for your quick review!

zacchiro:
The remaining messages have now been fixed in #1245 .
I went for the "for" variant as: 1) it's more concise, and 2) was already used in some other messages of the same card.

PR #1246, opened 2017-07-25
Remove the "new this phase" highlight when character leaves play

### LINKED ISSUE #1244: Some card border highlights don't get removed, ever

DukeTax:
* Fixes #1244.

### DIFF
--- a/server/game/drawcard.js
+++ b/server/game/drawcard.js
@@ -331,6 +331,7 @@ class DrawCard extends BaseCard {
         this.power = 0;
         this.wasAmbush = false;
         this.inChallenge = false;
+        this.new = false;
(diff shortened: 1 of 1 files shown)


server/game/drawcard.js
```diff
@@ -331,6 +331,7 @@ class DrawCard extends BaseCard {
         this.power = 0;
         this.wasAmbush = false;
         this.inChallenge = false;
+        this.new = false;
```
ystros: As long as we're doing this, might as well reset the other UI flags. `resetForChallenge()` will clear stealth info, `clearDanger()` will clear the inDanger and saved flags.

PR #1252, opened 2017-07-25
Convert phase events to merged style

ystros:


zacchiro:
This (and friends PR #1251 and #1254) looks good to me, thanks for working on this!
As a a related thing: once the migration is fully done, can we rename back raiseMergedEvent() to raiseEvent() ?

ystros:
Yup, once all of these PRs are merged, there will be no usages of the old raiseEvent and we can rename.

PR #1254, opened 2017-07-26
Convert challenge events to merged style

ystros:
Converts all of the remaining challenge events to merged events.

I don't particularly like using destructuring for parameters of the afterChallenge handlers. However, there are over 100 cards that use that event and updating each one manually would have been far more error prone and time consuming.

Finally, since we're already touching all of the afterChallenge events, this would be a good opportunity to rename the event. I was thinking of something like `onChallengeWinnerDetermined` as being more descriptive, but it's a bit long. Open to suggestions.


cryogen:
So TIL destructuring exists, and now have one more thing to hate about javascript :(

PR #1256, opened 2017-07-26
Auto time out cancellable/interruptable events

cryogen:


server/game/gamesteps/triggeredabilitywindow.js
```diff
@@ -35,8 +35,16 @@ class TriggeredAbilityWindow extends BaseAbilityWindow {
         return false;
     }
 
+    isCancellableEvent(player) {
+        let cancellableEvents = ['onCardAbilityInitiated', 'onClaimApplied'];
+
+        return !player.noTimer && this.abilityType === 'cancelinterrupt' && _.any(this.events, event => {
```
ystros: 'cancelinterrupt' is not quite right if we care about hiding information for Vengeance for Elia. Hand's Judgment and Treachery are in onCardAbilityInitiated:cancelinterrupt, but Vengeance is in onClaimApplied:interrupt.

server/game/gamesteps/triggeredabilitywindow.js
```diff
@@ -26,7 +26,7 @@ class TriggeredAbilityWindow extends BaseAbilityWindow {
     continue() {
         this.players = this.filterChoicelessPlayers(this.players || this.game.getPlayersInFirstPlayerOrder());
 
-        if(this.players.length === 0 || this.abilityChoices.length === 0) {
+        if(this.players.length === 0) {
```
ystros: Removing the abilityChoices check here may lead to empty prompts that you have to explicitly pass other than timed ones. Not 100% sure.

server/game/gamesteps/triggeredabilitywindow.js
```diff
@@ -50,8 +58,16 @@ class TriggeredAbilityWindow extends BaseAbilityWindow {
             if(abilityChoice.text !== 'default') {
                 title += ' - ' + abilityChoice.text;
             }
+
             return { text: title, method: 'chooseAbility', arg: abilityChoice.id, card: abilityChoice.card };
         });
+
+        if(this.isCancellableEvent(player)) {
+            buttons.push({ timer: true, method: 'pass', id: uuid.v1() });
```
ystros: Not sure I like the idea of pushing fake buttons onto the prompt in order to get the timer. Why not set this as a property at the prompt-level? e.g.
```javascript
activePrompt: {
    menuTitle: 'Any interrupts?',
    buttons: [
        // ...
    ],
    timer: { method: 'pass' }
}
```

Doing it that way, you could also generate the "I need more time" and "Don't ask again until X" buttons client-side as well, giving us more control over their placement / styling.

cryogen: I plan to refactor the whole way prompts and buttons work in the not too distant future so didn’t want to delve into that as part of that

ystros:
Couple of notes:
* Remember to add back in the green triangle to access window options during the game.
* Clicking "Pass" on a timed window does not stop the countdown - I haven't looked in depth at the client code but this could indicate there may be scenarios where I click a button, I have trouble sending the message to the server, and the timer runs out and also tries to send the 'pass' message too. Found a reproducible scenario:
> Player 1 is marshaling.
> Player 2 kneels one of their reducers (a valid move as they can activate actions during the other player's marshaling).
> Player 1 is prompted whether they want to cancel the reducer's ability. They click 'Pass' before the timer runs out. They continue marshaling.
> When Player 1's timer runs out, it acts as if they clicked 'Done' and moves to Player 2's marshaling.
* Styling for the bar seems a bit close to the plot deck at my resolution. 
screen shot 2017-07-26 at 6 47 07 pm
* A title for onClaimApplied should be added to triggeredabilitywindowtitles.js. Right now it just says "Any interrupts?" which is kind of unclear even though there's some slight context given challenge flow.
* When onClaimApplied's ability type is corrected from cancelinterrupt to just interrupt, the attacking player gets a timed prompt. I'm not sure that's necessary - to my knowledge, there's no cards we need to info hide on the attacker's side, most of the claim replacers other than Vengeance have to be in play (e.g. Mirri). Probably can be fixed by adding `player` to onClaimApplied (which is slightly weird but justifiable)
screen shot 2017-07-26 at 7 03 13 pm

Will add more as I find things.

cryogen:
> Clicking "Pass" on a timed window does not stop the countdown - I haven't looked in depth at the client code but this could indicate there may be scenarios where I click a button, I have trouble sending the message to the server, and the timer runs out and also tries to send the 'pass' message too. Found a reproducible scenario:

Fixed, this got broken by me trying to make the timer more efficient with react

> Saying "Don't ask again this round" seems to revert back to the previous behavior of doing untimed prompts while in hand. This seems like a good fallback and prevents situations where someone who doesn't care about ability cancels clicks "Don't ask again" but does want to play something like Vengeance for Elia.

Hmm, I hadn't really thought about that, it wasn't intentional.  Given that the window always prompts now, as you say I think this is Ok because there's no information leaked as the window always prompts (normally)

> Eliminating the abilityChoices.length === 0 check has definitely introduced some problems. Try:

Hrm, that's not going to be easy to fix the way I've done it

PR #1257, opened 2017-07-26
Implement Ritual of R'hllor and convert Loot

### LINKED ISSUE #1176: The Hand's Judgment vs Loot
HJ is free and cancels the effects of loot but loot has no chance to determine its cost. Essentially i should still be able to pay all your gold even if i dont discard your cards.

DukeTax:
* Fixes #1176.

### DIFF
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -1,5 +1,6 @@
 const _ = require('underscore');
 const ChooseCost = require('./costs/choosecost.js');
+const payXGoldPrompt = require('./costs/payxgoldprompt.js');
 
 const Costs = {
     /**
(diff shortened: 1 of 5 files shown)


server/game/costs.js
```diff
@@ -597,6 +598,37 @@ const Costs = {
...
+        return {
+            canPay: function(context) {
+                if(!opponentObj) {
+                    return context.player.gold >= 1;
+                }
+                return opponentObj.gold >= 1;
+            },
+            resolve: function(context, result = { resolved: false }) {
+                let gold = opponentObj ? opponentObj.gold : context.player.gold;
+                let list = [maxFunc(), gold];
```
ystros: This is non-intuitive. Why not calculate the limit here (you have enough info) and pass it into the prompt directly? The array doesn't seem to be used for anything else.

server/game/costs/payxgoldprompt.js
```diff
@@ -0,0 +1,41 @@
+const _ = require('underscore');
+
+const BaseStep = require('../gamesteps/basestep');
+
+class payXGoldPrompt extends BaseStep {
```
ystros: Class names should always begin with an uppercase letter. I'm surprised there's not a linting rule around this.

ystros:
Needs #1254 to be merged first.

PR #1259, opened 2017-07-27
Convert Time of Plenty and Early Frost to use an effect

DukeTax:


ystros:
I thought there may be a bug introduced by this:
```
Reveal Time of Plenty, effect applied drawPhaseCards += 1
Full round goes by
Start of new plot phase, drawPhaseCards is explicitly set to 2 in `Player.startPlotPhase`
New plot is revealed, Time of Plenty's effect is unapplied by drawPhaseCards -= 1
drawPhaseCards is at 1 for the new round
```

However, it seems to work fine and I'm not sure why. Just something to keep our eyes open for...

DukeTax:
Because of the condition property, the effect will already be unapplied when the marshalling phase starts?

ystros:
Oops, missed the condition. Yup, that'd be why! 👍

PR #1261, opened 2017-07-27
Revert controlled cards when they move

### LINKED ISSUE #1223: Ward + First Snow
Bit of a weird situation:

I played a character, my opponent used Ward to take control of that character.

I used First Snow, which returned that character to my hand (and discarded his Ward). All is as it should be.

I played the same character from hand, my opponent used a new Ward on the character. However the game would not give control of that character to my opponent. The Ward attached successfully but the character stayed on my side of the board, and was only able to be manipulated (challenges etc) by myself.

I attempted to use the /give-control command on the character, and the chat prompt confirmed I had done so but the character STILL didn't change control.

ystros:
Previously, if a card that had been taken control of was sacrificed or
returned to hand, it would properly update the `controller` property.
However, the card would not be removed from the list of cards controlled
by the opponent nor added back to the list of cards controlled by the
player. This caused bugs where a controlled character leaving play could
not be controlled again if they re-entered play somehow.

Fixes #1223

### DIFF
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -1065,8 +1065,11 @@ class Player extends Spectator {
 
     removeCardFromPile(card) {
         if(card.controller !== this) {
-            card.controller.removeCardFromPile(card);
+            let oldController = card.controller;
+            oldController.removeCardFromPile(card);
 
+            oldController.allCards = _(oldController.allCards.reject(c => c === card));
+            this.allCards.push(card);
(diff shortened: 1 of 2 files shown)


ystros:
Can't guarantee that this won't introduce more bugs, but the existing tests pass. We probably need to rework the entire concept of control to really get this right. 🙃

PR #1263, opened 2017-07-27
Update attachment card implementations

DukeTax:
* Updates all currently implemented attachments
* Makes sure the ability can only trigger or initiate if it would change the game state
* Standardises action labels, prompt titles and chat messages.
* Changes actions to use a handler instead of method, where needed.
* Fixes canAttach() methods to match the card limitations.

### DIFF
--- a/server/game/cards/attachments/01/bodyguard.js
+++ b/server/game/cards/attachments/01/bodyguard.js
@@ -5,13 +5,25 @@ class BodyGuard extends DrawCard {
         this.interrupt({
             canCancel: true,
             when: {
-                onCharactersKilled: event => event.cards.includes(this.parent) && event.allowSave,
-                onCardsDiscarded: event => event.cards.includes(this.parent) && event.allowSave
+                onCharactersKilled: event => {
+                    if(event.cards.includes(this.parent) && event.allowSave) {
+                        this.parentCard = this.parent;
+                        return true;
+                    }
+                    return false;
+                },
+                onCardsDiscarded: event => {
+                    if(event.cards.includes(this.parent) && event.allowSave) {
+                        this.parentCard = this.parent;
+                        return true;
+                    }
+                    return false;
+                }
             },
             cost: ability.costs.sacrificeSelf(),
             handler: context => {
-                context.event.saveCard(this.parent);
-                this.game.addMessage('{0} sacrifices {1} to save {2}', this.controller, this, this.parent);
+                context.event.saveCard(this.parentCard);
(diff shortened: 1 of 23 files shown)


server/game/cards/attachments/01/bodyguard.js
```diff
@@ -1,17 +1,17 @@
...
+    setupCardAbilities(ability) {
         this.interrupt({
             canCancel: true,
             when: {
                 onCharactersKilled: event => event.cards.includes(this.parent) && event.allowSave,
                 onCardsDiscarded: event => event.cards.includes(this.parent) && event.allowSave
             },
+            cost: ability.costs.sacrificeSelf(),
             handler: context => {
                 context.event.saveCard(this.parent);
```
ystros: This will no longer work, because the `parent` is gone after the sacrifice cost executes.

server/game/cards/attachments/02/stinkingdrunk.js
```diff
@@ -1,16 +1,15 @@
...
+    setupCardAbilities(ability) {
         this.reaction({
             when: {
                 onCardStood: event => event.card === this.parent
             },
+            cost: ability.costs.sacrificeSelf(),
             handler: () => {
                 this.game.addMessage('{0} sacrifices {1} to kneel {2}', this.controller, this, this.parent);
-
                 this.parent.controller.kneelCard(this.parent);
```
ystros: `parent` no longer exists if a sacrifice cost is used.

server/game/cards/attachments/03/needle.js
```diff
@@ -8,13 +8,18 @@ class Needle extends DrawCard {
 
         this.interrupt({
             when: {
-                onSacrificed: event => event.card === this.parent
+                onSacrificed: event => {
+                    if(event.card === this.parent) {
+                        this.parentCard = this.parent;
+                        return true;
+                    }
```
ystros: Should explicitly return false when it doesn't go into the `if`. Also don't especially like this pattern of storing off the parent card in the event handler. This should be framework automated, either by the cost system or (more likely) the event system.

In this case, it's unnecessary anyway - you can get it off of `context.event.card` in the handler.

DukeTax: Not a fan of it, but at least in the case of Bodyguard, it's still a necessary evil currently right?

ystros:
I'll merge for now, but definitely need to revisit handlers setting instance variables.

PR #1264, opened 2017-07-27
Reword timed window description on profile page

DukeTax:


cryogen:
Don't ask me why, but my preference is not to put full stops at the end of sentences for web stuff, but as long as it's consistent its ok with me :)

PR #1272, opened 2017-07-29
Rearrange canPutIntoPlay check

DukeTax:
* Checking whether a card by title is allowed to be put into play/marshalled, should happen before the unique check.
* Makes sure you can't marshal or put into play a non-unique character after returning it with "The Iron Bank Will Have Its Due".

### DIFF
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -457,12 +457,12 @@ class Player extends Spectator {
     canPutIntoPlay(card) {
         let owner = card.owner;
 
-        if(!card.isUnique()) {
-            return true;
-        }
-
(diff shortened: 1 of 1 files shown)


cryogen:
I’m sure this is correct, but what issue is it addressing?

DukeTax:
The issue where you would return any non-unique with "The Iron Bank Will Have Its Due", and then would be allowed to marshal or put it into play again.

> Action: If it is not the taxation phase, kneel your faction card and return a character to your hand to gain gold equal to that character's printed cost. Until the end of the round, you cannot marshal or put into play any card with the same title as that character.

cryogen:
Cool, thanks!

PR #1274, opened 2017-07-29
Prevent Ours is the Fury double adding characters to challenge

### LINKED ISSUE #1273: Ours is the Fury bug
Just noticed an issue with Ours is the Fury. My opponent used it to target a Stannis Baratheon who was already participating in the challenge as a defender - this is a legal play because he could potentially stand upon winning the challenge - but Throneteki counted his STR twice when we resolved the challenge because it thought there were two instances of him participating in the challenge.

DukeTax:
* Previously, when Ours is the Fury was played on a character that was already participating in the challenge, that character's strength would be added again. This PR fixes that by adding a check to challenge.addDefender.
* Fixes #1273.

### DIFF
--- a/server/game/cards/events/01/oursisthefury.js
+++ b/server/game/cards/events/01/oursisthefury.js
@@ -6,15 +6,14 @@ class OursIsTheFury extends DrawCard {
             title: 'Add knelt Baratheon as defender',
             condition: () => this.game.currentChallenge && this.game.currentChallenge.defendingPlayer === this.controller,
             target: {
-                activePromptTitle: 'Select character',
-                cardCondition: card => card.kneeled && card.controller === this.controller && card.isFaction('baratheon') && card.canParticipateInChallenge()
+                activePromptTitle: 'Select a character',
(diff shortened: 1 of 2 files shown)


cryogen:
I think this would be better fixed in the card itself, check if it's already participating and don't call addDefender()

server/game/cards/events/01/oursisthefury.js
```diff
@@ -6,15 +6,18 @@ class OursIsTheFury extends DrawCard {
...
-                cardCondition: card => card.kneeled && card.controller === this.controller && card.isFaction('baratheon') && card.canParticipateInChallenge()
+                activePromptTitle: 'Select a character',
+                cardCondition: card => card.location === 'play area' && card.kneeled && card.controller === this.controller &&
+                                       card.isFaction('baratheon') && card.canParticipateInChallenge()
             },
             handler: context => {
                 this.selectedCard = context.target;
 
-                this.game.currentChallenge.addDefender(context.target);
+                if(!this.game.currentChallenge.defenders.includes(this.selectCard)) {
```
cryogen: this.game.currentChallenge.isDefending(this.selectedCard)

PR #1276, opened 2017-07-31
Update location card implementations

DukeTax:
* Updates all currently implemented locations.
* Makes sure the ability can only trigger or initiate if it would change the game state.
* Standardises action labels, prompt titles and chat messages.
* Changes actions to use a handler instead of method, where needed.


server/game/cards/locations/01/castleblack.js
```diff
@@ -3,37 +3,26 @@ const DrawCard = require('../../../drawcard.js');
...
-            title: 'Kneel this card to stand and give a defending character +2 STR',
-            cost: ability.costs.kneelSelf(),
+            title: 'Stand and give defending character +2 STR',
             condition: () => this.game.currentChallenge,
-            method: 'kneel'
-        });
-    }
+            cost: ability.costs.kneelSelf(),
+            target: {
+                activePromptTitle: 'Select a character',
```
cryogen: Really bugs me that this isn't the default, remind me to change that sometime

cryogen:
There's some conflicts here that need resolving

DukeTax:
Github could use a feature alerting PR owners of when this happens.

cryogen:
Or I could just be more helpful and point it out sooner :)

PR #1278, opened 2017-07-31
Default card select prompt title

DukeTax:
* Defaults the title to 'Select a character' or 'Select x characters' when x > 1.


cryogen:
We need to be careful that there's no instances of this prompt being used which expects the old default (I didn't realise there was a default)

DukeTax:
Considering the old default prompt title wasn't even working correctly (see the change from numCard to numCards), it's probably safe to say it wasn't being used.

PR #1289, opened 2017-08-02
Remove unused onBeginChallengePhase method

ystros:
Fixes THRONETEKI-TD


cryogen:
RIP super old mega weird code

Issue #1217, opened 2017-07-22
Reserve discard should be a card select prompt


ystros:
During the taxation phase, instead of dragging cards to discard to get down to reserve, the player should be prompted which cards they want to discard via a card select prompt.

This will:
* Improve playability in browsers where drag-drop is finicky (Firefox, iOS browsers)
* Ensure abilities that react to 1 or more cards being discarded only fire once instead of once per card (LoCR Cersei)
* Be more consistent with the rest of the site

PR #1290, opened 2017-08-04
Convert reserve discard into a prompt

ystros:
Previously, the prompt that checked that players were below reserve
expected them to manually drag cards from hand into their discard pile.
This was inconsistent with the rest of the site, error prone (easy to
drag onto draw pile instead), and made things difficult for browsers
with finicky drag support (Firefox, touch-based dragging).

Now, when the player is above their reserve value, they will be prompt
to select the cards they wish to discard. This prompt maintains the
order of the cards selected so that the player can control the order the
cards end up in discard.

screen shot 2017-08-04 at 12 07 27 pm
screen shot 2017-08-04 at 12 07 41 pm

Completes #1217.


mplain:
Personally, I think that "bottom first" would be more intuitive (for both this and Valar etc.) because it would resemble real life experience better - the first card you decide to discard/kill ends up at the bottom, then others go on top of it.

cryogen:
I think it was me who asked for it to be this way and I must admit I've found it confusing a couple of times.  However now we've done it that way we need to keep it that way or everyone who is used to it now will be super confused

mplain:
I don't think people use that feature much. Ghosts of Harrenhal isn't all that widely played. The outcry won't be that loud. Might be easier to do it now than later.

PR #1292, opened 2017-08-05
Implement Valyrian draft only agendas

DukeTax:
* Deckbuilder validation for these agendas has already been implemented by ystros.

### DIFF
--- a/server/game/cards/agendas/thepowerofwealth.js
+++ b/server/game/cards/agendas/thepowerofwealth.js
@@ -27,11 +27,10 @@ class ThePowerOfWealth extends AgendaCard {
         });
 
         let factionToAnnounce = _.filter(factionsInDecks, faction => faction !== this.controller.getFaction() && faction !== 'neutral');
+        let message = '{0} names {1} as their {2} for {3}';
 
         if(factionToAnnounce.length > 1) {
-            this.game.addMessage('{0} has included more than the maximum allowed number of factions for {1} in their draw and/or plot deck',
-                this.controller, this);
-            return;
+            message += ' (this exceeds the maximum allowed number of factions)';
         }
 
         if(_.isEmpty(factionToAnnounce)) {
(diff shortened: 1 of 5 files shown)


server/game/cards/agendas/thepowerofwealth.js
```diff
@@ -0,0 +1,48 @@
...
+        this.controller.allCards.each(card => {
+            if(!factionsInDecks.includes(card.getPrintedFaction())) {
+                factionsInDecks.push(card.getPrintedFaction());
+            }
+        });
+
+        let factionToAnnounce = _.filter(factionsInDecks, faction => faction !== this.controller.getFaction() && faction !== 'neutral');
+
+        if(factionToAnnounce.length > 1) {
+            this.game.addMessage('{0} has included more than the maximum allowed number of factions for {1} in their draw and/or plot deck',
```
ystros: While I think such an error message is fine, it's probably better if you also announce which factions they did include. The deck the player is playing with will already be marked as invalid from the validation rules of Power for Wealth.

server/game/cards/agendas/treaty.js
```diff
@@ -0,0 +1,65 @@
...
+        this.registerEvents(['onDecksPrepared']);
+    }
+
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Draw card then discard card',
+            cost: ability.costs.kneelFactionCard(),
+            handler: () => {
+                this.controller.drawCardsToHand(1);
+                this.game.addMessage('0} uses {1} and kneels their faction card to draw 1 card', this.controller, this);
```
ystros: Typo - `0}` should be `{0}`.

server/game/cards/agendas/treaty.js
```diff
@@ -0,0 +1,65 @@
...
+
+        if(factionsToAnnounce.length > 2) {
+            this.game.addMessage('{0} has included more than the maximum allowed number of factions for {1} in their draw and/or plot deck',
+                this.controller, this);
+            return;
+        }
+
+        if(factionsToAnnounce.length < 2) {
+            //This is slightly more problematic than The Power of Wealth. Printing a message here doesn't allow bluffing a second faction,
+            //but not printing a message allows possibility to lie.
```
ystros: I'm iffy on not announcing the one faction they included here, but understand what you're going for. I'd say it's better overall to announce what they do have in the deck. If we really care about bluffing like this it should probably be built into the deck builder itself.

PR #1293, opened 2017-08-05
Implement Valyrian draft only plots

DukeTax:
* Needs player.getFaction() shorthand from #1292.

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -305,13 +305,7 @@ class BaseCard {
     }
 
     getFactions() {
-        let factions = [];
-
-        _.each(ValidFactions, faction => {
-            if(this.isFaction(faction)) {
-                factions.push(faction);
-            }
-        });
+        let factions = _.filter(ValidFactions, faction => this.isFaction(faction));
 
         if(_.isEmpty(factions)) {
             factions.push('neutral');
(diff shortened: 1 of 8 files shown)


server/game/basecard.js
```diff
@@ -292,6 +304,22 @@ class BaseCard {
         return !!this.factions[normalizedFaction];
     }
 
+    getFactions() {
+        let factions = [];
+
+        _.each(ValidFactions, faction => {
+            if(this.isFaction(faction)) {
+                factions.push(faction);
```
ystros: Nitpicky, but this would be more succinctly expressed as:
```javascript
let factions = _.filter(ValidFactions, faction => this.isFaction(faction));
```

server/game/cards/plots/00/commoncause.js
```diff
@@ -0,0 +1,32 @@
...
+
+    getNumberOfFactions() {
+        let charactersInPlay = this.controller.filterCardsInPlay(card => card.getType() === 'character');
+        let factionsInPlay = [];
+
+        _.each(charactersInPlay, card => {
+            let factions = card.getFactions();
+            _.each(factions, faction => {
+                if(!factionsInPlay.includes(faction) && faction !== 'neutral') {
+                    factionsInPlay.push(faction);
```
ystros: This whole things makes my teeth chatter but it'll work. 😬  The more functional way of expressing this isn't a lot better in this case.

You could potentially get rid of the neutral check by filtering for non-neutral characters in play.

PR #1294, opened 2017-08-05
Implement Valyrian draft only draw deck neutrals

DukeTax:
* Needs #1292 for player.getFaction() shorthand.
* Needs #1293 for card.getFactions().

### DIFF
--- a/server/game/cards/events/00/muster.js
+++ b/server/game/cards/events/00/muster.js
@@ -1,0 +1,34 @@
+const DrawCard = require('../../../drawcard.js');
+
+class Muster extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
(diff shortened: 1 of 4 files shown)


server/game/cards/events/00/muster.js
```diff
@@ -0,0 +1,34 @@
+const DrawCard = require('../../../drawcard.js');
+
+class Muster extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Search deck for Knight',
+            cost: ability.costs.kneel(card => card.hasTrait('Knight') && card.getType() === 'character'),
+            handler: context => {
```
ystros: This action should be restricted to marshal phase.

PR #1296, opened 2017-08-06
Convert A Game of Thrones to use an effect

DukeTax:


server/game/plotcard.js
```diff
@@ -62,7 +62,15 @@ class PlotCard extends BaseCard {
         return baseClaim + this.claimModifier;
     }
 
-    canChallenge() {
+    canChallenge(player, challengeType) {
+        if(!player.needsToWinIntrigueFirst) {
+            return true;
+        }
+
+        if((challengeType === 'power' || challengeType === 'military') && player.getNumberOfChallengesWon('intrigue') <= 0) {
```
ystros: Since this is just moving the implementation from the A Game of Thrones card into the base class, I don't really see this as an improvement.

There's already a `canInitiateChallenge` method on Player, so having the oddly named `canChallenge` method on PlotCard is redundant. It and its usages should be removed.

It should be possible to implement this using the existing `cannotInitiateChallengeType` effect with a condition without introducing a new effect type. Something like:
```javascript
this.persistentEffect({
    targetType: 'player',
    targetController: 'any',
    match: player => player.getNumberOfChallengesWon('intrigue') < 1,
    effect: [
        ability.effects.cannotInitiateChallengeType('military'),
        ability.effects.cannotInitiateChallengeType('power')
    ]
});
```

(don't take my word for it, be sure to test locally).

PR #1297, opened 2017-08-06
Cleanup PlayerRow component

ystros:
* Removes a bunch of dead code
* Extracts the hand rendering / functionality into its own component, `PlayerHand`.

I considered extracting the draw deck behavior into its own component as well, but it was pretty light after deleting the unused `getDrawDeck` method. Let me know if you think extracting it also would be worthwhile.

### DIFF
--- a/client/GameComponents/PlayerHand.jsx
+++ b/client/GameComponents/PlayerHand.jsx
@@ -1,0 +1,102 @@
+import React from 'react';
+import _ from 'underscore';
+import $ from 'jquery';
+
+import Card from './Card.jsx';
+import {tryParseJSON} from '../util.js';
+
+class PlayerHand extends React.Component {
+    onDragOver(event) {
+        $(event.target).addClass('highlight-panel');
+        event.preventDefault();
+    }
+
+    onDragLeave(event) {
+        $(event.target).removeClass('highlight-panel');
+    }
+
+    onDragDrop(event, target) {
+        event.stopPropagation();
+        event.preventDefault();
+
+        $(event.target).removeClass('highlight-panel');
+
+        let card = event.dataTransfer.getData('Text');
+
+        if(!card) {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


client/GameComponents/PlayerHand.jsx
```diff
@@ -0,0 +1,102 @@
...
+                onDragDrop={ this.props.onDragDrop } />);
+        });
+
+        return hand;
+    }
+
+
+    render() {
+        let className = 'panel hand';
+        let needsSquish = this.props.cards && this.props.cards.length * 64 > 342;
```
cryogen: I assume these are the constants you were talking about the other day?  These can be determined from the css properties at runtime if you stick a 'ref' on a card object and one on the hand panel itself

client/GameComponents/PlayerRow.jsx
```diff
@@ -70,68 +35,6 @@ class PlayerRow extends React.Component {
...
-                onMouseOver={ this.props.onMouseOver }
-                onMouseOut={ this.props.onMouseOut }
-                onClick={ this.props.onCardClick }
-                onDragDrop={ this.props.onDragDrop } />);
-        });
-
-        return hand;
-    }
-
-    getDrawDeck() {
```
cryogen: :+1: this was supposed to go away when i converted it to a CardCollection (which now I think of it is a really dumb name)

client/GameComponents/PlayerRow.jsx
```diff
@@ -70,68 +35,6 @@ class PlayerRow extends React.Component {
         }
     }
 
-    onDiscardedCardClick(event, cardId) {
```
cryogen: Wow this must be super old

cryogen:
I think draw deck is Ok as a CardCollection (which I'm going to rename soon to CardPile, because that makes way more sense)

PR #1301, opened 2017-08-07
Log which character attachment was marshalled on

### LINKED ISSUE #1300: Log should say which character got the attachment
says "RTT marshals Milk of the Poppy costing 1", but doesn't say on which character.

DukeTax:
* Fixes #1300.


DukeTax:
I'm actually not sure if this is the way to go. As an example, Littlefinger's reaction would now trigger, and display a message, before the message announcing his marshalling would display.

ystros:
This is trickier than it should be.

I'd say go with this basic solution, but keep the existing messages the way they are, and only do the attachment message as queued. There may still be some weird cases but it'll minimize the weirdness.

Other solutions: reworking how attachments are played, probably through a separate card action, so that you have the target parent by the time you get into executeHandler. That would be the most correct thing to do, but it will be tricky. 

Or you could move the message to sometime after the parent is chosen (e.g. in AttachmentPrompt), but you probably don't have enough information to say the cost without passing it around. Ugly.

PR #1310, opened 2017-08-10
Allow custom cards in decks

ystros:
Adds the ability to include custom cards in a deck for cases where the latest data hasn't been pulled from ThronesDB, isn't available yet on ThronesDB, or is a non-standard card.

Custom cards can now be added to a deck by manually adding a line to the deck in the format "Custom \<Type\> - \<Name\>", where type is the card type - Character, Event, Location, Attachment, or Plot. For example:

2 Custom Character - Ser Pounce

will add 2 copies of a character named Ser Pounce to the deck.

Such custom cards will be blank, neutral, traitless, and have 0 cost and 0 strength. The stats should be manipulated via in game commands to match the intended card.

Including an asterisk (*) in the card name will mark the card as unique and therefore have dupes work properly.

 Each card type uses the same placeholder image, but hovering over the card will reveal its name.


client/DeckEditor.jsx
```diff
@@ -256,7 +256,7 @@ class InnerDeckEditor extends React.Component {
         let cardName = line.substr(index, packOffset === -1 ? line.length : packOffset - index - 1);
         let packName = line.substr(packOffset + 1, line.length - packOffset - 2);
 
-        if(cardName.startsWith('Custom ')) {
```
cryogen: this is unnecessary now and the reason i imported half of the babel polyfills.  startsWith is way nicer

client/DeckEditor.jsx
```diff
@@ -257,6 +251,63 @@ class InnerDeckEditor extends React.Component {
         this.props.updateDeck(deck);
     }
 
+    lookupCard(line, index) {
+        let packOffset = line.indexOf('(');
+        let cardName = line.substr(index, packOffset === -1 ? line.length : packOffset - index - 1);
+        let packName = line.substr(packOffset + 1, line.length - packOffset - 2);
+
+        if(cardName.indexOf('Custom ') === 0) {
```
cryogen: startsWith is ok now

Issue #1231, opened 2017-07-24
Dupes should save during the 'can cancel' interrupt window


ystros:
Currently, dupes will automatically save a character when they would be killed / discarded / etc. However, this is done after all interrupts have fired to the character being killed. This can lead to bugs like Quentyn Martell's ability triggering while he has a dupe, then the dupe saves him from actually being killed.

Things aren't necessarily clear cut, though. Benjen Stark has a similar interrupt ability to being killed that the player might want to trigger even if they have a dupe on Benjen.

Issue #1024, opened 2017-06-07
Renly Baratheon (FFH) should not be savable when there's another King in town


kitoxme:
Treachery needs an unique lannister character to be played, currently works without one.

King renly should die in valar turn, when mi opponent controls another king (so cannot be saved), currently you are able to save him

thanks

ystros:
#1025 will fix the Treachery bug.

King Renly's 'cannot be saved' is a bit more complex.

PR #1313, opened 2017-08-11
Implement remaining cards from Oberyn's Revenge

DukeTax:


server/game/player.js
```diff
@@ -896,6 +896,19 @@ class Player extends Spectator {
         });
     }
 
+    moveCardToBottomOfDeck(card, allowSave = true) {
```
cryogen: I suspect @ystros might suggest that this work similarly to the above return to hand method and to factor out some of this common code

ystros: It probably should be wrapped in a `allowGameAction` for immunity purposes, like return to hand.

I would say hold off on factoring out a common method, though. Dupes + saves in general need to be reworked for issues like #1231 and #1024, which should probably be its own PR.

PR #1315, opened 2017-08-11
Prompt bestow when attachments enter play

### LINKED ISSUE #1314: Fever Dreams Bestow
Fever Dreams doesn't prompt the Bestow when you attach it. After using /bestow command, the card prompts the reaction as normal correctly.

DukeTax:
* Fixes #1314.


ystros:
Timing may be slightly weird here (Bestow prompt, then enter play event for attachments vs enter play event, then bestow prompt for everything else) but it's probably fine.

PR #1316, opened 2017-08-11
Unify all of the user logic so that defaults are in one place 

cryogen:
and we can guarantee they are set so don't have to check for them everywhere

### DIFF
--- a/client/Profile.jsx
+++ b/client/Profile.jsx
@@ -18,31 +18,16 @@ class InnerProfile extends React.Component {
             return;
         }
 
-        this.windowDefaults = {
-            plot: false,
-            draw: false,
-            challengeBegin: false,
-            attackersDeclared: true,
-            defendersDeclared: true,
-            dominance: false,
-            standing: false
-        };
-
-        this.timerSettingsDefaults = {
-            events: true,
-            abilities: false
-        };
-
         this.state = {
-            disableGravatar: this.props.user.settings ? this.props.user.settings.disableGravatar : false,
-            email: this.props.user ? this.props.user.email : '',
+            disableGravatar: this.props.user.settings.disableGravatar,
+            email: this.props.user.email,
             loading: false,
             newPassword: '',
             newPasswordAgain: '',
-            promptedActionWindows: this.props.user ? this.props.user.promptedActionWindows || this.windowDefaults : this.windowDefaults,
+            promptedActionWindows: this.props.user.promptedActionWindows,
             validation: {},
-            windowTimer: this.props.user.settings ? _.isUndefined(this.props.user.settings.windowTimer) ? 10 : this.props.user.settings.windowTimer : 10,
-            timerSettings: this.props.user.settings ? this.props.user.settings.timerSettings || this.timerSettingsDefaults : this.timerSettingsDefaults
+            windowTimer: this.props.user.settings.windowTimer,
+            timerSettings: this.props.user.settings.timerSettings
         };
 
         this.windows = [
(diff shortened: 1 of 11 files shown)


server/server.js
```diff
@@ -97,14 +88,16 @@ class Server {
...
-            app.get('*', function response(req, res) {
+            app.get('*', (req, res) => {
                 var token = undefined;
 
                 if(req.user) {
                     token = jwt.sign(req.user, config.secret);
                 }
 
-                var html = pug.renderFile('views/index.pug', { basedir: path.join(__dirname, '..', 'views'), user: req.user, token: token });
+                let user = Util.getUserWithDefaultsSet(req.user);
```
ystros: This doesn't seem like the "correct" place to be doing the defaulting and I'm kind of iffy on having the functionality live on a generic `Util` namespace. But it does handle the case of already authed users not getting new defaults without logging out/in, so 👌

server/util.js
```diff
@@ -25,7 +35,40 @@ function httpRequest(url) {
...
+function getUserWithDefaultsSet(user) {
+    let userToReturn = user;
+
+    if(!userToReturn) {
+        return userToReturn;
+    }
+
+    if(!userToReturn.settings) {
+        userToReturn.settings = {
+            disableGravatar: false,
```
ystros: This approach isn't granular enough. For example, if you add a new default under `settings` but the user already has the settings object, the new value won't be there. Instead of checking that they're missing a particular value, you could use `_.extend` to setup the defaults and copy over them with the user's existing values. E.g.
```javascript
// The if goes away, since _.extend(anything, undefined) === anything
userToReturn.settings = _.extend({}, { disableGravatar: false, windowTimer: 10 }, userToReturn.settings);
```

Unfortunately, `extend` isn't recursive so you can't provide all the defaults at once due to the way our settings are nested. Could probably write a recursive version without huge trouble, but not sure it's worth it for only having 4 keys.

server/lobby.js
```diff
@@ -428,7 +429,7 @@ class Lobby {
     }
 
     onLobbyChat(socket, message) {
-        var chatMessage = { user: { username: socket.user.username, emailHash: socket.user.emailHash, noAvatar: socket.user.settings ? socket.user.settings.disableGravatar : false }, message: message, time: new Date() };
+        var chatMessage = { user: { username: socket.user.username, emailHash: socket.user.emailHash, noAvatar: socket.user.settings }, message: message, time: new Date() };
```
ystros: `noAvatar` should be the disableGravatar value, not the settings hash.

server/lobby.js
```diff
@@ -227,7 +228,7 @@ class Lobby {
         this.sockets[ioSocket.id] = socket;
 
         if(socket.user) {
-            this.users[socket.user.username] = socket.user;
+            this.users[socket.user.username] = Util.getUserWithDefaultsSet(socket.user);
```
ystros: Is there a more central place that this can be done, rather than sprinkle it here and in `Server`? Seems like having it defaulted properly once already on the socket object would be less error prone.

client/Profile.jsx
```diff
@@ -18,31 +18,16 @@ class InnerProfile extends React.Component {
...
-        this.timerSettingsDefaults = {
-            events: true,
-            abilities: false
-        };
-
         this.state = {
-            disableGravatar: this.props.user.settings ? this.props.user.settings.disableGravatar : false,
-            email: this.props.user ? this.props.user.email : '',
+            disableGravatar: this.props.user.settings.disableGravatar,
+            email: this.props.user.email,
```
ystros: Lost the empty string default for email - will it matter anywhere?

server/server.js
cryogen: It might not even be necessary here, now that I've put it in deserialize, but hey you can't be too sure

server/lobby.js
cryogen: They're different user objects, one is the one we trust from the client, one is the one we get from the database

client/Profile.jsx
cryogen: Nope, in fact it was already redundant, we bail further up if this.props.user is null

PR #1330, opened 2017-08-14
Implement Barring the Gates

ystros:
* Reworks Marching Orders and Iron Bank Will Have Its Due to use the same mechanism to restrict playable cards.
* Implements Barring the Gates

### DIFF
--- a/server/game/cards/plots/06/barringthegates.js
+++ b/server/game/cards/plots/06/barringthegates.js
@@ -1,0 +1,14 @@
+const PlotCard = require('../../../plotcard.js');
+
+class BarringTheGates extends PlotCard {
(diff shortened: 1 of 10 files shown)


server/game/cards/plots/06/barringthegates.js
```diff
@@ -0,0 +1,14 @@
+const PlotCard = require('../../../plotcard.js');
+
+class BarringTheGates extends PlotCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
```
DukeTax: targetController defaults to 'current', while Barring the Gates impacts both players.

ystros: Whoops.

PR #1333, opened 2017-08-15
Record Roose Bolton's strength before sacrificing

DukeTax:
Previously, because costs are paid before resolution, any strength buffs on Roose had worn off by the time targets were chosen. This PR fixes that by recording his strength at the point of initiation of his ability.


DukeTax:
Ugly, but still the only way currently without modifying the engine right?

cryogen:
If this is the way to go can we have a better variable name? str seems likely to conflict with something at some point

PR #1334, opened 2017-08-15
Update Core Set event card implementations

DukeTax:
* Standardises messages.
* Revert previous destructuring of event object.
* Removes redundant activePromptitle properties.

### DIFF
--- a/server/game/cards/events/01/tearsoflys.js
+++ b/server/game/cards/events/01/tearsoflys.js
@@ -5,31 +5,21 @@ class TearsOfLys extends DrawCard {
         this.reaction({
             max: ability.limit.perChallenge(1),
             when: {
-                afterChallenge: ({challenge}) => (
-                    challenge.attackingPlayer === this.controller &&
-                    challenge.winner === this.controller &&
-                    challenge.challengeType === 'intrigue'
-                )
+                afterChallenge: event => event.challenge.attackingPlayer === this.controller && event.challenge.winner === this.controller &&
+                                         event.challenge.challengeType === 'intrigue'
             },
-            handler: () => {
-                // Use an explicit prompt instead of the target API because the
-                // card uses the word 'place' instead of 'choose'.
-                this.game.promptForSelect(this.controller, {
-                    activePromptTitle: 'Select a character',
-                    source: this,
-                    cardCondition: card => card.location === 'play area' && card.controller !== this.controller && card.getType() === 'character' && !card.hasIcon('intrigue'),
-                    onSelect: (p, card) => this.cardSelected(card)
-                });
+            //TODO Uses target API but does not technically target (contain the word 'choose')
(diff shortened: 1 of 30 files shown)


server/game/cards/events/01/tearsoflys.js
```diff
@@ -5,32 +5,22 @@ class TearsOfLys extends DrawCard {
...
             when: {
-                afterChallenge: ({challenge}) => (
-                    challenge.attackingPlayer === this.controller &&
-                    challenge.winner === this.controller &&
-                    challenge.challengeType === 'intrigue'
-                )
+                afterChallenge: event => event.challenge.attackingPlayer === this.controller && event.challenge.winner === this.controller &&
+                                         event.challenge.challengeType === 'intrigue'
             },
-            handler: () => {
```
cryogen: Seems like this choice was made for a reason, don't think we should be undoing that

server/game/cards/events/01/unbowedunbentunbroken.js
```diff
@@ -41,13 +38,14 @@ class UnbowedUnbentUnbroken extends DrawCard {
...
 
-        this.game.addMessage('{0} uses {1} to make {2} unable to initiate {3} challenges until the end of the phase', player, this, otherPlayer, challengeType);
+        this.game.addMessage('{0} plays {1} to make {2} unable to initiate {3} challenges until the end of the phase', player, this, otherPlayer, challengeType);
 
         return true;
     }
 
     cancel(player) {
-        this.game.addMessage('{0} cancel the resolution of {1}', player, this);
+        //I really want to get rid of stuff like this
```
cryogen: Things should always be cancellable and if they are then a message should be logged so people know if people are abusing it to cheat.  Also the code isn't the place to make these comments :)

server/game/cards/events/01/tearsoflys.js
DukeTax: I've discussed this with @ystros not too long ago. Using the target API, even when the card technically doesn't target has two advantages: cleaner looking card implementations and not getting prompted when the card effect does not have the chance to change the game state. A disadvantage could arise when something like High Septon gets implemented. Even so, that might be resolved by adding a flag to target, something like trueTarget: false in this case.

server/game/cards/events/01/unbowedunbentunbroken.js
DukeTax: I know your stance on this, I just have a lot of issues with it. :) It's only a handful cards in the entire cardpool that are implemented like this, and adding such functionality across the board would add extra clicks, cause problematic interactions (do you want to give new players/grievers the ability to cancel Pentoshi's When Revealed?) and be a shitload of work to boot.

Fair enough about the comment. I'll record it somewhere else. :)

PR #1336, opened 2017-08-16
Implement cost function for The Things I Do For Love

DukeTax:


server/game/costs.js
```diff
@@ -629,6 +629,26 @@ const Costs = {
                 }
             }
         };
+    },
+    /**
+     * Cost implemented specifically for The Things I Do For Love
+     * TODO: needs to be reducable for cards like Littlefinger's Meddling and Paxter Redwyne.
+     */
+    payXGoldForTTIDFL: function(getMinimumCharCost) {
```
ystros: Rather than introduce a new cost specific to Things I Do, it might be better to modify the existing payXGold cost to allow both a max and a min. Example usages:

```javascript
// Loot
cost: ability.costs.payXGold(1, () =>this.opponentDeckSize(), this.game.getOtherPlayer(this.controller))
// Ritual of R'hllor
cost: ability.costs.payXGold(1, () => this.getNumberOfStandingRhllor())
// Things I Do
cost: ability.costs.payXGold(() => this.getMinimumCharCost(), () => this.controller.gold)
```

Or, if that syntax feels too messy, something like:
```javascript
cost: ability.costs.payXGold({
    min: () => this.getMinimumCharCost(), 
    max: () => this.controller.gold
})
```

The latter is somewhat uglier but we could default min and max intelligently (min = 1 by default, max = player gold by default).

PR #1346, opened 2017-08-18
Allow blanking of Limited keyword

### LINKED ISSUE #1345: Nightmares + Limited + (Dagmer/Newly Made Lord)
Hi there,

When you use Nightmares on a Limited Location the Limited keyword is blanked. This should allow it to be targeted by effects that would normally work only on non-limited locations. e.g. Newly-Made Lord and Dagmer Cleftjaw (See comment from https://thronesdb.com/card/02111 )

I've run into a couple of issues with Nightmares not properly blanking limited cards.

1. Use Nightmares on a Limited location.
2. Marshal Newly-Made Lord and use his reaction.
Issue: Cannot select the Limited location from step 1.

The same applies to Dagmer Cleftjaw's ability.

This also looks like it might apply to Nightmares + 'Frozen Solid' and Nightmares + 'Raiding the Bay of Ice', not been able to test properly.

It looks like the !card.isLimited condition used to check targets is checking this.hasKeyword('Limited') OR this.hasPrintedKeyword('Limited'). It looks like this.hasPrintedKeyword('Limited') is an invalid condition for the above effects.

I would propose updating the !card.isLimited condition to !card.hasKeyword('Limited') in the above cases. I believe this is already the current implementation for Sea Bitch.

Thanks,
Oli

DukeTax:
* Fixes #1345.

### DIFF
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -527,10 +527,10 @@ const Costs = {
     playLimited: function() {
         return {
             canPay: function(context) {
-                return !context.source.hasPrintedKeyword('limited') || context.player.limitedPlayed < context.player.maxLimited;
+                return !context.source.isLimited() || context.player.limitedPlayed < context.player.maxLimited;
             },
             pay: function(context) {
-                if(context.source.hasPrintedKeyword('limited')) {
+                if(context.source.isLimited()) {
                     context.player.limitedPlayed += 1;
                 }
             }
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -527,10 +527,10 @@ const Costs = {
     playLimited: function() {
         return {
             canPay: function(context) {
-                return !context.source.isLimited() || context.player.limitedPlayed < context.player.maxLimited;
+                return !context.source.hasPrintedKeyword('limited') || context.player.limitedPlayed < context.player.maxLimited;
             },
             pay: function(context) {
-                if(context.source.isLimited()) {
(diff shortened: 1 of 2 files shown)


server/game/costs.js
```diff
@@ -527,10 +527,10 @@ const Costs = {
     playLimited: function() {
         return {
             canPay: function(context) {
-                return !context.source.isLimited() || context.player.limitedPlayed < context.player.maxLimited;
+                return !context.source.hasPrintedKeyword('limited') || context.player.limitedPlayed < context.player.maxLimited;
             },
             pay: function(context) {
-                if(context.source.isLimited()) {
+                if(context.source.hasPrintedKeyword('limited')) {
```
ystros: This and the check above should check whether it `isLimited()` OR `hasPrintedKeyword('limited')` in case there's ever an effect that, for instance, adds the Limited keyword to cards in hand (similar to the way certain cards add Ambush). It's unlikely to happen, but would rather not be caught flat footed.

server/game/drawcard.js
```diff
@@ -90,7 +90,7 @@ class DrawCard extends BaseCard {
     }
 
     isLimited() {
-        return this.hasKeyword('Limited') || this.hasPrintedKeyword('Limited');
+        return this.hasKeyword('limited') || !this.isBlank() && this.hasPrintedKeyword('limited');
```
cryogen: this needs some brackets!

PR #1350, opened 2017-08-20
Check buttons when updating prompt client side

ystros:
Previously, the ActivePlayerPrompt component was not checking buttons
when deciding whether to update. This lead to bugs where if two prompts
had the same titles but different buttons (or the same button names with
different arguments), the prompt would not refresh client side and the
buttons wouldn't work. The most common cases for this were multiple
Bastard Daughters being killed and prompts where multiple abilities were
available (e.g. Stark reactions to sacrifices)


cryogen:
github needs an upside down smiley

DukeTax:
🙃

PR #1357, opened 2017-08-21
Simultaneously raise kneel event during challenges

DukeTax:
Previously, the onCardKneeled event for kneeling an attacker/defender was raised before any of the other challenge initiation/declare defenders events. This made it for example impossible to prevent a Satin trigger with Winterfell. This PR fixes that by raising onCardKneeled simultaneously with the other associated challenge events.

### DIFF
--- a/server/game/gamesteps/challenge/challengeflow.js
+++ b/server/game/gamesteps/challenge/challengeflow.js
@@ -61,7 +61,17 @@ class ChallengeFlow extends BaseStep {
     }
 
     chooseAttackers(player, attackers) {
-        this.challenge.addAttackers(attackers);
+        this.attackersToKneel = [];
+        this.challenge.addAttackers(attackers, false);
+
+        _.each(attackers, card => {
+            if(!card.kneeled && !card.challengeOptions.doesNotKneelAs['attacker']) {
+                this.game.applyGameAction('kneel', card, card => {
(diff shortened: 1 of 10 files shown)


server/game/gamesteps/challenge/challengeflow.js
```diff
@@ -61,7 +61,17 @@ class ChallengeFlow extends BaseStep {
     }
 
     chooseAttackers(player, attackers) {
-        this.challenge.addAttackers(attackers);
+        this.attackersToKneel = [];
+        this.challenge.addAttackers(attackers, false);
```
ystros: It may be worth looking into whether we need the `kneel` second parameter now when adding attackers / defenders. My guess it's nearly always `false` now and the places it isn't the character may already be knelt as a cost of the ability.

server/game/gamesteps/challenge/challengeflow.js
```diff
@@ -164,9 +182,29 @@ class ChallengeFlow extends BaseStep {
             defenders = defenders.concat(this.forcedDefenders);
         }
 
-        this.challenge.addDefenders(defenders);
+        this.defendersToKneel = [];
```
ystros: No need to make this an instance variable since it's set, used and reset in the same method. A local variable would be sufficient.

PR #1358, opened 2017-08-21
Check attachments when faction / traits removed

### LINKED ISSUE #1319: Attachment should continuously check parent restrictions
Example: Randyl Tarly gets Sworn to the Watch and Dragonglass Dagger attached. Sworn to the Watch gets discarded, causing Dragonglass Dagger to be illegally attached. Because of this Dragonglass Dagger should get discarded too which currently does not happen.

ystros:
When a faction or trait is removed from the card, it is possible that
attachments on that card are no longer validly attached and must be
discarded.

Fixes #1319

### DIFF
--- a/server/game/drawcard.js
+++ b/server/game/drawcard.js
@@ -282,10 +282,24 @@ class DrawCard extends BaseCard {
         return true;
     }
 
+    removeTrait(trait) {
+        super.removeTrait(trait);
+        this.enforceAttachmentValidity();
+    }
+
+    removeFaction(faction) {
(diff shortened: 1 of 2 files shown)


ystros:
Note: This works with all released + spoiled attachment cards, but we may need to move to something more generic in the future if new cards have different types of attachment restrictions (e.g. can attach only to characters w/ 4 STR or higher only)

DukeTax:
Perhaps add a message when this happens? It could weird out some users.

PR #1363, opened 2017-08-22
Set correct cardtype for event+attach saves

DukeTax:
Previously, after playing and attaching a Risen from the Sea/Blood Magic Ritual to a character, these cards still retained their event cardtype. This made it for example impossible to discard these cards with Confiscation. This PR fixes that by setting the correct cardtype as long as these cards remain in play.

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -68,6 +68,7 @@ class BaseCard {
         this.setupCardAbilities(AbilityDsl);
 
         this.factions = {};
+        this.cardTypeSet = undefined;
         this.addFaction(cardData.faction_code);
     }
 
@@ -335,6 +336,7 @@ class BaseCard {
     }
 
     leavesPlay() {
+        this.cardTypeSet = undefined;
         this.tokens = {};
     }
 
@@ -397,7 +399,15 @@ class BaseCard {
         return this.blankCount > 0;
     }
 
+    setCardType(cardType) {
+        this.cardTypeSet = cardType;
+    }
+
     getType() {
+        return this.cardTypeSet || this.getPrintedType();
+    }
+
+    getPrintedType() {
         return this.cardData.type_code;
     }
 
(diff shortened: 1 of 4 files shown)


server/game/cards/events/01/risenfromthesea.js
```diff
@@ -14,6 +14,7 @@ class RisenFromTheSea extends DrawCard {
             handler: context => {
                 context.event.saveCard(context.target);
                 this.controller.attach(this.controller, this, context.target, 'play');
+                this.setCardType('attachment');
```
ystros: It might be better if this is part of the persistent effect below along with the 'Condition' and 'Terminal' effects. (Ditto Blood Magic Ritual). It may make handling of any attachment blanking in the future more consistent.

server/game/basecard.js
```diff
@@ -335,6 +336,7 @@ class BaseCard {
     }
 
     leavesPlay() {
+        this.cardTypeSet = undefined;
```
ystros: This won't be necessary if done with an effect instead.

server/game/cards/events/01/risenfromthesea.js
DukeTax: Blanking an attached Risen would not remove the attachment cardtype, as that's not part of the textbox at that point.

ystros: Hmm. Where would that leave the 'Condition' trait it gains? Technically it's outside the quotes that specify the text.

I guess we can worry about it if there are ever any attachment blankers released. I get the feeling an until-leaves-play lasting effect would be a better approach, but may not be worth it right now.

DukeTax: If a Nightmares-style blanking card for attachments would be printed, it would not blank the Condition trait, which I guess would lead to a bug then in this case.

On the other hand, we have seen a trait blanker as well (The White Shadows), so speculation on that front seems a bit pointless I think.

PR #1368, opened 2017-08-23
Explain Riddle vs Harvest interaction on How To Play page

DukeTax:
Also add a topic list with bookmark links.

### DIFF
--- a/client/HowToPlay.jsx
+++ b/client/HowToPlay.jsx
@@ -20,13 +20,13 @@ class HowToPlay extends React.Component {
                         <li><a href='#decks'>Adding Decks</a></li>
                         <li><a href='#profile'>Profile Options</a></li>
                         <ul className='htp-sub-list'>
-                            <li><a href='#action'>>Action Windows</a></li>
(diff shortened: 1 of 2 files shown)


cryogen:
Can I have a screenshot please?

Also typically we don’t style on ids, prefer to use classes but it’s not a huge deal

zacchiro:
mmmmmh, I've certainly missed some background/philosophy design about the (great!) "how to play" page, but this addition makes me wonder: where do we stop? we certainly don't want to replicate the FAQ in that page, nor link explicitly to any noteworthy FAQ entry, do we? wouldn't adding this entry be a classic slippery slope?

DukeTax:
Screenshot: http://i.imgur.com/HIirwsv.png

@zacchiro: This is specifically an interaction that doesn't work on entirely correct The Iron Throne, and it would require quite a bit of code to get it to work correctly, for what is a super-niche case. So it's more about explaining a (tiny) shortcoming of the platform, than teaching people the rules. :)

PR #1371, opened 2017-08-23
Validate number of different plots

ystros:
Per the RRG, you can only have a single plot by title that contains
multiple copies. All other plots can only contain a single copy by
title.


client/deck-validator.js
```diff
@@ -205,6 +205,8 @@ class DeckValidator {
             errors.push('Too few plot cards');
         } else if(plotCount > rules.requiredPlots) {
             errors.push('Too many plot cards');
+        } else if(deck.plotCards.length < rules.requiredPlots - 1) {
```
cryogen: Took me a good 5 minutes to figure out what you were doing here :)

PR #1373, opened 2017-08-24
Don't allow placing cards in setup before the game is started

### LINKED ISSUE #826: Cards can be played before keep/mulligan decision is made
Because of bad habbits I occasionally mulligan while having marshalled half way. 
That means that the cards i have left on board will not get mulliganed/shuffled/drawn. 

Its not much of an issue, but sometimes it bugs me :)

cryogen:
Also, clear the cards in play if a mulligan is taken so all cards are shuffled

Fixes #826

### DIFF
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -305,18 +305,35 @@ class Player extends Spectator {
         this.challenges.setCannotInitiateForType(type, value);
     }
 
-    initDrawDeck() {
-        this.hand.each(card => {
+    resetCardPile(pile) {
+        pile.each(card => {
             card.moveTo('draw deck');
             this.drawDeck.push(card);
         });
-        this.cardsInPlay.each(card => {
-            card.moveTo('draw deck');
-            this.drawDeck.push(card);
-        });
+    }
 
+    resetDrawDeck() {
+        this.resetCardPile(this.hand);
         this.hand = _([]);
+
+        this.resetCardPile(this.cardsInPlay);
         this.cardsInPlay = _([]);
+
+        this.resetCardPile(this.discardPile);
+        this.discardPile = _([]);
+
+        this.resetCardPile(this.deadPile);
+        this.deadPile = _([]);
+
+        _.each(this.additionalPiles, pile => {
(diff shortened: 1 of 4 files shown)


server/game/setupcardaction.js
```diff
@@ -18,6 +18,7 @@ class SetupCardAction extends BaseAbility {
 
     meetsRequirements(context) {
         return (
+            context.game.readyToStart &&
```
ystros: This variable doesn't seem to be set outside of tests??? There's a Player.readyToStart but not at the game level.

server/game/player.js
```diff
@@ -310,7 +310,14 @@ class Player extends Spectator {
             card.moveTo('draw deck');
             this.drawDeck.push(card);
         });
+        this.cardsInPlay.each(card => {
```
ystros: Probably should move things out of discard and dead pile as well, because someone somewhere is going to do that.

zacchiro: While we're at it, I'd appreciate if we can rename/generalize this method into something like resetDrawDeck(), if not even resetDecks() (plural) makingit also take care of the plot deck. I plan to use this to implement a /reset chat command, which we've discussed before, and it'd be nice to have most of its logic part of a method also used for other purposes, such as game initialization at startup.

server/game/setupcardaction.js
cryogen: Github needs more reaction emojis, my screwups are not accurately reactable in the current list

server/game/player.js
```diff
@@ -305,12 +305,36 @@ class Player extends Spectator {
...
+        this.resetCardPile(this.discardPile);
+        this.discardPile = _([]);
+
+        this.resetCardPile(this.deadPile);
+        this.deadPile = _([]);
+
+        _.each(this.additionalPiles, pile => {
+            this.resetCardPile(_(pile.cards));
+        });
+        this.additionalPiles = {};
```
ystros: I'm pretty sure this will kill the Rains schemes plot.

cryogen: In my testing it doesn't appear to

PR #1375, opened 2017-08-24
Do not allow elinor tyrell's ability to be active on setup.

### LINKED ISSUE #1174: Elinor Tyrell effect in set up.
When you play Elinor Tyrell in set up, teki allows you to play 2 limited cards in the set up.

cryogen:
Fixes #1174

### DIFF
--- a/server/game/cards/characters/06/elinortyrell.js
+++ b/server/game/cards/characters/06/elinortyrell.js
@@ -3,6 +3,7 @@ const DrawCard = require('../../../drawcard.js');
 class ElinorTyrell extends DrawCard {
     setupCardAbilities(ability) {
         this.persistentEffect({
+            condition: this.controller.phase !== 'setup',
             targetType: 'player',
             effect: ability.effects.modifyMaxLimited(1)
         });


server/game/cards/characters/06/elinortyrell.js
```diff
@@ -3,6 +3,7 @@ const DrawCard = require('../../../drawcard.js');
 class ElinorTyrell extends DrawCard {
     setupCardAbilities(ability) {
         this.persistentEffect({
+            condition: this.controller.phase !== 'setup',
```
ystros: I don't think a fix on the card is the right approach here. It only solves it for the one card instead of the general problem. I would suggest either:

1. An effect level fix - add a method to Player that returns the amount of limited cards allowed to be played but always return 1 if the current phase is setup, similar to the way getStrength works
2. A cost level fix - check the phase in the playLimited cost. Similar to above, but you wouldn't necessarily need to add a new method.
3. An engine level fix - in reality, effects should not be applied at all during setup phase. The above suggestions will only work specifically for the Limited keyword. There may be other effects now or in the future that could be inappropriately applied and influence setup.

The third option would be best but I suspect it will require rewriting the setup phase entirely, which is why I haven't done it myself already.

PR #1378, opened 2017-08-25
Add a proper window for taxation phase and remove the end round prompt

### LINKED ISSUE #1304: taxation phase action window
where is it ???

cryogen:
Fixes #1304


test/helpers/gameflowwrapper.js
```diff
@@ -73,12 +73,6 @@ class GameFlowWrapper {
         this.guardCurrentPhase('dominance');
     }
 
-    completeTaxationPhase() {
-        this.guardCurrentPhase('taxation');
-        // TODO: Discard down to reserve in case of tests that fill up the player's hand
-        this.eachPlayerInFirstPlayerOrder(player => player.clickPrompt('End Round'));
```
ystros: This method should stay for completeness (there's one for every other phase), but is should call skipActionWindow instead of clicking 'End Round'. Even if the default is now no pause, this method should be available in tests.

server/settings.js
```diff
@@ -5,7 +5,8 @@ const defaultWindows = {
     attackersDeclared: true,
     defendersDeclared: true,
     dominance: false,
-    standing: false
+    standing: false,
+    taxation: false
```
ystros: This changes the current behavior, which was always have a psuedo-window w/ the End Round prompt. Now by default it'll skip from the Challenges phase right back to the Plot phase with no pause (if they don't need to discard for reserve), which is a significant flow / tempo change. Are you sure this is what we want?

cryogen: Yes.  The current end round prompt isn't in the game rules and adds nothing that the new window doesn't give instead.  Sure there'll be some annoyed people during the transition, but when isn't there..

test/helpers/gameflowwrapper.js
cryogen: The phase has changed before it gets to the method so the assertion fails, I don't see how I can readd it without adding some artificial step or turning on the window just for tests

PR #1379, opened 2017-08-25
Add back missing complete phase methods for tests

ystros:


cryogen:
Oh, that's what you meant - should have just said :P

PR #1381, opened 2017-08-25
Implement Wardens of the South

DukeTax:


server/game/cards/locations/06/scorchingdeserts.js
```diff
@@ -3,26 +3,19 @@ const DrawCard = require('../../../drawcard.js');
 class ScorchingDeserts extends DrawCard {
     setupCardAbilities(ability) {
         this.reaction({
+            title: context => context.event.card.name,
             when: {
-                onAttackersDeclared: () => true,
-                onDefendersDeclared: () => true
+                onDeclaredAsAttacker: event => event.card.getNumberOfIcons() < 2 && event.card.controller !== this.controller,
+                onDeclaredAsDefender: event => event.card.getNumberOfIcons() < 2 && event.card.controller !== this.controller
```
ystros: I'm not 100% sure the changes to this card are necessary. Since Scorching Deserts has to be sacrificed to be triggered, it can't be used on more than one character per challenge initiation. Costs remain paid even if the ability is cancelled, so there's no risk of one instance being cancelled by Treachery but the other going through (since it'll be in discard after Treachery). So having them select the card rather than creating buttons for each card seems reasonable.

It technically isn't a targeting ability (no 'choose' in the card text), but the UI interaction is probably better allowing them to choose vs buttons.

Is there a scenario I'm missing where having separate initiation buttons matters?

DukeTax: Not really. The reason I converted was because this way it mirrors the card text more closely, it saves a click and this is a rare case where not using the target API actually results in a cleaner card implementation instead of the other way around. Downside like you said is having more UI buttons.

I'll leave the final call to you. :)

ystros: :woman_shrugging: We'll probably be revisiting UI at some point for these types of abilities, so it doesn't really matter much. I'd just avoid converting more at the moment unless there are scenarios where it matters.

PR #1384, opened 2017-08-26
remove the about-container class from about and profile pages

zacchiro:
Doing so removes the overkill vertical scroll bar from the About, How to Play,
and Profile pages.


cryogen:
This is necessary because the background does not g beyond 100% height. Please don’t merge this

PR #1385, opened 2017-08-27
Fallen from Favor: prompt only when there are characters in play

zacchiro:


cryogen:
Can this not use the target API and will that not automatically stop it prompting if there's no valid targets?

ystros:
Using the target API would do exactly as you describe - it just wouldn't allow a message to be printed. The message is in line w/ messages that @DukeTax recently added for forced interrupts / reactions, which When Revealeds are similar. But it's a language change, which is why I left it for you.

cryogen:
Oh right, the message was the other thing I meant to comment on.  I thought we were moving away from stuff like this? :confused:

zacchiro:
I've been out of touch for a while, so I might have missed relevant discussions about naming. What I care about here is not bothering users when there's nothing they could possibly do; I've copied the message almost verbatim from another card in master at the time.

If the message should be changed, just tell me what's the current best practice and I'll adapt.
If the message should be removed altogether, I can indeed use the targeting API.

DukeTax:
The messages I added to Wildfire Assault, Marched to the Wall and Political Disaster were a byproduct of adding the functionality of skipping the prompt when there is nothing to choose, so as to reduce clicks. So Zack's motive and mine are aligned here.

Using the target API with the aforementioned cards is not possible as those cards concern the opponent as well. It can be used on this card which results in a cleaner card implementation, at the cost of losing a bit of consistency in messaging.

cryogen:
I'd just remove the messages from all of them.  It can be seen from the board what is happening, I don't think messages are necessary, especially since we're removing 'pointer' messages everywhere else

zacchiro:
In conformance with the discussion here, I've now updated this PR to:
- port Fallen from Favor to the target API
- remove the message "player does not sacrifice anything"
- remove similar messages from Marched, Wildfire, and Political Disaster

(Note that I've been unable to test this locally, because my local deployment is fubar again, sigh.)

Issue #1351, opened 2017-08-20
control of characters and associated issues


Redal59:
Since building and testing with a NW/Wolf deck utilising a lot of the character stealing tech I've noticed there are more than a few bugs still to be ironed out. 

Often characters that are Ward'd or stolen with Yoren or Night's Watch Recruiter, when they are subsequently discarded or killed they can end up in the discard/dead pile of the person who stole them rather than their original owner. 

Also while trying to use Frozen Solid, it won't allow you to choose a limited location even after you blank it with Nightmares. 

Finally triggers for cards like Bastard Daughter, Robb Stark and Viserys Targaryen are all still being given to the original owner of even after characters stolen by the opponent have been killed or sacrificed.

DukeTax:
Thanks for reporting. The issue with Frozen Solid/Nightmares has been resolved and will be deployed soon. :)

ystros:
I think all of these issues have been resolved. Please reopen / open a new issue if that's not the case. Thanks!

PR #1387, opened 2017-08-28
Snapshot card state when killed or sacrificed

### LINKED ISSUE #837: attachment-induced traits, affiliation, etc. should remain long enough to trigger kill reactions

### LINKED ISSUE #1222: Robb Stark reaction when played from discard
When I played Robb Stark and Bran Stark with Night Gathers... from opponent's discard pile, there was no reaction from Robb to stand all controlling chars after Bran  was killed.

### LINKED ISSUE #1386: Robb Stark reaction when opponent controlled stark character dies
When an opponent takes control of a stark character (recruiter for the watch), and that character is killed, there is a reaction window for me to trigger robb stark. But this move is illegal as I am not controlling the killed character whilst it is on the board.

ystros:
Previously, card abilities that reacted to cards being killed could fire
inappropriately or not at all. The controller for cards under take
control would revert to their owners before the ability would fire.
Additionally, if the ability depended on traits or other effects
contributed by attachments, those traits would be lost before the
ability fired (e.g. AMaF Margaery reacting to a card being killed with a
King attachment).

Now, the onCharacterKilled event automatically records the state of the
card being killed on the `cardStateWhenKilled` property.

It is also snapshots for the onSacrificed event on the `cardStateWhenSacrificed` property.

Fixes #837 
Fixes #1222 
Fixes #1386 
Partial for #1351

### DIFF
--- a/server/game/cardsnapshot.js
+++ b/server/game/cardsnapshot.js
@@ -1,0 +1,68 @@
+const _ = require('underscore');
+
+class CardSnapshot {
+    constructor(card) {
+        this.controller = card.controller;
+        this.factions = _.clone(card.factions);
+        this.keywords = _.clone(card.keywords);
+        this.location = card.location;
+        this.parent = card.parent;
+        this.power = card.power;
+        this.strength = card.getStrength();
+        this.tokens = _.clone(card.tokens);
+        this.traits = _.clone(card.traits);
+        this.type = card.getType();
+
+        this.proxyCardMethods(card);
+    }
+
+    proxyCardMethods(card) {
+        const proxiedMethods = [
+            'getCost', 'getPrintedCost', 'getPrintedFaction', 'getPrintedType',
+            'hasPrintedKeyword', 'isLoyal', 'isUnique'
+        ];
+
+        proxiedMethods.forEach(method => {
+            this[method] = (...args) => {
+                card[method](...args);
+            };
+        });
+    }
+
+    getFactions() {
+        return _.keys(this.factions);
+    }
+
+    getStrength() {
+        return this.strength;
+    }
+
+    getType() {
+        return this.type;
+    }
+
+    hasKeyword(keyword) {
+        let keywordCount = this.keywords[keyword.toLowerCase()] || 0;
+        return keywordCount > 0;
+    }
+
+    hasToken(type) {
+        return !!this.tokens[type];
+    }
+
+    hasTrait(trait) {
+        return !!this.traits[trait.toLowerCase()];
+    }
+
+    isFaction(faction) {
+        let normalizedFaction = faction.toLowerCase();
+
+        if(normalizedFaction === 'neutral') {
+            return !!this.factions[normalizedFaction] && _.size(this.factions) === 1;
+        }
+
+        return !!this.factions[normalizedFaction];
+    }
+}
+
+module.exports = CardSnapshot;
--- a/server/game/cards/attachments/04/theboyking.js
+++ b/server/game/cards/attachments/04/theboyking.js
@@ -7,7 +7,7 @@ class TheBoyKing extends DrawCard {
         });
         this.reaction({
             when: {
-                onCharacterKilled: event => event.card.getCost() <= 3
+                onCharacterKilled: event => event.card.getPrintedCost() <= 3
             },
             cost: ability.costs.kneelSelf(),
             handler: () => {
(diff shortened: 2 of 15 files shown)


zacchiro:
Nice!

Just a nit: I can imagine a 3rd one being needed in the future: `cardStateWhenRemovedFromGame`. And, as 3 occurrences make a rule :-), how about a more general `cardStateBackup` properties, with keys pointing to the when it was backuped?

server/game/cardsnapshot.js
```diff
@@ -0,0 +1,68 @@
+const _ = require('underscore');
```
cryogen: Why not just copy the card itself? Seems you’re taking most of the core properties anyway

ystros:
@zacchiro re: naming - I originally wanted to go with something like that (probably just `cardSnapshot` but cryo was worried the naming there would be too generic.

cryogen:
I'd be ok if it was cardSnapshot but indexable as @zacchiro suggests

server/game/cardsnapshot.js
ystros: If you mean create another `DrawCard` object with the same properties - I explicitly separated it into its own class to get rid of the mutator methods. It would be best if the snapshot was a read-only object and this seemed like the best way. I'd rather code crash explicitly if you try to call, say, `modifyStrength` on a snapshot than it introduce subtle bugs by silently succeeding at modifying a card that's not real.

cryogen: The flip side is that now you have two places to maintain state and if a new card needs to react to state not provided already then it'll need adding into this.  Or if something core needs changing (admittedly unlikely) it'll need changing in both places.  I take the point about mutation, but if you make changes to an object representing historical state and not current state then you should expect it not to work as you intended :man_shrugging:

ystros: Agreed that it's not ideal that they parallel each other to the degree that they do. I think the larger problem with that is the logic duplication of checking factions, traits, etc than anything else. That could be solved by extracting those things to their own classes / static methods and using them both places.

Changes to either place at this point would require for an entirely new game element (like cost, strength, trait, etc) be introduced. Which as you say, is unlikely. Though it does look like I missed at least one thing (icons) - I was trying to get them all in one go so we don't have to worry about "a new card needs X"

On the mutation, obviously if someone does the following then they should not expect it to work:
```javascript
event.cardSnapshot.modifyStrength(1)
```

I'm more worried about subtle cases where a card gets stored off or passed around a bit and it isn't clear whether it's a card or a snapshot. If this were a statically typed language, I'd use a common interface between the two for read-only ops. As it is, the safest option seems to be crash horribly or accept that sometime somewhere someone is going to pass an object around, modify it, and it isn't obvious what went wrong.
```javascript
// This is contrived as it assumes all methods are on the same class. Could 
// easily be passed from object to object and attempt to modify the snapshot 
// a long ways away on a different class from the original call
this.doSomething(event.cardSnapshot)

doSomething(c) {
    this.doSomethingElse(c);
}

doSomethingElse(c) {
    c.modifyStrength(1);
}
```

ystros:
Happy to modify the naming but I don't quite understand the value here (`cardSnapshot.whenKilled` or `cardSnapshot.killed` vs `cardStateWhenKilled`). Each of these are event specific -
 `cardStateWhenKilled` only ever appears on the `onCharacterKilled` event, `cardStateWhenSacrificed` only on `onSacrificed`, etc. There isn't a situation where a card is simultaneously killed and sacrificed so the indexing doesn't seem to by much.

It might be different if the snapshot were stored on the card itself (e.g. `card.snapshot.whenKilled`) instead of a temporary event object, but that isn't how it's set up.

server/game/cardsnapshot.js
cryogen: I'd say that is equally unlikely given that you're mostly talking about event handlers here, which by their nature should be pretty concise.  Things that make changes to card state should really go in effects in most cases.  It almost feels like this should inherit from one of the card base objects, and if that still contains too much perhaps we need a new base object?

cryogen:
Just to be clear I'm not suggesting you change it, just removing my objection to the naming :)

PR #1392, opened 2017-08-30
Implement Oldtown, Dareon & Tris Botley

DukeTax:


server/game/cards/characters/08/trisbotley.js
```diff
@@ -0,0 +1,36 @@
...
+            },
+            handler: context => {
+                context.target.owner.moveCard(context.target, 'out of game');
+
+                this.lastingEffect(ability => ({
+                    until: {
+                        onCardLeftPlay: event => event.card === this
+                    },
+                    targetType: 'player',
+                    targetController: context.target.owner,
```
ystros: While I definitely want this to work for Melee purposes, I suspect currently this will apply the effect to both players. You could modify effect engine pretty easily to support this though - take a look at `Effect.isValidTarget`, around line 120-ish.

I think it will also work currently if you do `match: context.target.owner` instead, but not 100% sure.

PR #1394, opened 2017-08-30
Implement The House with the Red Door

DukeTax:


server/game/cards/agendas/thehousewiththereddoor.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetType: 'player',
+            effect: ability.effects.setSetupGold(4)
+        });
+
+        this.persistentEffect({
+            match: this.startLocation,
```
ystros: I don't think this actually works. At the time `setupCardAbilities` is called (when the game creates the agenda's Card object but before the deck has been fully prepared), `this.startLocation` is undefined. It's later defined after the search prompt but that doesn't update the reference here. So at the time the card is created, it's fixed to match an undefined card (I don't know what side effects that would have).

I think you'll either need to do `match: card => card === this.startLocation`, which *will* update the reference as new cards come into play, or move the call to `persistentEffect` inside the `cardSelected` method (which seems a bit weird but should work as well)

server/game/player.js
```diff
@@ -307,8 +309,10 @@ class Player extends Spectator {
 
     resetCardPile(pile) {
         pile.each(card => {
-            card.moveTo('draw deck');
-            this.drawDeck.push(card);
+            if(pile !== this.cardsInPlay || !this.cardsInPlayBeforeSetup.includes(card)) {
```
ystros: Something about this gives me a stomach ache but I'm not sure what else should be done for this weird card. 👌

PR #1396, opened 2017-08-31
Snapshot card state when killed or sacrificed

### LINKED ISSUE #837: attachment-induced traits, affiliation, etc. should remain long enough to trigger kill reactions

### LINKED ISSUE #1222: Robb Stark reaction when played from discard

### LINKED ISSUE #1386: Robb Stark reaction when opponent controlled stark character dies

ystros:
Take 2.

Fixes #837
Fixes #1222
Fixes #1386
Partial for #1351

### DIFF
--- a/server/game/cards/attachments/04/theboyking.js
+++ b/server/game/cards/attachments/04/theboyking.js
@@ -7,7 +7,7 @@ class TheBoyKing extends DrawCard {
         });
         this.reaction({
             when: {
-                onCharacterKilled: event => event.card.getCost() <= 3
+                onCharacterKilled: event => event.card.getPrintedCost() <= 3
             },
             cost: ability.costs.kneelSelf(),
             handler: () => {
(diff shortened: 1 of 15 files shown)


zacchiro:
LGTM

PR #1408, opened 2017-09-05
Remove additional pile mechanism

ystros:
Removes the ability to create additional piles and nests all agenda
specific piles at the same level as the other piles under the
`Player` object.

### DIFF
--- a/client/GameBoard.jsx
+++ b/client/GameBoard.jsx
@@ -245,7 +245,7 @@ export class InnerGameBoard extends React.Component {
             return [];
         }
 
-        let sortedCards = _.sortBy(player.cardsInPlay, card => {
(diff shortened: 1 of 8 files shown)


ystros:
This introduces breaking changes to how the client receives data and how the server sends it, so some caution needed during the deploy.

cryogen:
I'd maybe go one step further and wrap them in a cardPiles object in getState()

ystros:
Moved everything under `cardPiles` except faction, agenda and active plot (since they're all singular cards). I'd kind of like to do that with the server side piles as well, but the coupling directly is too great for the most part right now to be worth it.

cryogen:
👍

PR #1409, opened 2017-09-06
Convert remaining Mongo callbacks to promises

ystros:
* Renames existing repository classes to service classes
* Converts each service to use promises instead of callbacks
* Converts the deck API to fully use the deck service object instead of direct calls to Mongo, and to use async / await.
* Removes the dependency on mongoskin


server/api/decks.js
```diff
@@ -31,35 +31,35 @@ module.exports.init = function(server) {
             }
 
             res.send({ success: true, deck: deck });
-        });
-    });
+        } catch(err) {
+            res.send({ success: false, message: 'Error fetching deck' });
+            logger.info(err.message);
+            return next(err);
```
ystros: I'm not particularly happy with the repetition of these catch statements but couldn't find a good way to generalize it. The error logging itself probably should move to Express middleware since we seem to do it most places (also should probably log as error instead of info). But the response message is harder to pull out.

Suggestions?

cryogen: Have the service catch database errors and log them and not throw, and return null if it's a get operation.  Any other errors should get caught by the express generic error handling as you say.  Also agree with info -> error

PR #1414, opened 2017-09-07
Move cards around

cryogen:
@ystros is this what you had in mind?


ystros:
Not sure about using the full pack name as the directory - it's kind of long and has spaces. While it's not super important since cards are loaded automatically by the card loader and not required individually, I'd recommend either switching to pack codes as defined by ThronesDB (Core, WotN, LoCR, etc) or at least strip spaces from the names.

PR #1418, opened 2017-09-07
Implement cost function for Renly Baratheon's (FFH) reaction

DukeTax:
* Solves the issue where a cancel (Treachery, Begging Brothers) could be baited when there was nothing to trigger in the first place.
* Makes sure that when it does get cancelled, the right card was revealed as it is a cost (previously, the revealing of the card got cancelled as well).

### DIFF
--- a/server/game/cards/04.3 FFH/RenlyBaratheon.js
+++ b/server/game/cards/04.3 FFH/RenlyBaratheon.js
@@ -9,26 +9,15 @@ class RenlyBaratheon extends DrawCard {
             match: this,
             effect: ability.effects.cannotBeSaved()
         });
+
         this.reaction({
             when: {
-                onInsight: event => {
-                    if(event.source.controller !== this.controller) {
-                        return false;
-                    }
-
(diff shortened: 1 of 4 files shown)


DukeTax:
Solves: https://media.discordapp.net/attachments/329700681743335424/355444897467334656/1.png

ystros:
I'm fine with this, but we should be explicit that this will leak game state - Renly will now only prompt if the card drawn was loyal. So on insight draw opponents will be able to know whether a card is loyal or not by whether there's a pause. May not be a big deal but we explicitly avoided it before.

DukeTax:
That was way back when cancels weren't yet implemented. As it is now, there is going to be leakage either way. Pre this PR, if somebody triggers Renly when he didn't draw a loyal card, and his opponent Treacheries, his opponent has in the best case leaked his Treachery (if the other player fesses up and the Treachery gets taken back), or wasted his Treachery (without ever knowing there was nothing to cancel in the first place). I find preventing that to be much more important than not leaking an ability that gets triggered 99% of the time.

ystros:
👍

PR #1429, opened 2017-09-11
Implement newly spoiled cards from House of Thorns second preview article

DukeTax:


test/helpers/playerinteractionwrapper.js
```diff
@@ -88,8 +88,8 @@ class PlayerInteractionWrapper {
     }
 
     clickPrompt(text) {
-        var currentPrompt = this.player.currentPrompt();
-        var promptButton = _.find(currentPrompt.buttons, button => button.text.toLowerCase() === text.toLowerCase());
+        let currentPrompt = this.player.currentPrompt();
+        let promptButton = _.find(currentPrompt.buttons, button => button.text.toString().toLowerCase() === text.toLowerCase());
```
ystros: This change shouldn't have been necessary. What problem were you running into that required it?

DukeTax: The PayXGold prompt (as well as the bestow prompt) pass direct numbers instead of a string into the prompt button, which leads to number.toLowerCase() and an error.

```"toLowerCase" in Number.prototype; // false```

ystros: Ah! I think it would be better if the test framework change was reverted and the PayXGold cost updated to explicitly do a `toString()` on the text. While `arg` can continue to be a number, I'm betting that having the `text` argument not be a string can bite us in other ways.

ystros:
Good catch on the bestow prompt too! 👍

PR #1431, opened 2017-09-11
Add block list

cryogen:


server/lobby.js
```diff
@@ -165,9 +161,37 @@ class Lobby {
         gameSummaries = _(gameSummaries).chain().sortBy('createdAt').sortBy('started').reverse().value();
 
         if(socket) {
-            socket.send('games', gameSummaries);
+            let filteredGames = [];
+
+            if(socket.user) {
```
ystros: From a UX perspective, what's the case where we are sending to a specific socket but there isn't a user on that socket? In whatever that case is, they'll see no active games.

ystros:
From user safety perspective, we should consider whether to hide messages made by the person blocking from the person being blocked. Right now:

* Blocker cannot see / join blockee's games
* Blockee cannot see / join blocker's games
* Blocker cannot see blockee's messages
* Blockee can see blocker's messages

The game joining is the most important, so I think this PR gets that right. The messages portion right now functions like "I don't want to see this person", not "I don't want this person to see me". On social media sites blocking often functions as both, with the blocker sometimes being able to click and see blockee's messages in case they need to report it for further moderation (e.g. blockee is posting address / phone number of blocker to encourage harassment). I can see situations where the blocker is in lobby chat, the blockee sees their message and goes on a toxic tirade about them, which the blocker does not know about.

It may not matter much in this case, since we don't have things like notifications when a user mentions you, or direct messaging, and the lobby is highly visible in case we need to ban an especially abusive user, but it's something to consider.

cryogen:
I've addressed the comments but as discussed the performance hit of doing the reverse block list checking puts me off doing it for now.  It's a valid point and I do think we should do it at some point, but maybe as a 'v2'

PR #1432, opened 2017-09-12
Remove deprecated `method` property from card actions

ystros:
Converts remaining cards using the old `method` property instead of `handler` to use the handler + cost + target APIs.


server/game/cards/01-Core/MessengerRaven.js
```diff
@@ -1,21 +1,17 @@
...
+            title: 'Draw 1 card',
+            phase: 'dominance',
+            cost: ability.costs.returnSelfToHand(),
+            handler: context => {
+                context.player.drawCardsToHand(1);
 
-        this.game.addMessage('{0} returns {1} to their hand and draws 1 card', player, this);
-
-        return true;
+                this.game.addMessage('{0} returns {1} to their hand and draws 1 card', context.player, this);
```
zacchiro: minor nit, this should probably be "{0} returns {1} to their hand *to* draw 1 card"

server/game/cards/01-Core/PowerBehindTheThrone.js
```diff
@@ -10,32 +10,17 @@ class PowerBehindTheThrone extends PlotCard {
...
-        }
-
-        this.game.promptForSelect(player, {
-            activePromptTitle: 'Select a character',
-            source: this,
-            cardCondition: card => card.location === 'play area' && card.controller === player && card.kneeled,
-            onSelect: (p, card) => this.onCardSelected(p, card)
+            cost: ability.costs.discardTokenFromSelf('stand'),
+            target: {
+                cardCondition: card => card.location === 'play area' && card.controller === this.controller && card.kneeled
```
zacchiro: it looks like Power Behind the Throne can be use also to stand a character controlled by your opponent, so the controller check should be removed

server/game/cards/01-Core/VerteranBuilder.js
```diff
@@ -1,32 +1,21 @@
...
-    setupCardAbilities() {
+    setupCardAbilities(ability) {
         this.action({
-            title: 'Sacrifice this card to stand a location',
-            method: 'sacrifice'
+            title: 'Stand a location',
+            cost: ability.costs.sacrificeSelf(),
+            target: {
+                activePromptTitle: 'Select a location',
+                cardCondition: card => card.location === 'play area' && card.controller === this.controller && card.getType() === 'location'
```
zacchiro: same here: Veteran Builder can be used to stand opponents' locations, so we should remove the controller check

server/game/cards/02.2-TRtW/SyrioForel.js
```diff
@@ -4,42 +4,23 @@ class SyrioForel extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
             title: 'Give a character military icon and stealth',
-            method: 'giveIcon',
-            limit: ability.limit.perPhase(1)
+            limit: ability.limit.perPhase(1),
+            phase: 'challenge',
+            target: {
+                cardCondition: card => card.getType() === 'character' && card.controller === this.controller
```
zacchiro: Syrio can be used on opponent's character, so the controller check here should be removed

ystros:
@zacchiro Since these are all actions that must be explicitly triggered, I agree that the controller restrictions were not necessary. I've removed them + updated the message. Thanks for the review!

PR #1433, opened 2017-09-12
README: avoid cd server ; cd .. gymnastic to fetch data

zacchiro:


zacchiro:
unless I'm missing something, fetchdata.js works just fine when invoked from the root dir

(the diff on the last line of README is spurious, due to previously missing newline at the end of the line)

cryogen:
Yup, and that's exactly how I run it on production, so no idea why it is like this in here :)

PR #1437, opened 2017-09-13
Add a prominent message at the end of a round as well as a general alerting method

### LINKED ISSUE #1188: Make certain game messages / states more obvious
It's easy to miss certain important messages in the chat log:

* Your opponent leaves while you're focused on marshaling, which itself adds more messages to the game log, making the leave message less obvious or potentially even pushing it above the top of the chat log.
* You get to 15 power and win the game during a complex challenge where multiple abilities fire.

These could be fixed either by using a different styling for these type of messages in chat log, by indicating the state on the player card, or both.

cryogen:
Fixes #1188

### DIFF
--- a/server/game/gamesteps/taxationphase.js
+++ b/server/game/gamesteps/taxationphase.js
@@ -36,6 +36,24 @@ class TaxationPhase extends Phase {
 
     roundEnded() {
         this.game.raiseEvent('onRoundEnded');
+
+        let players = this.game.getPlayers();
+        let playerStr = '';
+        let firstPlayer = true;
+
+        _.each(players, player => {
+            if(firstPlayer) {
+                firstPlayer = false;
+            } else {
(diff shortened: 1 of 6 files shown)


server/game/gamesteps/taxationphase.js
```diff
@@ -36,6 +36,24 @@ class TaxationPhase extends Phase {
...
+        let firstPlayer = true;
+
+        _.each(players, player => {
+            if(firstPlayer) {
+                firstPlayer = false;
+            } else {
+                playerStr += ', ';
+            }
+
+            playerStr += `${player.name}: ${player.getTotalPower()}`;
```
ystros: Something like this would be more succinct:
```javascript
let playerStr = players.map(player => `${player.name}: ${player.getTotalPower()}`).join(', ')
```

PR #1447, opened 2017-09-16
Record card snapshots in replacement handlers

ystros:
Because card snapshots for kill / sacrifice events are done within the
normal handler, any replacement effects like Benjen Stark must
explicitly also create a snapshot. Otherwise, cards that react to other
cards being killed (Starks, Funeral Pyre, etc) will encounter an
undefined object when looking for card state.

Fixes THRONETEKI-1E5
Fixes THRONETEKI-1EF
Fixes THRONETEKI-1EA
Fixes THRONETEKI-1EK
Fixes THRONETEKI-1E0
Fixes THRONETEKI-1E7
Fixes THRONETEKI-1EE

### DIFF
--- a/server/game/cards/01-Core/BenjenStark.js
+++ b/server/game/cards/01-Core/BenjenStark.js
@@ -15,6 +15,7 @@ class BenjenStark extends DrawCard {
 
                 this.game.addPower(this.controller, 2);
                 context.replaceHandler(() => {
+                    context.event.cardStateWhenKilled = this.createSnapshot();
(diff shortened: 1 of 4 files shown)


ystros:
This may be a temporary step - I want to see if it will quiet down the errors. If it does, then it is probably safe to create the snapshot prior to interrupts triggering instead of during the handler. I'm just not sure whether the previous onSacrificed errors were related to things happening before the handler or not and don't want to hide those errors as they may indicate a deeper problem with the event / ability system.

PR #1450, opened 2017-09-16
add card links to chat

cryogen:


server/game/game.js
```diff
@@ -470,6 +471,17 @@ class Game extends EventEmitter {
             if(this.chatCommands.executeCommand(player, args[0], args)) {
                 return;
             }
+
+            let card;
+            card = _.find(this.cardData, c => {
```
ystros: Why the separation between declaration and initial assignment?

Issue #1237, opened 2017-07-25
Implementing Melee


ystros:
A checklist of things that need to happen before it's possible to implement Melee. Will add on to this list as other things come up.

* [ ] UI must be reworked to fit at least 3 players. Realistically, if this is possible then 4 player mode should be possible as well. 5+ players is likely extremely difficult to do well UI-wise.
* [ ] Title cards must be implemented and a prompt to select them added in accordance with RRG Plot Phase 1.4
* [ ] Cards that say 'choose an opponent' must prompt the player to choose an opponent instead of assume there is one other player. In two player games, the opponent should still be auto-selected without prompt as it is currently.
* [ ] Usages of `Game.getOtherPlayer` must be replaced with the proper player - the result of 'choose an opponent' as above, or the defending player for usages of 'opponent' during a challenge.
* [ ] Effects that are applied directly to player objects must be examined and ensure support for applying them directly to a specific player. Currently the `targetController: 'opponent'` option would apply the effect to all opponents.
* [ ] Effects that are applied to card objects that use the `targetController: 'opponent'` option must be examined to make sure they're intended to be applied to all opponents and not just specific ones.
* [ ] Card messages may need some reworking to make it clear which player an ability is targeting.

cryogen:
Game.getOtherPlayer should be abstracted in the same way that targets and
costs are. Cards should perform common actions which will be in one central
place and will only need changing once Esther than everywhere

On 25 Jul 2017, at 07:40, Brian Upton <notifications@github.com> wrote:

A checklist of things that need to happen before it's possible to implement
Melee. Will add on to this list as other things come up.

   - UI must be reworked to fit at least 3 players. Realistically, if this
   is possible then 4 player mode should be possible as well. 5+ players is
   likely extremely difficult to do well UI-wise.
   - Title cards must be implemented and a prompt to select them added in
   accordance with RRG Plot Phase 1.4
   - Cards that say 'choose an opponent' must prompt the player to choose
   an opponent instead of assume there is one other player. In two player
   games, the opponent should still be auto-selected without prompt as it is
   currently.
   - Usages of Game.getOtherPlayer must be replaced with the proper player
   - the result of 'choose an opponent' as above, or the defending player for
   usages of 'opponent' during a challenge.
   - Effects that are applied directly to player objects must be examined
   and ensure support for applying them directly to a specific player.
   Currently the targetController: 'opponent' option would apply the effect
   to all opponents.
[...]

PR #1451, opened 2017-09-17
Remove `getOtherPlayer` from cards to allow Melee

ystros:
* Adds `getOpponents` method to game to get opponents instead of the only other player.
* Replaces in-challenge opponent with the appropriate player within the challenge (winner, loser, defendingPlayer, attackingPlayer)
* Records the appropriate opponent when the usage is in another method after a separate prompt or in an ability Then clause.
* Updates most other usages of getOtherPlayer, including abilities that affect each opponent or need one/all opponents to meet a check.

Does not update:
* Cards that interact with an opponent faction card (Chamber of the Painted Table, Vanguard Lancer) - need to rework click handlers to allow clicking on the faction card.
* Cards that require each player to do something - A Gift of Arbor Red (already hugely complicated card, probably will be a PR on its own) and Jojen Reed.
* Loot - there is an implicit 'choose opponent' that needs to be executed prior to costs, while the current `chooseOpponent` option executes just prior to targets.
* Varys' Riddle - probably could be done by converting it into the target API but like Chamber of the Painted Table, it would be better if it were automatic in 2 player mode.

Progress toward #1237.

### DIFF
--- a/server/game/cards/01-Core/KingsHuntingParty.js
+++ b/server/game/cards/01-Core/KingsHuntingParty.js
@@ -3,25 +3,18 @@ const DrawCard = require('../../drawcard.js');
 class KingsHuntingParty extends DrawCard {
     setupCardAbilities(ability) {
         this.persistentEffect({
-            condition: this.effectCondition.bind(this),
+            condition: () => this.anyOpponentHasKing(),
             match: this,
             effect: ability.effects.addIcon('intrigue')
         });
     }
 
-    effectCondition() {
-        var otherPlayer = this.game.getOtherPlayer(this.controller);
-        if(!otherPlayer || this.controller.phase === 'setup') {
-            return false;
-        }
-
-        if(otherPlayer.anyCardsInPlay(card => {
-            return card.getType() === 'character' && card.hasTrait('King');
-        })) {
-            return true;
-        }
-
-        return false;
+    anyOpponentHasKing() {
+        return this.game.anyCardsInPlay(card => (
+            card.controller !== this.controller &&
+            card.getType() === 'character' &&
+            card.hasTrait('King')
+        ));
     }
 }
 
--- a/server/game/cards/03-WotN/YoungSpearwife.js
+++ b/server/game/cards/03-WotN/YoungSpearwife.js
@@ -10,11 +10,8 @@ class YoungSpearwife extends DrawCard {
     }
 
     hasLessFactionPower() {
-        var otherPlayer = this.game.getOtherPlayer(this.controller);
-        if(!otherPlayer) {
-            return false;
-        }
-        return this.controller.faction.power < otherPlayer.faction.power;
+        let opponents = this.game.getOpponents(this.controller);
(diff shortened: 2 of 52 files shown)


ystros:
The `Object.values` failures should be resolved by the explicit upgrade to Node 8 in #1450, though I'm surprised our polyfills don't take care of it. Not sure about the Jenkins failure though.

server/game/cards/01-Core/KingsHuntingParty.js
```diff
@@ -3,25 +3,18 @@ const DrawCard = require('../../drawcard.js');
 class KingsHuntingParty extends DrawCard {
     setupCardAbilities(ability) {
         this.persistentEffect({
-            condition: this.effectCondition.bind(this),
+            condition: () => this.anyOpponentHasKing(),
```
cryogen: It's a little weird to wrap a function call in a lamda when you could just pass the function itself, but on my list of things I care about, this is pretty low

server/game/cards/03-WotN/YoungSpearwife.js
```diff
@@ -10,11 +10,8 @@ class YoungSpearwife extends DrawCard {
     }
 
     hasLessFactionPower() {
-        var otherPlayer = this.game.getOtherPlayer(this.controller);
-        if(!otherPlayer) {
-            return false;
-        }
-        return this.controller.faction.power < otherPlayer.faction.power;
+        let opponents = this.game.getOpponents(this.controller);
+        return opponents.some(opponent => this.controller.faction.power < opponent.faction.power);
```
cryogen: who on earth decided 'some' was better than 'any'? :shakes_head_at_es6:

server/game/game.js
```diff
@@ -102,11 +102,17 @@ class Game extends EventEmitter {
     }
 
     getPlayers() {
-        return _.omit(this.playersAndSpectators, player => this.isSpectator(player));
+        return Object.values(this.playersAndSpectators).filter(player => !this.isSpectator(player));
```
cryogen: A little less succinct than _.filter() (which this should have been..) but also removes the ambiguity of arrays/objects one sometimes gets into when working with underscore, so on the whole i think i like it 👍

server/game/cards/03-WotN/YoungSpearwife.js
ystros: Still waffling about whether to introduce aliases for `any` and `all`. Trying to stick w/ standard JS for now but I agree, the standard naming is terrible.

server/game/cards/01-Core/KingsHuntingParty.js
ystros: The rules for `this` are so screwed up that I'm not 100% that's true in this case. But either way the lambda reads better to me than `this.method.bind(this)`.

cryogen: binding this means you get to use this in the function, if the function doesnt need this then it's not needed.  The need for binding this is going away in a future node version i believe

PR #1452, opened 2017-09-17
Treat faction card like a normal card

ystros:
* Removes special client and server side handling for clicking on a faction card, allowing faction card to be clicked during card select prompts.
* Removes the 'Faction Card' button during prompts for Vanguard Lancer and Long Winter in favor of allowing the faction card to be clicked directly (makes Melee possible for Lancer)
* As a bonus, /power command now works consistently for either cards in play or the faction card.

### DIFF
--- a/client/GameBoard.jsx
+++ b/client/GameBoard.jsx
@@ -189,10 +189,6 @@ export class InnerGameBoard extends React.Component {
         this.props.sendGameMessage('cardClicked', card.uuid);
     }
 
-    onFactionCardClick() {
-        this.props.sendGameMessage('factionCardClicked');
-    }
-
     onDrawClick() {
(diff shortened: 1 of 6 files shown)


cryogen:
I predict this will cause some confusion at first

zacchiro:
> As a bonus, /power command now works consistently for either cards in play or the faction card.

So, should we remove the +/- buttons to explicitly modify power?
It'd seem consistent to me, and reduce UI clutter.

cryogen:
The ui buttons are a better user experience than having to type commands

zacchiro:
Well, yes, but we're talking about situations that shouldn't happen in the first place.
My point here is that either we had the same buttons to individual cards to fiddle with power tokens on those cards, or we remove the special case for the faction card, no?

PR #1458, opened 2017-09-17
Add canPutIntoPlay check to abilities

DukeTax:
Previously, when a character that could not be put into play was chosen as a target for such an effect, even though the card would not be put into play, costs were paid and messages for said abilities were printed anyway. This PR fixes that by adding canPutIntoPlay checks as a targeting restriction or play condition on all cards that put characters into play.


ystros:
I feel like there's probably a general solution waiting to be discovered here around repeated ability types (choose and put an X into play, choose and kneel/stand a X, etc) but it'll need more thought before we do it.

PR #1466, opened 2017-09-19
sort the block list

zacchiro:
I haven't actually tried this, as i don't yet have locally deployed a recent enough node on my dev laptop, sorry about that.


cryogen:
Would prefer if this sorted server side

zacchiro:
> Would prefer if this sorted server side

Done.

(Although not tested locally due to the unrelated local setup issue I've just mentioned on Discord.)

PR #1468, opened 2017-09-19
Implement King's Blood

DukeTax:
I'd love to use the snapshot functionality here, but it's not that straightforward. Ideas?

### DIFF
--- a/server/game/cards/06.6-TBWB/KingsBlood.js
+++ b/server/game/cards/06.6-TBWB/KingsBlood.js
@@ -1,0 +1,45 @@
+const _ = require('underscore');
+
+const DrawCard = require('../../drawcard.js');
+
+class KingsBlood extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Discard power from opponent\'s faction',
+            condition: () => {
+                //Possible TODO: incorporate snapshot functionality
+                if(this.hasToken('gold')) {
+                    this.parentCard = this.parent;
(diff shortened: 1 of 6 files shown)


ystros:
My idea for these situations is to take the snapshot at the beginning of AbilityResolver and put it on `context` as `cardStateWhenInitiated`. Then you wouldn't need to record the parent and gold tokens in the event listener, just reference them directly in the handler off of `context.cardStateWhenInitiated.parent/gold`. Alternatively, the snapshot could occur only for costs that sacrifice a card, but that may be less intuitive than something that's always available.

Then we'd need to go through cards that sacrifice themselves as a cost and update accordingly.

If you want to take a crack at this, then great. If not, I'd at least recommend renaming `this.parentCard` and `this.gold` to something that look less like the normal properties of the card. I think the convention we've used elsewhere is `this.originalX`.

server/game/cards/06.6-TBWB/KingsBlood.js
```diff
@@ -0,0 +1,46 @@
...
+                }
+                return false;
+            },
+            phase: 'plot',
+            cost: [
+                ability.costs.kneelParent(),
+                ability.costs.sacrificeSelf()
+            ],
+            handler: context => {
+                _.each(this.game.getPlayers(), player => {
```
ystros: While identical in functionality, looping through `this.game.getOpponents(context.player)` would closer match the card text and thus read better (and also not need the extra if)

PR #1469, opened 2017-09-19
Implement Patchface

DukeTax:


server/game/effects.js
```diff
@@ -305,6 +305,18 @@ const Effects = {
...
         };
     },
+    addIcons: function(iconsFunc) {
+        return {
+            apply: function(card) {
+                let icons = iconsFunc();
+                _.each(icons, icon => card.addIcon(icon));
+            },
+            unapply: function(card) {
+                let icons = iconsFunc();
```
ystros: For both this and `addKeywords`, you need to store off which icons / keywords were applied, and use that list to remove them. And like other dynamic effects, it needs to be marked as state dependent. Otherwise, the effect won't be recalculated as new Fool cards come into play, and won't be properly removed as Fool cards leave play.

test/server/cards/characters/06/06107-patchface.spec.js
```diff
@@ -0,0 +1,55 @@
...
+            it('should match the keyword', function() {
+                expect(this.patchface.hasKeyword('insight')).toBe(true);
+            });
+        });
+
+        describe('when the fool trait is added to a character', function() {
+            beforeEach(function() {
+                this.player1.clickCard(this.motley);
+                this.player1.clickCard(this.renly);
+                this.player1.clickCard(this.patchface);
```
ystros: Per the comment above, you'll note that you're having Patchface enter play after Renly gets the Fool trait. A good way to test the changes you need to make would be to have Patchface in play already, then attach Motley to Renly.

PR #1472, opened 2017-09-20
Implement Maester Ballabar & Queen of the Seven Kingdoms

DukeTax:


server/game/costs.js
```diff
@@ -378,6 +378,34 @@ const Costs = {
...
+        };
+    },
+    /**
+     * Cost that will remove the parent card the current card is attached to from the challenge.
+     */
+    removeParentFromChallenge: function(challengeFunc) {
+        return {
+            canPay: function(context) {
+                let challenge = challengeFunc();
+                return !!context.source.parent && challenge.isParticipating(context.source.parent);
```
ystros: The `challengeFunc` function could potentially return a null/undefined challenge. If it does so, this will crash.

Issue #1474, opened 2017-09-21
Better immunity support vs non-targeting abilities


ystros:
Immunity should prevent an ability that would modify the card from triggering even if they don't explicitly target it.

Example:
A character with immunity to character abilities gains power. Edmure Tully should not trigger, as that character is immune to his effect. However, since that card is part of the underlying 'gain power' event and not a target, immunity isn't checked automatically.

ystros:
**Scenario:**
A player has Your King Commands It out. A character has an ability that would cause it to gain power normally. That ability should not prompt to trigger since YKCI will prevent it from gaining that power.

cryogen:
The way I see it (which is probably wrong) either every card that does simple things like call 'modifyPower' directly needs to do so via an effect, or it needs to explicitly state what gameAction it intends to perform in its handler.  I've tested the latter and it works fine, and I don't see the former working terribly differently.

Thoughts?

ystros:
The way I'd like it to ultimately work involves the pre-fab game actions discussed in #1799, which solves other problems at the same time. E.g. Instead of this:
```javascript
class TheonGreyjoy extends DrawCard {
    setupCardAbilities() {
        this.reaction({
            when: {
                afterChallenge: event => (
                    event.challenge.winner === this.controller && 
                    event.challenge.isParticipating(this) && 
                    event.challenge.isUnopposed() &&
                    this.allowGameAction('gainPower')
                )
            },
            handler: () => {
                this.modifyPower(1);
                this.game.addMessage('{0} gains 1 power on {1} after winning an unopposed challenge', this.controller, this);
            }
        });
    }
}
``` 
you would specify a game action constructor instead of a handler, something like:

```javascript
class TheonGreyjoy extends DrawCard {
    setupCardAbilities() {
        this.reaction({
            when: {
                afterChallenge: event => (
                    event.challenge.winner === this.controller && 
                    event.challenge.isParticipating(this) && 
                    event.challenge.isUnopposed()
                )
            },
            gameAction: ability.gameActions.gainPower({ card: this, amount: 1 })
        });
    }
}
```

This would allow the ability system to automatically check for immunity vs different types of game actions, as well as reduce handler repetition / inconsistency in game messages for abilities. It could also be made to check that the handler will actually modify game state. Example, for an ability that kneels a character, we explicitly check that the character is standing and that they're allowed to be knelt:

```javascript
class FilthyAccusations extends PlotCard {
    setupCardAbilities() {
[...]

PR #1475, opened 2017-09-21
Distinguish between gaining and moving power

ystros:
🙃 

Per #1474, Edmure will still trigger inappropriately for an immune character, but will no longer actually allow the player to gain power (since the immune character cannot lose the power).


server/game/gamesteps/triggeredabilitywindowtitles.js
```diff
@@ -1,6 +1,7 @@
 const EventToTitleFunc = {
     onCardAbilityInitiated: event => 'the effects of ' + event.source.name,
-    onCardPowerChanged: event => event.card.name + ' gaining power',
+    onCardPowerGained: event => event.card.name + ' gaining power',
+    onCardPowerMoved: event => 'power moved from ' + event.source.name + ' to ' + event.target.name,
```
cryogen: Bit nitpicky but we've been starting to use interpolation and it seems to fit here

PR #1484, opened 2017-09-23
add /reveal-hand chat command

zacchiro:


zacchiro:
I'm fishing for comments to the idea of this chat command, that would allow to workaround cards like Ruby of R'hllor, that i've recently seen played. An alternative would be to implement the reveal logic in that card itself, but the chat command seemed like a more general solution to me.

DukeTax:
I think this is a good idea for a card like Ruby of R'hllor, or a botched His Viper Eyes for example. You might want to phrase the chatmessage as something like "Player uses the reveal-hand command to reveal their hand as:", so players that don't read the About page (ie everyone) can see how it was accomplished.

Also, updating submodules should fix the tests.

cryogen:
This PR forcefully downgrades the submodule, which probably isn't what you want, as Duke says you'll need to update submodule

server/game/chatcommands.js
```diff
@@ -408,6 +409,10 @@ class ChatCommands {
         });
     }
 
+    revealHand(player) {
+        this.game.addMessage('{0} reveals their hand as: {1}', player, player.hand);
```
cryogen: This (and all slash commands) should use addAlert('danger, ...)

cryogen:
I'm ok with the slash command existing, but ruby of r'hllor should not be implemented like this

zacchiro:
Thank you all for your quick feedback.
I've aligned submodule with master (discovering in the process that git rebase doesn't propagate changes to submodules, sigh), adapted the chat message, and switched to a "danger" alert.

PR #1494, opened 2017-09-27
Unify select card prompt and target API logic

### LINKED ISSUE #1398: Abilities should take target immunity into account when prompting
**Scenario:**
Player 1 has a single location out that cannot be discarded (e.g. from House with the Red Door)
Player 2 wins a military challenge by 5+ STR and has Put to the Torch in hand.

**Currently:**
Player 2 gets prompted for Put to the Torch and triggers it
Player 2 gets a prompt to select a location to discard, but there are no valid targets to select.

**Desired:**
Player 2 should not get prompted at all for Put to the Torch since there are no valid targets.

ystros:
Refactors the select card prompt behaviour into distinct classes, one for each type of selection prompt we have currently:
* Select a single card
* Select up to X cards
* Select unlimited cards (rare but there's a few)
* Select unlimited cards up until they reach X max stat (e.g. Robert's Warhammer).

The target API uses this mechanism as well to ensure that both the prompts and itself detect targets the same way.

Paves the way for implementing a "Select exactly X cards" prompt as well as adding "Each player selects X" support for the target API.

EDIT:
Fixes #1398


server/game/CardSelector.js
```diff
@@ -0,0 +1,141 @@
+const defaultProperties = {
```
cryogen: Can I be annoying and ask that these go in separate files?  In a folder together if necessary

ystros: Not annoying, probably should have done it in the first place. It was convenient to have them all in one place during refactoring.

PR #1509, opened 2017-10-03
Refresh user auth when blocklist is modified

### LINKED ISSUE #1505: Keep getting blocked players in my lobbies
I'll see the already created lobby drop back to the list of games, I'll recreate a new one and then I'll see people on my block list show up in my game.

ystros:
In order to update the blocklist on both the client side JWT and server
side user objects, a new token is sent back with the updated blocklist,
similar to what is done on profile updates. Then, the client reauths
with the server using the token, which updates the user object on the
corresponding socket on the server side.

Fixes #1505 (once #1507 is merged)

### DIFF
--- a/client/BlockList.jsx
+++ b/client/BlockList.jsx
@@ -47,6 +47,7 @@ class InnerBlockList extends React.Component {
             successPanel = (
                 <AlertPanel message='Block list entry added successfully' type={ 'success' } />
             );
+            this.props.socket.emit('authenticate', this.props.token);
         }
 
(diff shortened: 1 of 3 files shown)


ystros:
This feels super messy, but the alternative seems to be not to use the callApiMiddleware being used for the block list and revert to manual calls to `$.ajax` as is done on the profile page.

cryogen:
This is broadly similar to how I’d have done it so 🤷‍♂️

PR #1518, opened 2017-10-04
Implement Faceless Man

DukeTax:


server/game/basecard.js
```diff
@@ -467,6 +471,10 @@ class BaseCard {
         this.game.raiseEvent('onCardTraitChanged', { card: this });
     }
 
+    getTraits() {
+        return _.keys(_.omit(this.traits, trait => trait < 1));
```
cryogen: Not sure what this is trying to do, but I suspect it isn't actually doing what you want it to

server/game/cards/09-HoT/FacelessMan.js
```diff
@@ -0,0 +1,34 @@
...
+                let copyObj = {
+                    icons: context.target.getIcons(),
+                    keywords: context.target.getPrintedKeywords(),
+                    factions: context.target.getFactions(),
+                    traits: context.target.getTraits()
+                };
+
+                this.untilEndOfPhase(ability => ({
+                    match: this,
+                    effect: ability.effects.gainIconsKeywordsFactionsTraits(copyObj)
```
cryogen: Although concise, this is super specific to this card.  Would it not be better to have separate effects for each 'part' gained? @ystros ?

server/game/drawcard.js
```diff
@@ -223,6 +225,10 @@ class DrawCard extends BaseCard {
         return Math.max(0, baseStrength + this.dominanceStrengthModifier);
     }
 
+    getIcons() {
+        return _.filter(Icons, icon => this.hasIcon(icon));
```
cryogen: What other icons are in the icons collection that necessitates this filter?

server/game/basecard.js
DukeTax: It removes any trait from traits that has value 0, then takes the keys from the trait object and puts it in an array. It returns ```[ 'king', 'lord' ]``` for Robert Baratheon for example. The omit is necessary for traits that have been gained (creates new key/value pair in traits) and then lost (value is set to 0, but the key/value pair remains), for example through a crown attachment.

server/game/drawcard.js
DukeTax: Not sure I understand this one...

server/game/basecard.js
cryogen: Ah yes, traits are a refcount aren't they... Should probably get fixed to be removed if it hits 0 but we can deal with that separately. I don't like how this looks but it'll probably have to do for now

server/game/drawcard.js
cryogen: Similar issue to above, weird looking code because of how icons is implemented

server/game/cards/09-HoT/FacelessMan.js
ystros: All options are bad here.
1. (Current) Introducing an effect that will likely only ever be used for one card - this one - is annoying.
2. You could add individual effects that will add an array of factions, array of traits, etc, but again, that feels somewhat one-off.
3. You could modify the `addTrait`, `addFaction`, etc effects to take either a value or an array. But doing so could make those effects less clear / implementation would have to be careful otherwise it would introduce bugs.
4. You could build up this effect using existing effects and bundling it into an array either using a bunch of concats or destructuring. Syntax gets nasty though.
```javascript
  match: this,
  effect: [
    ...context.target.getIcons().map(icon => ability.effects.addIcon(icon),
    ...context.target.getTraits().map(trait => ability.effects.addTrait(trait),
    // etc
  ]
});
```

DukeTax: I had considered options 1 to 3, but 4 is probably the most straightforward solution and it has the advantage of keeping all annoying card-specific code on the card itself. So I went with that.

Issue #1389, opened 2017-08-30
Improvements to target API + implement target interaction cards


ystros:
In order to support [cards that interact with targeting](http://thronesdb.com/find?q=x%3Atarget&sort=set&view=card), the following updates to the target API will be needed.

* [x] Directly support 'each player chooses' cards (e.g. Wildfire Assault).
* [x] Directly support 'choose X card controlled by each player' cards (e.g. Even Handed Justice)
* [x] Directly support 'select exactly X cards' prompts (e.g. Duel, >= 20 STR Frey Hospitality)
* [x] For convenience, either a flag needs to be added to indicate prompts that choose a card but the ability does not contain the word 'choose' (e.g. Tears of Lys), or it needs to be separated out to another property that acts similar to `target`.
* [x] Convert cards to consistently use the target API (examine every card that still uses `promptForSelect` - some will probably still be necessary because the choose occurs in a 'then' clause ).
* [x] Make it easier to detect that the ability only targets a single card (hard to do without the above, but could do earlier if an indeterminate number of bugs are tolerated in the meantime)
* [x] Allow targets to be updated on the context object (High Sparrow specific)

DukeTax:
Adding to this: a card like Arianne Martell is currently implemented using the target API. This works out well, because even though Arianne doesn't target, the card she puts into play has to be revealed during Step 1 of Initiating Abilities, which is now taken care of by the target API. However, such cards might need a separate flag or a similar but different property as well, so it won't conflict with cards that interact with cards that truly target.

Some related fresh rulings: http://www.cardgamedb.com/forums/index.php?/topic/37127-players-choices-vs-cancels/

PR #1522, opened 2017-10-09
Implement High Septon

ystros:
Progress toward #1389


DukeTax:
It's missing the limit on High Septon, but looks good otherwise.

ystros:
Changed `isTargeting` boolean flag to a `type` property, and added the limit to High Septon.

PR #1526, opened 2017-10-11
Implement Corpse Lake

zacchiro:


server/game/cards/09-HoT/CorpseLake.js
```diff
@@ -0,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+
+class CorpseLake extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardsDiscarded: event => event.originalLocation === 'draw deck' && event.card.getType() === 'character'
```
DukeTax: Needs to only trigger when a character is discarded from an opponent's deck.

server/game/cards/09-HoT/CorpseLake.js
```diff
@@ -0,0 +1,19 @@
...
+            },
```
DukeTax: Missing the limit.

zacchiro:
…all true, …and now all fixed :)
thanks for the review!

PR #1529, opened 2017-10-11
Display ability source location when it's triggering not from hand/play

### LINKED ISSUE #1154: Indicate location of event for Annals of Castle Black
**Scenario**
I have two copies of an event, one in hand, one in discard. I reveal Annals of Castle Black as my plot. I would prefer to play the copy that's already in discard, saving the copy in hand for later.

Currently, the prompt for interrupt / reaction based events does not give you a way to tell which one is in hand and which is in discard.

DukeTax:
schermafbeelding 2017-10-05 om 15 38 42

Fixes #1154.

### DIFF
--- a/server/game/gamesteps/triggeredabilitywindow.js
+++ b/server/game/gamesteps/triggeredabilitywindow.js
@@ -95,7 +95,7 @@ class TriggeredAbilityWindow extends BaseAbilityWindow {
             if(abilityChoice.text !== 'default') {
                 title += ' - ' + abilityChoice.text;
             }
-            if(abilityChoice.card.location !== 'hand' && abilityChoice.card.location !== 'play area') {
+            if(abilityChoice.card.location !== abilityChoice.ability.location) {
                 title += ' (' + abilityChoice.card.location + ')';
             }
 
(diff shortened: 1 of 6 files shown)


DukeTax:
This is a very simple stab at fixing ####, simply showing the source location in the button prompt for any ability that’s triggering not from hand or play. Another option would be only showing source locations whenever a duplicate ability, with differing source locations, are present in the same prompt. But that would be quite a bit more code to get right, for possibly not much gain.

A side effect is a card like Freedmen, which triggers from discard pile by default, will always have its source location shown, which might or might not be desired.

server/game/gamesteps/triggeredabilitywindow.js
```diff
@@ -95,6 +95,9 @@ class TriggeredAbilityWindow extends BaseAbilityWindow {
             if(abilityChoice.text !== 'default') {
                 title += ' - ' + abilityChoice.text;
             }
+            if(abilityChoice.card.location !== 'hand' && abilityChoice.card.location !== 'play area') {
```
ystros: Per the test failures, you also need to allow it to be triggered from `'active plot'` and `'agenda'` locations without tacking on the location.

For Freedmen and things like The Prince's Plan that normally trigger from discard, you can probably boil this check down entirely to:
```javascript
if(abilityChoice.card.location !== abilityChoice.ability.location) {
```

which should only pick up abnormal cases (e.g. Annals event from discard)

cryogen:
I'm not sure I like it showing for when it's just triggering from the discard pile and there's no other effect to trigger

Feels like this should be a little more visual too, but happy for that to be part of a 2.0

PR #1532, opened 2017-10-12
Make payXGold cost function reducable

DukeTax:


server/game/costs/payxgoldprompt.js
```diff
@@ -18,21 +19,29 @@ class PayXGoldPrompt extends BaseStep {
...
 
-        let buttons = _.map(range, gold => {
-            return { text: gold.toString(), method: 'resolveCost', arg: gold };
+        let buttons = _.map(range, xValue => {
+            return { text: xValue.toString(), method: 'resolveCost', arg: xValue };
         });
 
+        let titleStr;
+        if(this.reduction > 0) {
+            titleStr = 'Select value of X (before reduction of ' + this.reduction + ')';
```
ystros: I'm not sure we need this specific of a title. It's also a bit long. I think just offering the buttons for the reducer-boosted gold values with the old title is sufficient.

PR #1534, opened 2017-10-12
Display aberrant ability source location in button prompt

### LINKED ISSUE #1154: Indicate location of event for Annals of Castle Black

DukeTax:
schermafbeelding 2017-10-05 om 15 38 42

Fixes #1154.


ystros:
I'm fine with this but sounded like @cryogen had some objections from the last PR.

cryogen:
I mean, it works, but it’s clumsier than I’d like

PR #1542, opened 2017-10-16
Fix NaN reserve bug

### LINKED ISSUE #1325: Wraiths in Their Midst combined with Rains leads to reserve bug
cameronhawkins17:59
Just hit a reserve bug: I used Wraiths in their Midst, opponent triggered Rains first turn. Wraiths applied to the new plot (as it should). The next turn, the opponent's reserve value was blank, even when he triggered Rains and revealed a 4th plot. During Taxation, the opponent was asked to discard down to 0 cards. Cancel-prompt could not bypass. Cancel-prompt should always bypass.
Opponent manually manipulating reserve did no good.

### LINKED ISSUE #1506: "Discard NaN cards to reserve"
image

Bit of a weird one. Game was progressing fine, then we hit end of turn and I had the message to discard "NaN cards" to reserve. I was definitely within reserve (1 card in hand etc). If I simply pressed "Done" it wouldn't advance past that screen. If I selected my card and hit "Done" nothing happened. If I dragged Balon to discard and hit Done nothing happened. If I picked him back up and hit Done or tried to select him and hit Done nothing happened. If I used /cancel-prompt nothing happened.

I was definitely still connected as me and my opponent were chatting about it in chat, and I tried refreshing the page twice.

ystros:
Previously, the setMinReserve effect stored the previous value for the
player's min reserve using the player ID as a key. This lead to a bug
where if a player who was affected by Wraiths in their Midst reconnected
after it was revealed but before the next plot was revealed, their ID
would change. Thus, when the effect was unapplied, the stored value
could not be retrieved because it was under the original ID.

Now, the original min reserve is stored under the player's name, which
should not be able to change mid-game.

Fixes #1506  
Fixes #1325

### DIFF
--- a/server/game/effect.js
+++ b/server/game/effect.js
@@ -1,5 +1,4 @@
 const _ = require('underscore');
-const logger = require('../log.js');
 
 const Effects = require('./effects.js');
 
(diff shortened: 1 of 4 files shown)


zacchiro:
w00t !! congrats for this

PR #1544, opened 2017-10-18
Minor bug fix that resolve user being stuck on game list.

Grugn0:
Bug description:
When user click on "watch" on a game still pending nothing happen, nevertheless he could no watch nor join any other game in list. Page refresh was needed or wait until game starts.

Ps. 
I hope I have not make any mess 'cause these are my very first steps with "git".


ystros:
Whoops. Thanks!

zacchiro:
nice catch, it has bitten me various times in the past, but i never tracked it down to this typo!

Issue #1104, opened 2017-06-30
Manual actions should override 'cannot' conditions


ystros:
* [x] Manually clicking a knelt card that cannot stand should stand it anyway.
* [x] Manually clicking a standing card that cannot kneel should kneel it anyway.
* [x] Dragging a card that cannot be killed into the dead pile should kill it anyway.
* [x] Dragging a card that cannot be discarded into the discard pile should discard it anyway.
* [x] Dragging a card that cannot be returned to hand into the hand should return it anyway.
* [x] Dragging a character into play when Barring the Gates is in effect should work.
* [x] Dragging a character into play when Iron Bank Will Have Its Due or Tris Botley is blocking that card name.

DukeTax:
- Using /power X on a card that cannot gain power should have it gain power anyway.

zacchiro:
> * Dragging a card that cannot be returned to hand into the hand should return it anyway.

is this one the same of #1216 or something else?

ystros:
Separate. This is about effect restrictions preventing certain player actions. #1216 is a bug where drag-and-drop is not implemented at all for non-standard piles (Out of Game, Schemes, Conclave).

ystros:
Another scenario to consider - playing The Iron Bank Will Have Its Due on a character should not prevent the player from manually dragging the character back into play if they want to undo the event.

PR #1548, opened 2017-10-19
Fix Mace Tyrell's remove from game ability

ystros:
Previously, when using Mace Tyrell's ability to remove a card from the
game, the card would leave then immediately re-enter play. This is
because the effects engine was fixed to remove effects from cards that
leave play.

Now, 'custom' duration lasting effects are not removed from cards when
they leave play. This change should have no effect on other cards using
custom lasting effects as they explicitly listen to onCardLeftPlay
already where appropriate.

Additionally, Barring the Gates will no longer prevent the card from
re-entering play.

### DIFF
--- a/server/game/effects.js
+++ b/server/game/effects.js
@@ -562,7 +562,7 @@ const Effects = {
                 card.owner.moveCard(card, 'out of game');
             },
             unapply: function(card, context) {
-                card.owner.putIntoPlay(card);
+                card.owner.putIntoPlay(card, 'play', { isEffectExpiration: true });
                 context.game.addMessage('{0} is put into play because of {1}', card, context.source);
             }
         };
--- a/server/game/effectengine.js
+++ b/server/game/effectengine.js
@@ -103,7 +103,7 @@ class EffectEngine {
     removeTargetFromEffects(card, location) {
         let area = location === 'hand' ? 'hand' : 'play area';
         _.each(this.effects, effect => {
-            if(effect.targetLocation === area && effect.location !== 'any' || location === 'play area' && effect.duration !== 'persistent') {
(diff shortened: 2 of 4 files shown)


server/game/effects.js
```diff
@@ -562,7 +562,7 @@ const Effects = {
                 card.owner.moveCard(card, 'out of game');
             },
             unapply: function(card, context) {
-                card.owner.putIntoPlay(card);
+                card.owner.putIntoPlay(card, 'play', { isEffectExpiration: true });
```
cryogen: It's not immediately obvious what this flag is for without looking at the implementation of putIntoPlay, is the implication that it will be forced into play? if so perhaps it should be called that

ystros: I considered that naming, but it won't quite force it into play. It will put it into play even if:
* Barring the Gates is out restricting cards from coming into play
* Iron Bank Will Have Its Due somehow activates on the card (e.g. somehow get another copy out and Iron Bank'd it - normally couldn't bring it back into play this round)
It will NOT bring the card back into play if:
* A copy of the card has ended up in the dead pile while the other copy was out of game (e.g. Tarred Heads, pillage + Ser Gregor, pillage + Crone of Vaes Dothrak, etc).

If we're okay with calling it `force` and having it NOT work when the character is in the dead pile, then we could do that. But that seems wrong. I figure we should save that for drag-drop support (which also currently won't let you drag a dead character into play, related #1104)

I could rename it `ignorePlayRestrictions` but that could also imply that it should ignore the restriction placed on the card by the dead pile.

cryogen: I see, tricky.

Doesnt feel quite right, but in lieu of a better suggestion (which i don't have right now) it can stay as is

PR #1549, opened 2017-10-20
Do not place events in discard immediately

ystros:
When an event card is played, it goes into a temporary "being played"
pile to indicate that it is currently resolving. Once the resolution of
the ability has completed, it is then placed in discard.

From a UI perspective, cards that are currently being played are
displayed as if they are already in the discard pile.

### DIFF
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -120,7 +120,11 @@ const Costs = {
                 return context.player.isCardInPlayableLocation(context.source, 'play') && context.player.canPlay(context.source, 'play');
             },
             pay: function(context) {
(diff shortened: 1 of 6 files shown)


ystros:
This is prep-work for implementing [Hidden Thorns](https://thronesdb.com/card/09024) and [Emissary of Hightower](https://thronesdb.com/card/09011) which add replacement effects to where event cards are placed after playing them. This is in contrast w/ Annals of Castle Black which is a forced reaction *after* a card has been placed in discard.

cryogen:
👍 just need to resolve the conflict

ystros:
Conflicts resolved, merging.

PR #1552, opened 2017-10-23
Implement Hidden Thorns

ystros:


ystros:
This feels hacky but the "correct" solution would be to implement first class support for delayed effects PLUS the newly introduced concept of a delayed replacement effect (more Emissary of Hightower than this one). I'd rather get the card working than try to implement a whole new subsystem under a release time crunch.

PR #1559, opened 2017-10-27
Combine fetchdata and importdata scripts

ystros:
Now, the fetchdata script can take options to determine card data source
(--card-source of either 'thronesdb' (default) or 'json') as well as an
image source (--image-source of 'thronesdb' (default), 'cardgamedb',  or
'none' for no images). A --no-images flag is provided as a shortcut for
skipping images.


cryogen:
The build servers dont seem to be able to build this, what additional (OS) packages are needed?

server/scripts/fetchdata/CardImport.js
```diff
@@ -0,0 +1,63 @@
...
+        this.db = db;
+        this.dataSource = dataSource;
+        this.imageSource = imageSource;
+        this.imageDir = imageDir;
+        this.cardService = new CardService(db);
+    }
+
+    async import() {
+        try {
+            await Promise.all([this.importCards(), this.importPacks()]);
```
cryogen: why not just `await this.importCards();\nawait this.importPacks(); `?

server/scripts/fetchdata/ThronesDbImageSource.js
```diff
@@ -0,0 +1,21 @@
+/*eslint no-console:0 */
+
+const request = require('request');
+const fs = require('fs');
+
+class ThronesDbImageSource {
+    fetchImage(card, imagePath) {
+        let url = `https://thronesdb.com/bundles/cards/${card.code}.png`;
```
cryogen: I see why youve done this but this means we're no longer using the imageUrl property of the card, as long as we're ok with that

server/scripts/fetchdata/CardImport.js
ystros: The `all` allows them to be fetched in parallel, which was the previous behavior. I could split it into two `await`s for clarities sake if you want, but that will change the behavior to fetch each piece of data sequentially.

cryogen: Ahha, of course, no it's find as is

server/scripts/fetchdata/ThronesDbImageSource.js
ystros: The `imagesrc` property used before was only ever being used in fetchdata, and really had no use outside of there (since it refers to a ThronesDB specific path). Since it's only available in the ThronesDB API response and not in the JSON data, had to get rid of its usage so that you could mix card data sources (JSON) with image data sources (ThronesDB)

ystros:
I was hoping the `imagemagick-native` package would automatically install imagemagick in the same way phantomjs-native installs phantom, but I guess I'll need to add it to the Travis config and it will have to be manually installed on Jenkins.

server/scripts/fetchdata/ThronesDbImageSource.js
cryogen: Yeah that's what i meant by i see why you've done it. It means that in the unlikely event thronesdb has a reorg it will break, but it's a reasonable risk to accept

PR #1581, opened 2017-11-06
Default fields for other player on client

ystros:
Instead of checking that otherPlayer is null for each field passed into
a component, a placeholder object is used with the fields pre-blanked.


cryogen:
I see what you're going for, but this is still pretty fugly.  Gameboard is still on my list of things to refactor

PR #1584, opened 2017-11-07
Implement Selyse Baratheon (FotOG), Dragonstone Castle & Greensight

DukeTax:


server/game/cards/08.4-FotOG/SelyseBaratheon.js
```diff
@@ -0,0 +1,28 @@
+const DrawCard = require('../../drawcard.js');
+
+class SelyseBaratheon extends DrawCard {
+    setupCardAbilities() {
+        this.forcedReaction({
+            when: {
+                onCardEntersPlay: event => event.card === this && event.playingType === 'marshal' && !this.controller.faction.kneeled
```
zacchiro: Given the card text uses "then", I think you'd be forced to kneel all non-Lady characters even if your faction card is initially knelt. If so, the last part of this test should be removed, and a corresponding if() should be added to the handler.

DukeTax: From the [RRG](http://thronesdb.com/rulesreference#The_word_Then):

> If the pre-then aspect of an effect does not successfully resolve in full, the post-then aspect does not attempt to resolve.

PR #1586, opened 2017-11-09
Only remove from a challenge characters that are leaving play if they were participating

DukeTax:


server/game/challenge.js
```diff
@@ -241,7 +241,9 @@ class Challenge {
     }
 
     onCardLeftPlay(event) {
-        this.removeFromChallenge(event.card);
+        if(this.isParticipating(event.card)) {
```
ystros: It might be better to check this directly in the removeFromChallenge method and return if it isn't participating. There could be other cases where `removeFromChallenge` is called on cards not necessarily in the challenge, which we wouldn't want to raise the event.

PR #1590, opened 2017-11-11
Extract plots to separate component

ystros:
This will probably conflict slightly with #1587, just let me know.

### DIFF
--- a/client/GameBoard.jsx
+++ b/client/GameBoard.jsx
@@ -11,7 +11,6 @@ import PlayerStats from './GameComponents/PlayerStats.jsx';
 import PlayerRow from './GameComponents/PlayerRow.jsx';
 import ActivePlayerPrompt from './GameComponents/ActivePlayerPrompt.jsx';
 import CardZoom from './GameComponents/CardZoom.jsx';
-import CardPile from './GameComponents/CardPile.jsx';
 import GameConfiguration from './GameComponents/GameConfiguration.jsx';
 import PlayerBoard from './GameComponents/PlayerBoard.jsx';
 import GameChat from './GameComponents/GameChat.jsx';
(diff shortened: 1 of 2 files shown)


client/GameBoard.jsx
```diff
@@ -280,28 +281,35 @@ export class InnerGameBoard extends React.Component {
...
-                    onMouseOver={ this.onMouseOver } onMouseOut={ this.onMouseOut } orientation='horizontal' onMenuItemClick={ this.onMenuItemClick }
-                    onCardClick={ this.onCardClick } onDragDrop={ this.onDragDrop } size={ this.props.user.settings.cardSize } />
-                <CardPile className={ thisPlayer.plotSelected ? 'plot plot-selected' : 'plot' }
-                    title='Plots' source='plot deck' cards={ thisPlayer.cardPiles.plotDeck } topCard={ { facedown: true, kneeled: true } } orientation='horizontal'
-                    onMouseOver={ this.onMouseOver } onMouseOut={ this.onMouseOut } onCardClick={ this.onCardClick } onDragDrop={ this.onDragDrop }
-                    closeOnClick size={ this.props.user.settings.cardSize } />
-                { this.getSchemePile(thisPlayer, !this.state.spectating) }
-            </div>
+            <PlayerPlots
+                { ...commonProps }
```
ystros: Let me know if you absolutely hate this syntax and I can drop the corresponding commit. It seemed like a good idea to help emphasize what's different between the two PlayerPlots components (i.e. mostly the player).

If you like it, I should probably go back and apply it to the CardPile rendering in PlayerPlots itself. There'd be more benefit there (3x mostly identical components) than there is here.

cryogen: I think it's fine, but going forward I probably want to encapsulate more of these things in an object that can either be saved/sent via redux or passed around whole so that it doesnt require a zillion attributes to every component.  This will do for now, though

cryogen:
There's a confict to resolve

ystros:
Resolved

PR #1593, opened 2017-11-11
Automatically close old pending games

ystros:
Check the list of games on the lobby every minute and automatically
close games that haven't started within 15 minutes of their creation
time.

### DIFF
--- a/server/lobby.js
+++ b/server/lobby.js
@@ -40,7 +40,7 @@ class Lobby {
 
         this.loadCardData();
 
-        setInterval(() => this.clearStalePendingGames(), 60000);
+        setInterval(() => this.clearStalePendingGames(), 60 * 1000);
     }
 
     async loadCardData() {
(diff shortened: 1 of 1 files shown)


server/lobby.js
```diff
@@ -39,6 +39,8 @@ class Lobby {
         this.lastUserBroadcast = moment();
 
         this.loadCardData();
+
+        setInterval(() => this.clearStalePendingGames(), 60000);
```
cryogen: I would typically break this out into its lower factors so it's slightly more clear what timeframe it is: ie 60 * 1000

server/lobby.js
```diff
@@ -242,6 +244,19 @@ class Lobby {
         this.broadcastGameList();
     }
 
+    clearStalePendingGames() {
+        const timeout = 900000;
```
cryogen: And here, while you're at it

PR #1604, opened 2017-11-20
Implement Stark, Greyjoy and Targaryen cards from TAK

DukeTax:


server/game/cards/08.1-TAK/BloodyArakh.js
```diff
@@ -0,0 +1,36 @@
...
+            },
+            handler: context => {
+                context.player.sacrificeCard(this);
+
+                this.untilEndOfPhase(ability => ({
+                    targetType: 'player',
+                    effect: ability.effects.modifyChallengeTypeLimit('military', 1)
+                }));
+
+                this.game.addMessage('{0} sacrifices {1} and is then able to initiate an additional {2} challenge this phase',
```
zacchiro: proposed simpler wording "… and can initiate an additional …"

server/game/cards/08.1-TAK/BloodyArakh.js
```diff
@@ -0,0 +1,36 @@
+const DrawCard = require('../../drawcard.js');
+
+class BloodyArakh extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            max: ability.limit.perPhase(1),
+            when: {
+                afterChallenge: event => event.challenge.winner === this.controller && event.challenge.challengeType === 'military' &&
```
zacchiro: Unless I'm missing something Bloody Arakh can be attached to opponent characters and, in that case, will need to trigger when the opponent wins a challenge, not the attachment controller. So this check needs to be a bit more complicated, unfortunately.

server/game/cards/08.1-TAK/DaarioNaharis.js
```diff
@@ -0,0 +1,33 @@
...
+                cardCondition: card => card.location === 'play area' && card !== this && (card.kneeled || card.controller !== this.controller) &&
+                                       (card.hasTrait('Ally') || card.hasTrait('Companion') || card.hasTrait('Mercenary'))
+            },
+            handler: context => {
+                if(context.target.kneeled) {
+                    context.target.controller.standCard(context.target);
+                }
+
+                if(context.target.controller !== this.controller) {
+                    this.untilEndOfRound(ability => ({
```
zacchiro: should be untilEndOfPhase

server/game/cards/08.1-TAK/NaggasRibs.js
```diff
@@ -0,0 +1,26 @@
+const DrawCard = require('../../drawcard.js');
+
+class NaggasRibs extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: this,
+            effect: ability.effects.modifyDominanceStrength(2 * this.controller.deadPile.size())
```
zacchiro: Might be a tad paranoid, but this will break if we'll ever have non-character cards in the dead pile. If we want to be defensive, this should count only character cards. Not sure if it's worth it.

server/game/cards/08.1-TAK/BloodyArakh.js
DukeTax: That's probably better. I've always tried to signify 'then'-effects by including the word, but they keep introducing weirder effects making for such awakward messages.

PR #1611, opened 2017-11-23
Redirect immediately after saving deck

ystros:
Previously, you had to click the "Add/Save" button when editing decks
twice before it would take you back to the deck list. It was saving both
times, but the navigation check was happening prior to the properties
being updated for the components. Thus, it would only detect the change
the second time the button was pressed.


cryogen:
This should actually use will receive props. It was on my list of things to refactor as part of the api/spinner/etc refactoring

Issue #1570, opened 2017-10-31
Issues with Highgarden Courtier


DukeTax:
* [ ] It's not possible to use Highgarden Courtier to stand Ser Colen of Greenpools when declaring him in a power challenge. This is due to the fact that during challenge initiation, challenge type is chosen (and used to recalculate effects by the engine) before attackers are chosen and knelt. So the moment Colen gets boosted, he is not knelt yet, meaning the Courtier can not trigger.

* [ ] When multiple character's strength are boosted at the same time (for example through Horn Hill), it is not possible to choose which character to stand with the Courtier. A possible (easy) solution would be to display the boosted character in the button prompt for the Courtier, although this could require clicking pass multiple times before the desired character is prompted. A better solution would be to have the boosting by such cards happen atomically.

Issue #1504, opened 2017-09-30
Marriage Pact + Valar


wj6527:
When I Valar and have a Marriage Pact on my char (who then dies to the Valar), the opponent is prompted to sacrifice a character only after Valar has resolved. I have been told it should be timed such that they can sacrifice something that is going to die to Valar for the Marriage Pact?

ystros:
Thanks for the report. Per the FAQ for Marriage Pact on ThronesDB and the general rules about kill, that ruling is correct - the owner of Marriage Pact should sacrifice a card before the simultaneous kills take place since it's a forced interrupt to leaving play and the thing triggering the leave play hasn't fully occurred yet (i.e. the kill).

This bug would also extend to if the character with Marriage Pact were sacrificed (e.g. Fallen from Favor), or discarded from play (e.g. Marched to the Wall). It just happens to matter a lot more during Valar since it allows the Stark player to avoid a kill.

PR #1615, opened 2017-11-29
Refactor event system

ystros:
Refactors the event system to have a hierarchy of event classes instead of separate classes that process the different kinds of events. Also introduces the concept of 'child events' which share the same interrupt / reaction windows as their parent (thus happening simultaneously) but can have separate handlers or be cancelled independently.

In general, it should be easier to compose events together and ensure consistent resolution, as well as allow new event types to be introduced in the future as needed.

**Motivation:**
* Allow atomic events have child events that happen at the same time but whose cancellation does not cancel the overall event. For example, if a Stealth choice is cancelled via [Fear Cuts Deeper Than Swords](https://thronesdb.com/card/04022), it should not cancel the overall challenge, only that stealth choice. Currently, you would have to cancel the entire challenge and start over.
* Start to make more use of atomic events in general. For example, the fix to #1504 and general rules require that the interrupt window for onCharacterKilled to have a simultaneous window for onCardLeftPlay. If either event is cancelled, both events are cancelled. Currently, they are improperly nested.
* Allow nested reaction windows to flow into each other. For example, when a character is killed, the reaction to it being killed and the reaction to it being placed in the dead pile are supposed to occur simultaneously. However, the interrupt windows for the two events can't be the same because the location isn't determined until the kill actually occurs (e.g. Benjen replaces the handler and shuffles the card back into deck). So the placement event needs to be merged into the existing kill event upon resolution. While this PR does not make that change, the concept of child events should allow that to occur.
* Allow more usage of simultaneous events. Currently we hard code both the killing and discarding of multiple cards into wrapped events (translated here as GroupedCardEvent) to simulate simultaneous resolution. But there are increasing situations where a card should be able to respond to the same event across multiple cards (Marya Seaworth vs multiple stealth bypass, #1570 Highgarden Courtier vs multiple strength increases). Ideally I want to make this automatic rather than manually wrap in a GroupedCardEvent. That will be difficult but this PR should help make that possible.


cryogen:
I'm not even going to pretend I've read even half of this :)

PR #1617, opened 2017-11-30
Implement (most) Sands of Dorne first preview cards

DukeTax:


server/game/cards/10-SoD/TyeneSand.js
```diff
@@ -0,0 +1,22 @@
+const DrawCard = require('../../drawcard.js');
+
+class TyeneSand extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                afterChallenge: event => event.challenge.loser === this.controller && event.challenge.defendingPlayer === this.controller
+            },
+            location: 'hand',
+            condition: () => this.controller.canPutIntoPlay(this),
```
ystros: `condition` doesn't work with reactions / interrupts, you'll need to check this in the `when` handler.

I really should add some checks when creating the ability to help catch things like this.

Utilisateur2:
Correct me if I'm wrong but right now the reaction part implemented for Doran Martell is based on if you target a character you control (gains an icon) or don't control (loses an icon) ?

If I understood that correctly, it can become problematic since you may want to remove one of your character icon or add an icon to an opponent character.

DukeTax:
That's correct, whether the icon is gained or lost is currently automated based on card controller. It makes for a clean card implementation and saves the user a click.

I'm not sure whether there are enough (viable) situations in which you'd want to have an opponent's character gain an icon, or an own character lose an icon, but I could be convinced.

Utilisateur2:
If melee is to be implemented, giving another player an icon for him to defend or attack in a challenge could be a viable option.

A niche case would be to give my opponent an icon that he doesn't currently have on the board to make him attack and resolve effects like "vengeance for elia", "his viper's eyes" that are played in martell

And currently, having multiple icons is beneficial but in the first edition, they printed characters for each houses that made having more than one icon not count your str. If they print similar effects in the future,
 it could be viable to give opp characters icons or remove yours.(http://www.cardgamedb.com/index.php/GoTCards.html/_/conquest-and-defiance/the-prize-of-the-north/ashemark-councilor-r1828 & http://www.cardgamedb.com/index.php/GoTCards.html/_/conquest-and-defiance/the-prize-of-the-north/riverlands-hunter-r1833 )

DukeTax:
Thanks. :) These are some good arguments.

PR #1620, opened 2017-11-30
Add some styling to the How to Play page

DukeTax:
schermafbeelding 2017-11-30 om 20 48 17


cryogen:
+1 for pictures

ystros:
This does look better! From an HTML / CSS perspective, it might be more semantic if the borders were done through only CSS and not by introducing the `<hr />` elements. `<hr />` is meant more to [divide the content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr) before and after than just render a line in between. Semantically, the titles and the contents should be considered grouped together, but adding `<hr />` causes them to be considered separated.

You may be able to get the same rendering via CSS using `border-bottom`. Normally borders are solid colors, but you can use the same gradient mechanism you're currently using for the background with [border-image](https://developer.mozilla.org/en-US/docs/Web/CSS/border-image#Gradient) property instead.

DukeTax:
Thanks for the feedback. Switched out the <ht/>s for border bottoms and some other stuff. Haven't looked into the gradient yet, but this looks fine as well (to me).

schermafbeelding 2017-12-01 om 21 57 30

PR #1625, opened 2017-12-02
Add alert when using actions outside of window

ystros:
To make it more clear that players are not following the rules, Action
abilities used outside an explicit action window now add a danger alert
in the game log.

screen shot 2017-12-01 at 11 35 18 pm

### DIFF
--- a/server/game/ambushcardaction.js
+++ b/server/game/ambushcardaction.js
@@ -12,6 +12,10 @@ class AmbushCardAction extends BaseAbility {
         this.title = 'Ambush';
     }
 
+    isAction() {
+        return true;
+    }
+
     isCardAbility() {
         return false;
     }
--- a/server/game/baseability.js
+++ b/server/game/baseability.js
@@ -177,7 +177,7 @@ class BaseAbility {
     }
 
     isAction() {
-        return true;
+        return false;
     }
 
     isPlayableEventAbility() {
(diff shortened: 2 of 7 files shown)


cryogen:
I think the wording is a little strong/negative

Perhaps err more on the side of pointing out it's a manually taken action?

Something like 'x has triggered y manually' ?

DukeTax:
Not a fan of 'manually' here, as in the posted example it's still an action taken care of by the automation (doing Qotho manually would print different alert messages).

cryogen:
"outside the proper action window" just isn't very beginner friendly imo - i think we can do better

ystros:
Agree it's a bit aggressive (particularly "proper" - "normal" would be a bit softer language-wise), but I also think it's important to mention "action window" somewhere in the message. Most people using actions outside of the window are probably doing so because they don't understand the rules super well rather than out of a desire to cheat.

As such, the message should achieve two things:
1. Alert both players that something against the normal rules has occurred.
2. At least point the player to why the alert is happening - if at all possible, in such a way that helps guide them to the proper way of doing it.

The second part is hard to achieve while keeping the message succinct. Need to think on it some more.

cryogen:
How about “x uses/plays/triggers/whatever-verb-you-want outside of an action window”?

zacchiro:
Thumbs up to latest @cryogen proposal, with a preference for "X *uses* foo outside of an action window", because it's the most generic verb option.

PR #1631, opened 2017-12-06
Implement Arianne Martell (JtO) & Change of Plans

DukeTax:
https://scontent-amt2-1.xx.fbcdn.net/v/t1.0-9/24296270_1736675056365233_1355402977147842896_n.jpg?oh=a04154f29b4d9df0512de56487acd067&oe=5A8CB8A1

https://scontent-amt2-1.xx.fbcdn.net/v/t1.0-9/24174221_1736675059698566_6395687336416462752_n.jpg?oh=1a814f183734d6d7e70d3c2e6e4754e6&oe=5AC5D07A

### DIFF
--- a/server/game/cards/08.2-JtO/ArianneMartell.js
+++ b/server/game/cards/08.2-JtO/ArianneMartell.js
@@ -4,10 +4,15 @@ class ArianneMartell extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
             title: 'Return character to hand',
+            //TODO: This should ideally use the snapshot functionality instead
+            condition: () => {
+                this.strengthWhenInitiated = this.getStrength();
+                return true;
+            },
             cost: ability.costs.returnSelfToHand(),
             target: {
                 cardCondition: card => card.location === 'play area' && card.getType() === 'character' &&
-                                       card.getStrength() < this.getStrength()
+                                       card.getStrength() < this.strengthWhenInitiated
             },
             handler: context => {
                 context.target.owner.returnCardToHand(context.target);
--- a/server/game/cards/08.2-JtO/ArianneMartell.js
+++ b/server/game/cards/08.2-JtO/ArianneMartell.js
@@ -1,0 +1,23 @@
+const DrawCard = require('../../drawcard.js');
+
+class ArianneMartell extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Return character to hand',
+            cost: ability.costs.returnSelfToHand(),
+            target: {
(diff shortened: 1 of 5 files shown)


server/game/player.js
```diff
@@ -628,7 +628,11 @@ class Player extends Spectator {
     recyclePlots() {
         if(this.plotDeck.isEmpty()) {
             this.plotDiscard.each(plot => {
-                this.moveCard(plot, 'plot deck');
+                if(this.agenda && this.agenda.code === '05045' && plot.hasTrait('Scheme')) {
```
ystros: It would be better if the Player class had no knowledge whatsoever of the rules for the Rains agenda. Is there an alternate way for this to happen on the RainsOfCastamere class instead?

Perhaps if Player moves them all to the plot deck per normal rules, raises an event after the plots have been recycled, which Rains then listens for and then re-separates the Scheme pile (like it does onDeckPrepared at the start of the game)? You get a double move which could be a problem for future cards, but it wouldn't be a problem currently.

You might also be able to listen to the existing onCardMoved event for Schemes being moved to the plot deck and move it again. Same problem of double move but wouldn't require introducing a new event. Though additional care may be required there - that event is fired on every card move, it may interfere with the existing onDecksPrepared (might make it no longer needed, might interact unpredictably), etc.

server/game/cards/08.2-JtO/ArianneMartell.js
```diff
@@ -0,0 +1,23 @@
+const DrawCard = require('../../drawcard.js');
+
+class ArianneMartell extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Return character to hand',
+            cost: ability.costs.returnSelfToHand(),
+            target: {
+                cardCondition: card => card.location === 'play area' && card.getType() === 'character' &&
+                                       card.getStrength() < this.getStrength()
```
ystros: You need to get the snapshotted strength for Arianne instead of just `this.getStrength()`, since returning her to hand will reset her strength to always be the printed value.

DukeTax:
Thanks for the quality feedback as usual.

As discussed over Discord, using the snapshot functionality for Arianne Martell requires the snapshot to be taken earlier than at the start of the ability resolver, as eligible targets are checked before that. Taking the snapshot as soon as the context object is created is probably right, but causes other pipeline errors. For now, saving the strength at initiation on the card itself through a condition function is ugly, but works.

jeremylarner:
Why can't you just use the snapshot if it exists, and use getStrength() if not?

DukeTax:
Yeah, that should pretty much result in the same thing, looking slightly better.

PR #1632, opened 2017-12-06
Clarify second Mag kill to resolve only when first was not saved

DukeTax:
Have already seen this go wrong a couple of times, so a reminder seems appriopriate.

mag_screen

### DIFF
--- a/server/game/cards/08.1-TAK/MagTheMighty.js
+++ b/server/game/cards/08.1-TAK/MagTheMighty.js
@@ -15,6 +15,7 @@ class MagTheMighty extends DrawCard {
 
                 //TODO Technically should only trigger when the first kill was not saved
                 this.game.promptForSelect(context.event.challenge.loser, {
+                    activePromptTitle: 'Select a character (only when first kill was not saved)',
                     cardCondition: card => card.location === 'play area' && card.controller === context.event.challenge.loser && card.getType() === 'character',
                     source: this,
                     onSelect: (player, card) => this.onCardSelected(player, card)


zacchiro:
Shouldn't we simply avoid prompting if the first kill was saved, rather than this messaging workaround?

ystros:
To re-litigate around the current implementation for Mag - like most other actions, kill is not an immediate kill, so the player will be prompted with normal interrupts / reactions. So by the time it reaches the code that generates the prompt (which simply queues the prompt into the pipeline after the "kill" step, not immediately prompts), the kill has not resolved.

You could get around this by queuing a simple step, checking the card's location in the step and then only prompting if it is now in the dead pile. Except that isn't correct either - in the case of Benjen, Davos, or Spears of the Merling King, the kill was not cancelled but the character isn't in the dead pile. What you need to actually do is check that the previous kill event wasn't cancelled (not possible to do currently).

In general, I'd rather be consistent here (always prompt and rely on player action) than introduce subtle bugs (automate for 98% of cases but leave player's heads scratching the 2% of times it doesn't work). Which is why it was implemented this way for now w/ the TODO. The fact that Benjen is a prime target for the kind of deck that may use Mag makes it even more likely players would run into such a bug.

I'm fine w/ using the message as a temp fix until it can be automated properly (first-class support for "Then" in the engine). I haven't merged it, though, because I tend to be less picky about messaging.

Duckbringer:
I am not sure how do you implement this, but shouldn't killed characters (or rather killcharacter function) have some return value as to indicate the character was really killed? How do you handle Funeral pyre for example?

ystros:
Funeral Pyre is easy because it is a reaction to a character being killed - if the kill is cancelled, it never gets to the reaction stage. Whereas Mag needs to both initiate the kill and then wait until the kill resolves to continue the ability.

`killCharacter` can't return a simple value about whether the character was killed or not. It initiates the kill but does not fully resolve it - because you need to allow interrupts to be triggered by the players. Interrupts is where the kill is usually cancelled, so you don't know if the kill succeeded until the interrupt window closes. To properly implement 'Then' abilities, it would need to return something that executes asychronously (either a `Promise` or something that operates like a promise). I imagine it would look something like:

```javascript
this.game.killCharacter(context.target).thenResolve(() => {
  // Post-then portion of the ability
});
```

But this is beyond the scope of this specific pull request, which is meant as a temporary quick-fix.

PR #1635, opened 2017-12-08
Flag unreleased agendas in deck validation

ystros:


zacchiro:
Thanks for this. It happened several time to me to play against people using House with the Red Door while I could swear I didn't see the "not valid" warning before game. … and now I realize it was indeed the case :-)

PR #1648, opened 2017-12-18
Implement The House of Black and White, Stormcrows & Devan Seaworth

DukeTax:


Grugn0:
Regarding to the House of Black and White. the minimum gold to discard is 1 (and it should not depend on characters printed strengths currently in play).
It's obvious that you may not be able to kill anyone if you discard only 1 gold. but you should be able to do it just to do the second part of the effect (move gold).
The second part is not linked with the first part (by the word "then"), and iit's likely "castle black"; Kneel Castle black to just give a standing participating defender the +2 to his strenght

DukeTax:
The minimum passed into the cost function is minimum'd itself at 1 in the function (line 60).

I'm quite sure you do need to kill something in order to be able to refill. The kill can be saved for you to be able to refill, as the refill is not a 'then'-effect. But since House of B&W is a targeting ability, it can only resolve if you actually have a legal target.

ystros:
I'm pretty sure @DukeTax is right. Castle Black, for instance, the target gets two effects (stand and +2 STR) and is eligible as long as it can receive one of those effects. For House of Black + White, the targeting requirement is that there is a character that 1. has printed cost X or less, 2. can be killed. 

Per the RRG:
> If an ability requires the choosing of a target (or targets), and there is no valid target (or not enough valid targets), then the ability cannot be initiated. 

So if there isn't a killable character, you can't initiate the ability and thus can't move the additional gold onto the card. It would be different if the "You may move" portion were either a separate constant ability, or as a separate Action ability. But as printed, it seems like you need to kill first if you want to move more gold onto it.

PR #1673, opened 2018-01-07
Clarify discard to reserve message

DukeTax:


server/game/gamesteps/taxation/discardtoreserveprompt.js
```diff
@@ -38,7 +38,7 @@ class DiscardToReservePrompt extends BaseStep {
         // on the top of the discard pile.
         cards = cards.reverse();
         player.discardCards(cards, false, () => {
-            this.game.addMessage('{0} discards {1}', player, cards);
+            this.game.addMessage('{0} discards {1} to match reserve', player, cards);
```
cryogen: As a minor nitpick i'd say that it probably makes more sense to say 'meet reserve' here

DukeTax: That's better. 👍

PR #1674, opened 2018-01-08
Only set cardtype to attachment for Risen & BMR if attached

DukeTax:


test/server/cards/01-Core/RisenFromTheSea.spec.js
```diff
@@ -73,7 +73,7 @@ describe('Risen from the Sea', function() {
 
             it('should not attach the event to the character', function() {
                 expect(this.noAttachmentCharacter.attachments.size()).toBe(0);
-                expect(this.event.location).toBe('being played');
+                expect(this.event.location).toBe('discard pile');
```
ystros: This should have been an obvious indication something was wrong before this PR. :woman_facepalming:

server/game/cards/01-Core/RisenFromTheSea.js
```diff
@@ -14,7 +14,10 @@ class RisenFromTheSea extends DrawCard {
             handler: context => {
                 context.event.saveCard(context.target);
                 this.controller.attach(this.controller, this, context.target, 'play');
-                this.setCardType('attachment');
+
+                if(this.parent) {
```
ystros: I'd maybe check `this.controller.canAttach(this, context.target)` instead and move the actual `attach` call inside the the if. Result is the same - at least currently - but it reads better and should be slightly more resilient to any changes made to the attachment system.

PR #1676, opened 2018-01-10
Fixes to attachment styling

ystros:
* Increases the display offset for attachments when the card size is above the default
* Fixes spacing issues for kneeled attachments when the card size is not the default
* Increases the styling limit for attachments up to 10 attachments on a single card (previously 5)
* Properly clear "new" status from attachments.

Also offsets attachments to the right as suggested in lobby chat. This should make clicking on attachments a bit easier overall. 
screen shot 2018-01-09 at 7 03 00 pm

If we don't want the offset (or want to think about it more), it's easy enough to drop the specific commit.

### DIFF
--- a/less/cards.less
+++ b/less/cards.less
@@ -249,7 +249,7 @@
   .generate-attachments(@n, (@i + 1));
 }
 
-.generate-attachments(5, 1);
+.generate-attachments(10, 1);
 
 .card-dupe {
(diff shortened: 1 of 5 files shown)


ystros:
Per Discord discussion, I've removed the right offset for now. We'll look for other solutions to the clickability problem.

PR #1677, opened 2018-01-11
Implement Martell cards from Kingsmoot

DukeTax:


server/game/cards/08.3-Km/SandSteed.js
```diff
@@ -0,0 +1,20 @@
+const DrawCard = require('../../drawcard.js');
+
+class SandSteed extends DrawCard {
+    setupCardAbilities() {
+        this.attachmentRestriction(card => card.getType() === 'character' && card.attachments.every(attachment => attachment.name !== 'Sand Steed')),
```
ystros: This is clever and works when initially attached. However, when attachment validity is checked after Sand Steed is attached, it will fall off because there's a copy attached already (itself). Even though we don't check attachment validity as much as we should currently, it's still checked when the character is blanked / unblanked. So it's still a problem currently if, say, Fortified Position is played or if the character is Milk'ed.

You should be able to get around this by checking that the attachment is the current card:
```javascript
card.attachments.every(attachment => attachment === this || attachment.name !== 'Sand Steed')
```

PR #1682, opened 2018-01-13
Add a 'singleController'-flag to the target API

### LINKED ISSUE #1669: Consolidation of Power
Consolidation of Power allows the selection of a character from either player, when it is supposed to only be characters controlled by one player.

DukeTax:
This flag restricts SelectCardPrompt targets to a single controller. Fixes a bug where Consolidation of Power could be used to select characters from either player.

Fixes #1669.


ystros:
Hmm. I'd probably prefer something more general purpose / less specific. But until we run into a case that needs it, this should work for now.

PR #1690, opened 2018-01-16
More refactoring

cryogen:
image
image
image


client/reducers/cards.js
```diff
@@ -60,11 +59,16 @@ export default function(state = {}, action) {
...
 
-            return Object.assign({}, state, {
+            newState = Object.assign({}, state, {
                 cards: action.response.cards,
                 agendas: agendas,
                 banners: banners
             });
+
+            // In case the card list is received after the decks, updated the decks now
+            processDecks(newState.decks, newState);
```
ystros: I'm mixed on this. On the one hand, it's necessary if decks come back before the list of cards. On the other, I think it would be better to put the deck list pages into a "loading" state if the cards are missing. Because as it stands, the cards in the deck will all be named as "undefined" until the cards load. (A loading state would be also useful for "Why are all my decks gone?" questions we get when the deck load fails / takes a long time)

But since this PR is primarily about styling / component extraction, we should fix that separately.

PR #1691, opened 2018-01-16
Implement Tyrell cards from Kingsmoot

DukeTax:


server/game/cards/08.3-Km/SerHyleHunt.js
```diff
@@ -0,0 +1,24 @@
+const DrawCard = require('../../drawcard.js');
+
+class SerHyleHunt extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Remove character from challenge',
```
ystros: Missing limit.

PR #1697, opened 2018-01-18
implement Targaryen cards from Kingsmoot

zacchiro:
- Jhiqui
- The Skahazadhan

### DIFF
--- a/server/game/cards/08.3-Km/Jhiqui.js
+++ b/server/game/cards/08.3-Km/Jhiqui.js
@@ -1,0 +1,28 @@
+const DrawCard = require('../../drawcard.js');
+
+class Jhiqui extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                afterChallenge: ({challenge}) => challenge.winner === this.controller
+            },
+            cost: ability.costs.discardFromHand(),
+            target: {
+                type: 'select',
+                cardCondition: card =>
+                    card.location === 'play area' && card.getType() === 'character'
+                        && this.game.currentChallenge.isParticipating(card)
+                        && (card.hasTrait('Lord') || card.hasTrait('Lady'))
+            },
+            handler: context => {
+                context.target.modifyPower(1);
+                this.game.addMessage('{0} uses {1} and discards {2} to have {3} gain 1 power',
+                    this.controller, this, context.costs.discardFromHand, context.target);
+            }
+        });
+    }
+}
+
+Jhiqui.code = '08053';
+
+module.exports = Jhiqui;
--- a/server/game/cards/08.3-Km/TheSkahazadhan.js
+++ b/server/game/cards/08.3-Km/TheSkahazadhan.js
@@ -1,0 +1,27 @@
+const _ = require('underscore');
+
+const DrawCard = require('../../drawcard.js');
+
+class TheSkahazadhan extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onPlotsRevealed: event => _.any(event.plots, plot => plot.controller === this.controller)
+            },
+            cost: [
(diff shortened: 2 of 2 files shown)


server/game/cards/08.3-Km/Jhiqui.js
```diff
@@ -0,0 +1,37 @@
...
+        this.reaction({
+            when: {
+                afterChallenge: ({challenge}) =>
+                    challenge.winner === this.controller && this.hasParticipatingLordOrLady()
+            },
+            cost: ability.costs.discardFromHand(),
+            target: {
+                activePromptTitle: 'Select a character',
+                cardCondition: card =>
+                    card.location === 'play area' && this.game.currentChallenge.isParticipating(card)
```
DukeTax: The target character needs to be a Lord or Lady.

Also, this ability is not a "real" target (doesn't include the word 'choose'), which is taken care of by adding ```type: 'select'``` to target. Lastly, 'Select a character' is the default prompt title so you could get rid of that.

server/game/cards/08.3-Km/Jhiqui.js
```diff
@@ -0,0 +1,37 @@
...
+            handler: context => {
+                context.target.modifyPower(1);
+                this.game.addMessage('{0} uses {1} and discards {2} to have {3} gain 1 power',
+                    this.controller, this, context.costs.discardFromHand, context.target);
+            }
+        });
+    }
+
+    hasParticipatingLordOrLady() {
+        let cards = this.controller.filterCardsInPlay(card => {
```
DukeTax: This could be done more easily with:

```javascript
return this.controller.anyCardsInPlay(card => this.game.currentChallenge.isParticipating(card) && (card.hasTrait('Lord') || card.hasTrait('Lady'))
```

Also, technically you should be able to win a challenge, and then trigger Jhiqui to give an opponent's Lord or Lady 1 power, which should then be reflected in this check. I'm not sure if that's something we should worry about, but if you do, checking ```this.game.anyCardsInPlay()``` as above would take care of that.

server/game/cards/08.3-Km/TheSkahazadhan.js
```diff
@@ -0,0 +1,21 @@
+const DrawCard = require('../../drawcard.js');
+
+class TheSkahazadhan extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onPlotsRevealed: () => true
+            },
+            handler: context => {
```
DukeTax: Forgot the cost property.

zacchiro:
Thanks for the review! All of your comments should now be addressed in the current version of this PR.

server/game/cards/08.3-Km/TheSkahazadhan.js
```diff
@@ -0,0 +1,22 @@
+const DrawCard = require('../../drawcard.js');
+
+class TheSkahazadhan extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onPlotsRevealed: () => true
+            },
+            cost: ability.costs.discardFromHand(),
```
DukeTax: Only missing the kneelSelf() cost here (and you could incorporate that in the message as well if you like).

server/game/cards/08.3-Km/TheSkahazadhan.js
```diff
@@ -0,0 +1,25 @@
+const DrawCard = require('../../drawcard.js');
+
+class TheSkahazadhan extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onPlotsRevealed: () => true
```
ystros: This isn't quite correct. The triggering condition is "when **you** reveal a plot card." As it stands, this will trigger when your opponent reveals a plot using Rains. See Caswell's Keep for the proper condition.

PR #1705, opened 2018-01-22
Allow effects to be applied before Bestow prompt

### LINKED ISSUE #1462: Fortified Position should prevent Bestow as the character is blank immediately upon entering play

ystros:
Previously, if a character with Bestow came out during Fortified
Position, the player would be prompted for the bestow amount even though
Bestow keyword would be blanked. Now the prompt is delayed until after
the effects engine has a chance to apply effects to the card.

Fixes #1462

### DIFF
--- a/server/game/drawcard.js
+++ b/server/game/drawcard.js
@@ -138,7 +138,7 @@ class DrawCard extends BaseCard {
     }
 
     isBestow() {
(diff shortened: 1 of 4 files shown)


ystros:
This isn't my preferred way to fix this, but it is the lowest impact fix without reworking the event system and the way effects are applied.

PR #1727, opened 2018-02-07
Timeout games after finished instead of created

ystros:


ystros:
Updated to only look at `finishedAt` time (which should cover wins, concedes, and leaves) and not disconnects.

PR #1735, opened 2018-02-13
Implement Nymeria (FotOG)

DukeTax:


server/game/costs/SelectCardCost.js
```diff
@@ -35,7 +36,11 @@ class SelectCardCost {
                 return true;
             },
             onCancel: () => {
-                result.value = false;
+                if(this.proceedOnCancel) {
+                    result.value = true;
```
ystros: Stylistically, this could be rephrased as just:
```javascript
result.value = this.proceedOnCancel;
```

PR #1741, opened 2018-02-14
Update fetchdata script

ystros:
ThronesDB no longer hosts their own images, so it should no longer be
used as an image source.


cryogen:
Sad 😢

PR #1752, opened 2018-02-20
implement Tyrell cards from Someone Always Tells chapter pack

zacchiro:
Not to be merged yet, as we don't have neither the corresponding JSON data yet,
nor the proper english card name for the event. For review purpose, I've kept
at the beginning of each card implementation a URL pointing to spoiler image;
it should be removed before merging too.

### DIFF
--- a/server/game/cards/08.6-SAT/MaceTyrell.js
+++ b/server/game/cards/08.6-SAT/MaceTyrell.js
@@ -25,7 +25,7 @@ class MaceTyrell extends DrawCard {
 
     putCardBack(player, card) {
         card.owner.moveCardToTopOfDeck(card);
-        this.game.addMessage('{0} places a card on top of their draw deck due to {1}', this.controller, this);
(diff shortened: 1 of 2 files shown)


server/game/cards/08.6-SAT/MaceTyrell.js
```diff
@@ -0,0 +1,45 @@
...
+        this.reaction({
+            when: {
+                onCardKneeled: event => event.card === this
+            },
+            handler: () => {
+                this.controller.drawCardsToHand(2);
+                this.game.addMessage('{0} uses {1} to draw 2 cards', this.controller, this);
+
+                this.game.promptForSelect(this.controller, {
+                    numCards: 1,
```
DukeTax: The numCards property isn't needed, and I think it causes the prompt to not fire on selection as well (meaning it requires an extra click on the 'Done'-button currently).

DukeTax:
👍

PR #1753, opened 2018-02-20
Create SweetCersei.js

hanneshelge:
Hi, I really like theironthrone.net and would like to help developing it. To get a feeling how the code works I started this partial integration of 'Sweet Cersei' It should work as far that gold is added to the card but there is no mechanic to spend it yet.
I would like to wait for the way the 'Iron Bank' is implemented and when take the same approach for the matter of consistency.

### DIFF
--- a/SweetCersei.js
+++ b/SweetCersei.js
@@ -4,7 +4,7 @@ class SweetCersei extends DrawCard {
     setupCardAbilities(ability) {
         this.reaction({
             when: {
-                afterChallenge: ({challenge}) => challenge.winner === this.controller && 'intrigue' && !this.controller.cannotGainGold
+                afterChallenge: event => event.challenge.winner === this.controller && event.challenge.challengeType === 'intrigue'
             },
             handler: () => {
                 this.modifyToken('gold', 1);
--- a/SweetCersei.js
+++ b/SweetCersei.js
@@ -1,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+
+class SweetCersei extends DrawCard {
+    setupCardAbilities(ability) {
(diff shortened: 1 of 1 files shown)


SweetCersei.js
```diff
@@ -0,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+
+class SweetCersei extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                afterChallenge: ({challenge}) => challenge.winner === this.controller && 'intrigue' && !this.controller.cannotGainGold
```
DukeTax: Couple of comments:

- The ```player.cannotGainGold``` flag is used to switch on/off the effect of Devout Freerider, which is not applicable to the effect of this card.
- You want to check 'intrigue' vs. the challengeType of the challenge.
- Even though there's a bunch of them left in the code, we try to not deconstruct event objects anymore.

Putting these together, this should do the trick:

```javascript
afterChallenge: event => event.challenge.winner === this.controller && event.challenge.challengeType === 'intrigue'
```

ystros:
Thanks for this @hanneshelge! I do have some reservations about a partial implementation, but I'll merge anyway. Players will still need to use manual commands to "spend" the money for the moment, but not having to use manual commands each round to load up gold is probably a net positive.

If you wouldn't mind, could you also do a similar partial implementation for Iron Bank (the gold doubling mechanic), so both of these cards are at the same level of implementation? If you can't or don't want to, no sweat, we can get one in pretty easy. In the meantime, I'll continue to look at options on how to implement the "spend" portion of the two cards.

PR #1756, opened 2018-02-20
implement Host of the Boneway (SAT)

zacchiro:


DukeTax:
This can go in right? Do you want to keep the commented image link?

zacchiro:
Right, this can go in (modulo the fact that I don't think its JSON data is available yet, but I'm not sure that's a blocker). Anyway, I've just updated the PR to drop the comment pointing to the spoiler image.

PR #1757, opened 2018-02-20
implement Arya Stark (TFM)

zacchiro:


server/game/cards/08.5-TFM/AryaStark.js
```diff
@@ -0,0 +1,15 @@
+const DrawCard = require('../../drawcard.js');
+
+class AryaStark extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
```
DukeTax: This needs a ```targetController: 'any'``` flag, as the effect will default to characters you control otherwise.

zacchiro: tnx, fixed!

PR #1761, opened 2018-02-22
Partly implementation of The Iron Bank

hanneshelge:
This is a partly implementation of  'The Iron Bank' (The Archmasters Key).

What this should do: Calculate the interest for the Iron bank.
What this does not do: Provide ways to use the gold

### DIFF
--- a/server/game/cards/08.1-TAK/TheIronBank.js
+++ b/server/game/cards/08.1-TAK/TheIronBank.js
@@ -7,7 +7,7 @@ class TheIronBank extends DrawCard {
                 onIncomeCollected: event => event.player === this.controller
             },
             handler: () => {
-                var interest = this.tokens['gold'];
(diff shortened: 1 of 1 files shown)


server/game/cards/08.1-TAK/TheIronBank.js
```diff
@@ -0,0 +1,20 @@
+const DrawCard = require('../../drawcard.js');
+
+class TheIronBank extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onIncomeCollected: event => event.player === this.controller
+            },
+            handler: () => {
+                var interest = this.tokens['gold'];
```
DukeTax: We generally declare variables using ```let``` instead of ```var```. Looks good otherwise.

PR #1765, opened 2018-02-22
port Plaza of Pride to the target API

zacchiro:


server/game/cards/07-WotW/PlazaOfPride.js
```diff
@@ -1,35 +1,27 @@
...
             cost: [
                 ability.costs.kneelSelf(),
                 ability.costs.discardFromHand()
             ],
+            target: {
+                cardCondition: (card, context) =>
+                    card.location === 'play area'
+                    && card.getType() === 'character'
+                    && card.kneeled
+                    && card.getCost() <= context.costs.discardFromHand.getCost() + 3
```
DukeTax: Because targets are checked before costs are paid, this will likely crash (probably a reason this card wasn't converted to the target API earlier). You can look at Tower of the Hand for an example to get around this.

PR #1768, opened 2018-02-23
implement Mhysa (TFM)

zacchiro:
French spoiler image here: https://i.redd.it/cp5ydmxlchg01.jpg

When ready this PR can be merged, as we are (reasonably) sure the english card
name will be the same of the french one.

### DIFF
--- a/server/game/cards/08.5-TFM/Mhysa.js
+++ b/server/game/cards/08.5-TFM/Mhysa.js
@@ -1,0 +1,21 @@
+const DrawCard = require('../../drawcard.js');
+
+class Mhysa extends DrawCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction({ trait: 'Lady' });
+        this.whileAttached({
+            condition: () =>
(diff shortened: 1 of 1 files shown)


server/game/cards/08.5-TFM/Mhysa.js
```diff
@@ -0,0 +1,21 @@
+const DrawCard = require('../../drawcard.js');
+
+class Mhysa extends DrawCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction({ trait: 'Lady' });
```
DukeTax: I think it needs to be a unique Lady character.

zacchiro: > I think it needs to be a unique Lady character.

Nope. "Personnage Lady uniquement" means "Lady character only".

DukeTax: Aah ok. 👍 Tricksy French language!

PR #1769, opened 2018-02-23
partially implement Daenerys Targaryen (TFM)

zacchiro:
French spoiler: https://i.redditmedia.com/SxCGctsBb71h87NWgeUy8uFPE0JKbC_U2rPjGr-N2vg.jpg?w=500&s=e8cdc218e218f28254767e63c69aba20

### DIFF
--- a/server/game/cards/04.6-TC/RenlysPavilion.js
+++ b/server/game/cards/04.6-TC/RenlysPavilion.js
@@ -8,7 +8,8 @@ class RenlysPavilion extends DrawCard {
             targets: {
                 toLower: {
                     activePromptTitle: 'Select a character to get -1 STR',
-                    cardCondition: card => this.cardCondition(card)
+                    cardCondition: card => this.cardCondition(card),
+                    gameAction: 'decreaseStrength'
                 },
                 toRaise: {
                     activePromptTitle: 'Select a character to get +1 STR',
--- a/server/game/cards/06.2-GtR/ADragonIsNoSlave.js
+++ b/server/game/cards/06.2-GtR/ADragonIsNoSlave.js
@@ -6,7 +6,8 @@ class ADragonIsNoSlave extends DrawCard {
             title: 'Give -2 STR',
             phase: 'challenge',
             target: {
-                cardCondition: card => card.location === 'play area' && card.getType() === 'character' && card.attachments.size() === 0
+                cardCondition: card => card.location === 'play area' && card.getType() === 'character' && card.attachments.size() === 0,
+                gameAction: 'decreaseStrength'
             },
             handler: context => {
                 this.untilEndOfPhase(ability => ({
--- a/server/game/cards/06.3-TFoA/Astapor.js
+++ b/server/game/cards/06.3-TFoA/Astapor.js
@@ -10,7 +10,8 @@ class Astapor extends DrawCard {
                 cardCondition: card => (
                     card.location === 'play area' &&
                     card.getType() === 'character' &&
-                    this.game.currentChallenge.isParticipating(card))
+                    this.game.currentChallenge.isParticipating(card)),
+                gameAction: 'decreaseStrength'
             },
             handler: context => {
                 this.untilEndOfChallenge(ability => ({
--- a/server/game/cards/07-WotW/WeirwoodBow.js
+++ b/server/game/cards/07-WotW/WeirwoodBow.js
@@ -14,7 +14,8 @@ class WeirwoodBow extends DrawCard {
                 cardCondition: card => (
                     card.location === 'play area' &&
                     card.getType() === 'character' &&
-                    this.game.currentChallenge.isDefending(card))
+                    this.game.currentChallenge.isDefending(card)),
+                gameAction: 'decreaseStrength'
             },
             handler: context => {
                 this.untilEndOfChallenge(ability => ({
--- a/server/game/cards/08.2-JtO/ConsumingFlames.js
+++ b/server/game/cards/08.2-JtO/ConsumingFlames.js
@@ -7,8 +7,10 @@ class ConsumingFlames extends DrawCard {
             condition: () => this.game.currentChallenge,
             cost: ability.costs.discardFromHand(card => card !== this && card.isFaction('targaryen')),
             target: {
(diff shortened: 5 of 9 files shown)


zacchiro:
Second commit in this PR now has a proposal of how to implement Dany immunity. It doesn't quite work yet though: STR-related tests fail with `this.game.applyGameAction is not a function`. It's not clear to me if that's because I'm using `applyGameAction` incorrectly, or if something else needs to be mocked in the tests.
@ystros: advice on this and more generally on the proposed approach are very welcome.

server/game/drawcard.js
```diff
@@ -188,6 +188,20 @@ class DrawCard extends BaseCard {
         });
     }
 
+    modifyStrength(amount, applying = true) {
+        if(amount > 0) {
+            this.game.applyGameAction('increaseStrength', this, card => {
+                card._modifyStrength(amount, applying);
+            });
+        } else if(amount < 0) {
+            this.game.applyGameAction('decreaseStrength', this, card => {
```
ystros: While using `applyGameAction` is probably the right approach (tests are failing likely due to game not being mocked), this usage is not going to be correct. Since you're only checking that the amount is less than 0 and not whether the effect is being applied or unapplied (the `applying` param), this will prevent Dany from **expiring** any STR increases. i.e.

* Song of Summer is revealed. 
* `dany.modifyStrength(1, true)` is called. 
* Dany's STR is now +1
* A new plot is revealed. Song of Summer's effect is unapplied.
* `dany.modifyStrength(-1, false)` is called.
* Dany's STR remains +1 because the effect doesn't take into consideration that it's a positive buff being unapplied rather than a negative effect being applied.

In general, I think it might be better to do this check from the effect rather than on the player. I've been trying to pull a lot of these types of checks (as well as raising events) off of `DrawCard` and `Player` in prep for #1799 and other changes. Something like the following may work better:
```javascript
modifyStrength: function(value) {
    let gameAction = value < 0 ? 'decreaseStrength' : 'increaseStrength';
    return {
        apply: function(card, context) {
            context.game.applyGameAction(gameAction, card, card => {
                card.modifyStrength(value, true);
            });
        },
        unapply: function(card, context) {
            // Use the same game action as the apply - e.g. if they were immune to the increase then we don't want the expiration decrease here to occur.
            context.game.applyGameAction(gameAction, card, card => {
                card.modifyStrength(-value, false);
            });
        },
        order: value >= 0 ? 0 : 1000
    }
}
```

(you'd also need to make similar mods to the `dynamicStrength` effect if we have any cases where STR decrease is dynamic)

I don't think this approach is perfect either (or future proof) but it keeps DrawCard a little more agnostic. You could probably get rid of the `applying` parameter by shuffling things around a bit too. It also helps in that we don't have this weird underscored private version of the method.

zacchiro:
On Tue, Mar 13, 2018 at 01:57:27PM -0700, Brian Upton wrote:
> In addition to the below, we should consider whether to add checks for
> `decreaseStrength` on targeting abilities that will decrease strength
> (e.g. Dracarys) to ensure player's don't try to choose Dany as a
> target and have nothing happen. Probably not super critical when
> compared with The High Sparrow's restrictions preventing abilities,
> but worth considering.

Just a quick comment on this: yep, this is planned. I wanted to get the
immunity work first --- and I'm not there yet :-) --- before tending to
that.

zacchiro:
On Tue, Mar 13, 2018 at 01:57:27PM -0700, Brian Upton wrote:
> In general, I think it might be better to do this check from the
> effect rather than on the player. I've been trying to pull a lot of
> these types of checks (as well as raising events) off of `DrawCard`
> and `Player` in prep for #1799 and other changes. Something like the
> following may work better:

OK, here's my 2nd take on this, trying to implement your suggestion
(thanks!).

For dynamicStrength I'm not sure what to do for reapply, as there it
looks like we might lose information about whether it was a STR
reduction / increase --- unless, as I hope :), I'm wrong on this.

Tests still don't work, might be because I haven't yet found the right
mock incantation, or because this implementation is still horribly
wrong...

ystros:
This will fix your mocking issues: https://github.com/cryogen/throneteki/commit/599cddbd2ffa34674979344a7a83abf5409d8173

As for the dynamicStrength issue... yeah, I'm not sure, since they'll sometimes start at 0, at which point you don't know whether the overall effect is going to increase STR or decrease STR. Thankfully, there are no cards where it can be both - so you'll never go from +1 STR to -1 STR on the effect. So I think your current check is more or less correct.

However, I think the following in the `reapply` method:
```javascript
let value = newStrength - currentStrength;
let gameAction = value < 0 ? 'decreaseStrength' : 'increaseStrength';
context.game.applyGameAction(gameAction, card, card => {
    card.modifyStrength(value, true);
});
```

should actually be this:
```javascript
let value = newStrength - currentStrength;
let gameAction = newStrength < 0 ? 'decreaseStrength' : 'increaseStrength';
context.game.applyGameAction(gameAction, card, card => {
    card.modifyStrength(value, true);
});
```

This is because even though a STR increase may be lessened (and thus apply a negative modification to the card), it is still applied as a STR increase overall. Contrived example:
* Player has Dany and a R'hllor character out.
* Player attaches Red God's Blessing to Dany. She should get +2 STR.
* dynamicStrength applies a `increaseStrength` and calls `dany.modifyStrength(2, true)`
* The R'hllor character leaves play. The bonus from Red God's Blessing on Dany should be reduced to +1 STR
* dynamicStrength needs to reduce Dany's STR by 1. However, if it applies it as a `decreaseStrength` action Dany will be immune and not have her STR reduced. But the effect isn't a decrease in strength, it's a reduced increase in strength.
* Therefore, dynamicStrength needs to apply a `increaseStrength` action and call (somewhat counter intuitively) `dany.modifyStrength(-1, true)`.

Finally, re: Obsidian Arrows, I think you should be able to just call `allowGameAction('decreaseStrength')` on the card condition within the cost. i.e.
```javascript
cost: ability.costs.moveTokenFromSelf('gold', 1, card => this.game.currentChallenge.isDefending(card) && card.allowGameAction('decreaseStrength')),
```
but I haven't really tested it to see if there are complications.

DukeTax:
Closing this as #1846 finishes the implementation.

PR #1772, opened 2018-02-23
implement Secret Pact (TFM)

zacchiro:
French spoiler image:
https://i.redditmedia.com/i5HUPDjKuemXv1Is077rHNtrnBJ4uXRwiapcqk85axs.jpg?w=495&s=c34a879d5d2b217e296a0adf8c199398

Not to be merged yet, as the english card name is guesswork on my part.

### DIFF
--- a/server/game/cards/08.5-TFM/SecretPact.js
+++ b/server/game/cards/08.5-TFM/SecretPact.js
@@ -1,0 +1,18 @@
+const DrawCard = require('../../drawcard.js');
+
+class SecretPact extends DrawCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction(card => card.getType() === 'character' && !card.isFaction('martell'));
+        this.whileAttached({
+            effect: ability.effects.addKeyword('Renown')
+        });
+        this.whileAttached({
+            condition: () => this.game.currentChallenge && this.game.currentChallenge.defendingPlayer === this.controller,
+            effect: ability.effects.cannotBeDeclaredAsAttacker()
+        });
+    }
+}
+
+SecretPact.code = '08096';
+
+module.exports = SecretPact;


server/game/cards/08.5-TFM/SecretPact.js
```diff
@@ -0,0 +1,17 @@
+const DrawCard = require('../../drawcard.js');
+
+class SecretPact extends DrawCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction(card => card.getType() === 'character' && !card.isFaction('martell'));
+        this.whileAttached({
+            effect: [
+                ability.effects.cannotBeDeclaredAsAttacker(), // TODO need generalization for melee
```
DukeTax: This needs to be split off into a separate ```whileAttached``` with a ```condition``` property as it's only for challenges initiated against you. This card is interesting as it can be used as both a positive and a negative attachment.

server/game/cards/08.5-TFM/SecretPact.js
```diff
@@ -0,0 +1,17 @@
+const DrawCard = require('../../drawcard.js');
+
+class SecretPact extends DrawCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction(card => card.getType() === 'character' && !card.isFaction('martell'));
```
DukeTax: Needs to be unique non-Martell character I think.

zacchiro: Nope. "Uniquement" means "only" in this context.

server/game/cards/08.5-TFM/SecretPact.js
zacchiro: So, this was the point of my TODO comment. To elaborate: in a joust setting, one can only initiate challenges against its (sole) opponent (as they cannot initiate challenges against themselves), so AFAICT the implementation is *currently* correct. It will indeed break when we implement melee, hence my comment.
(I'm fine with requiring that new code should be melee-friendly, I'm commenting here just in case there is a simpler misunderstanding here.)

DukeTax: As it currently stands, attaching this to a character you control yourself, makes the character unable to be declared as an attacker in any case. Which is not correct as the parent character should only be unable to be declared as an attacker in a challenge against yourself. So attaching it to a character you control basically makes that whole drawback go away.

Splitting it off into a separate ```whileAttached``` with the following condition:

```javascript
condition: () => this.game.currentChallenge && this.game.currentChallenge.defendingPlayer === this.controller
```

should fix this for both Joust and Melee purposes.

zacchiro: You're absolutely right, my bad (I was focusing on the more complex melee case, and overlooked the drawback of being unable to attack when the attachment is used "positively"). Fixed.

zacchiro:
can be merged, card name confirmed: http://www.cardgamedb.com/index.php/agameofthrones2ndedition/a-game-of-thrones-2nd-edition-cards/_/flight-of-crows-cycle/the-faith-militan/

PR #1776, opened 2018-02-23
Implement The Crow is a Tricksy Bird

DukeTax:
https://i.redditmedia.com/3uyEYaQdp7tx9r-ekjvp2hnwcefmNxOHoTXxECTSOmo.jpg?w=508&s=2ad988b04de9e7ff0c6640c0858ca3e7

Plot action: kneel your faction card to choose an opponent and look at his plot deck. Choose one plot in that deck. This player must reveal that plot the next time he reveals a plot card, if possible.

Name seems pretty much confirmed, as there is an old CCG card with the same name: https://vignette.wikia.nocookie.net/cardguide/images/7/7a/Thecrowisatricksybird-ASoT.JPG/revision/latest/scale-to-width-down/350?cb=20170602061554

### DIFF
--- a/server/game/cards/08.6-SAT/TheCrowIsATricksyBird.js
+++ b/server/game/cards/08.6-SAT/TheCrowIsATricksyBird.js
@@ -1,0 +1,57 @@
+const DrawCard = require('../../drawcard.js');
+
+class TheCrowIsATricksyBird extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Look at plot deck',
+            phase: 'plot',
+            chooseOpponent: () => true,
+            cost: ability.costs.kneelFactionCard(),
+            handler: context => {
+                this.context = context;
+                this.game.addMessage('{0} plays {1} and kneels their faction card to look at {2}\'s plot deck',
+                    context.player, this, context.opponent);
+                
+                let buttons = context.opponent.plotDeck.map(card => {
+                    return { method: 'cardSelected', card: card };
+                });
+
+                this.game.promptWithMenu(context.player, this, {
+                    activePrompt: {
+                        menuTitle: 'Select a plot',
+                        buttons: buttons
+                    },
+                    source: this
+                });
+            }
+        });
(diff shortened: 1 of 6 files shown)


server/game/gamesteps/plot/selectplotprompt.js
```diff
@@ -25,6 +29,10 @@ class SelectPlotPrompt extends AllPlayerPrompt {
 
     onMenuCommand(player, arg) {
         if(arg === 'changeplot') {
+            if(player.mustRevealPlot) {
```
ystros: Preventing them from changing plots is good, also removing the Change Plot button entirely if they aren't allowed to change would be better UX.

server/game/player.js
```diff
@@ -616,6 +617,13 @@ class Player extends Spectator {
         this.drawPhaseCards = DrawPhaseCards;
     }
 
+    announceForcedPlotSelection() {
+        //TODO Melee: The forced plot should not be revealed to all players so needs to be whispered
```
ystros: Hmm, this is a good point. Perhaps the message shouldn't say the plot at all?

Also, I've been trying to move away from putting these types of methods onto the Player class, particularly one like this that doesn't modify the player in anyway. Probably could just be done directly in the new simple step in the plot phase.

server/game/cards/08.6-SAT/TheCrowIsATricksyBird.js
```diff
@@ -0,0 +1,57 @@
...
+            until: {
+                onCardEntersPlay: event => event.card.getType() === 'plot' && event.card.controller === this.context.opponent
+            },
+            targetType: player,
+            match: this.context.opponent,
+            effect: ability.effects.mustRevealPlot(card)
+        }));
+ 
+        //TODO Melee: The choice should not be revealed to anyone other than the chosen opponent, 
+        //so this message, as well as the announcement message in the plot phase will have to be whispered.
```
ystros: This will be fun. 😬

PR #1779, opened 2018-02-24
implement Saltcliffe Sailor (SAT)

zacchiro:
incomplete English spoiler:
https://images-cdn.fantasyflightgames.com/filer_public/07/b8/07b87d96-2edd-4869-b7f0-c87e0a9c49c0/gt28_a1_cardfan.png

full French spoiler: 
https://i.redditmedia.com/bpUBzsCefjoO3T9HwF51UImKcQIJNirxit4hBKF99LU.jpg?w=510&s=5953033e5154de82bc0231662b1af61f

### DIFF
--- a/server/game/cards/08.6-SAT/SaltcliffeSailor.js
+++ b/server/game/cards/08.6-SAT/SaltcliffeSailor.js
@@ -1,0 +1,32 @@
+const DrawCard = require('../../drawcard.js');
+
+class SaltcliffeSailor extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: card => card.hasToken('gold'),
+            effect: ability.effects.addKeyword('Stealth')
+        });
+        this.action({
+            title: 'Move 1 gold to character',
+            condition: () => this.hasToken('gold'),
+            target: {
+                type: 'select',
+                cardCondition: card =>
+                    card.location === 'play area' &&
+                    card.getType() === 'character' &&
+                    card.controller === this.controller &&
+                    !card.isFaction('greyjoy')
+            },
+            handler: context => {
+                this.modifyToken('gold', -1);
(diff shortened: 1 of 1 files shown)


server/game/cards/08.6-SAT/SaltcliffeSailor.js
```diff
@@ -0,0 +1,31 @@
...
+
+class SaltcliffeSailor extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: card => card.tokens['gold'] > 0,
+            effect: ability.effects.addKeyword('Stealth')
+        });
+        this.action({
+            title: 'Move 1 gold to character',
+            cost: ability.costs.discardGold(),
```
DukeTax: This ability has no cost, as moving the token is solely part of the resolution. As you already modified the token on this card in the handler (so currently, 2 tokens would get removed to "create" a single new token on the other card) just getting rid of the cost property will work. But then you also need to make sure the ability cannot initiate if this card has no gold tokens, which is accomplished by

```javascript
condition: () => this.hasToken('gold')
```

The hasToken() shorthand method could also be used in the match function in the persistentEffect, but that doesn't matter much.

server/game/cards/08.6-SAT/SaltcliffeSailor.js
```diff
@@ -0,0 +1,31 @@
...
+            target: {
```
DukeTax: As this ability does not explicitly target (use the word "choose"), it needs a ```type: 'select'``` property so the opponent will not know which character you intend to move the token to before deciding whether to cancel.

zacchiro:
thanks for your feedback @DukeTax , all fixed!
(including using hasToken() in the persistentEffect, which is indeed more elegant)

PR #1780, opened 2018-02-25
implement Greyjoy cards from The Faith Militant chapter pack

zacchiro:
not to be merged yet, as official english card names are still unknown

French spoiler images:
- Tarle: https://i62.servimg.com/u/f62/17/07/69/48/27858510.jpg
- Offer: https://scontent-cdg2-1.xx.fbcdn.net/v/t1.0-9/27973413_10211210705523648_4881871956079145362_n.jpg?oh=085f4c4b18b6a31ac20789980888e48b&oe=5B17B03C

PR #1781, opened 2018-02-25
restrict Night Gathers to only marshal characters

zacchiro:
should wait for #1780 to be merged as this requires the new canMarshal() effect


zacchiro:
actually, if $someone could merge this PR, I'll then rebase #1780 on this one, so that we don't have to wait indefinitely for the Night Gathers fix

cryogen:
I think I still count as $someone :)

PR #1783, opened 2018-02-27
implement The Faith Militant agenda (TFM)

zacchiro:
card spoiler: https://images-cdn.fantasyflightgames.com/filer_public/72/c3/72c3e0f0-5368-4f99-83c6-a79bf760aed4/gt27_the-faith-militant.png

### DIFF
--- a/server/game/cards/08.5-TFM/TheFaithMilitant.js
+++ b/server/game/cards/08.5-TFM/TheFaithMilitant.js
@@ -1,0 +1,43 @@
+const AgendaCard = require('../../agendacard.js');
+
+class TheFaithMilitant extends AgendaCard {
+    setupCardAbilities(ability) {
+
+        this.persistentEffect({
+            match: card =>
+                !card.hasTrait('The Seven') &&
+                ['character', 'location'].includes(card.getType()) &&
+                card.controller === this.controller,
+            effect: ability.effects.cannotGainPower()
+        });
+
+        this.reaction({
+            when: {
+                afterChallenge: event =>
+                    event.challenge.winner === this.controller &&
+                    event.challenge.loser.faction.power >= 1
+            },
+            cost: ability.costs.kneelFactionCard(),
+            target: {
+                cardCondition: card =>
+                    card.location === 'play area' &&
+                    card.controller === this.controller &&
+                    card.getType() === 'character' &&
+                    card.hasTrait('The Seven') &&
+                    this.game.currentChallenge.isParticipating(card)
+            },
+            handler: context => {
+                let otherPlayer = context.event.challenge.loser;
+                this.game.movePower(otherPlayer.faction, context.target, 1);
+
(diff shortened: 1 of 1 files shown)


server/game/cards/08.5-TFM/TheFaithMilitant.js
```diff
@@ -0,0 +1,43 @@
+const AgendaCard = require('../../agendacard.js');
+
+class TheFaithMilitant extends AgendaCard {
+    setupCardAbilities(ability) {
+
+        this.persistentEffect({
+            match: card =>
+                !card.hasTrait('The Seven') &&
+                ['character', 'location'].includes(card.getType()) &&
+                card.controller === this.controller,
```
DukeTax: Not an issue, but also not needed as persistentEffects default to the current controller.

server/game/cards/08.5-TFM/TheFaithMilitant.js
```diff
@@ -0,0 +1,43 @@
...
+            target: {
+                cardCondition: card =>
+                    card.location === 'play area' &&
+                    card.controller === this.controller &&
+                    card.getType() === 'character' &&
+                    card.hasTrait('The Seven') &&
+                    this.game.currentChallenge.isParticipating(card)
+            },
+            handler: context => {
+                var otherPlayer = context.event.challenge.loser;
```
DukeTax: var --> let

server/game/cards/08.5-TFM/TheFaithMilitant.js
```diff
@@ -0,0 +1,43 @@
...
+                this.game.movePower(otherPlayer.faction, context.target, 1);
+
+                this.game.addMessage('{0} uses {1} to move 1 power from {2}\'s faction to {3}',
```
DukeTax: As we include costs in the message, it would be "{0} uses {1} and kneels their faction card to...".

server/game/cards/08.5-TFM/TheFaithMilitant.js
zacchiro: Thanks for the review. All comments addressed except this one. I was aware of the default controller, but I quite like when there is a 1-to-1 mapping between card text ("you control") and card implementation (the "===" requirement). This probably deserves a more broad discussion about whether we prefer to have shorter card implementations or more explicit ones.

server/game/cards/08.5-TFM/TheFaithMilitant.js
zacchiro: as this was coming from copy/past from another card, I've reviewed "var" use in card implementation and filed #1785

server/game/cards/08.5-TFM/TheFaithMilitant.js
DukeTax: 👍

Issue #1787, opened 2018-03-01
It's possible to play the bounty of HIghgarden during draw phase and another limited during marshalling.


Grugn0:
If you have enogh money during draw/plot phase you can play the bounty of highgarden during these phases and then marshal another limited card during the marshalling phase.

![image](https://user-images.githubusercontent.com/8015654/36840631-fad145ba-1d45-11e8-8630-3b3c434a318f.png)

By default each player can play or marshal only one card with "limited" keyword each round.

Maybe this issue depends on the fact that the count (for each player) is reset at the beginning of the marshalling (and not at the begininng of each round as it should be).

https://github.com/cryogen/throneteki/blob/27157a1f57e87fc5b5fd66e3b83a355747e605f9/server/game/player.js#L657

zacchiro:
Looks like it, yes! Would you be willing to test this more? If the hypothesis is correct, it shouldn't be possible to play Bounty during draw phase if you played a limited during the *previous* turn. And if it is, would you be willing to submit a PR fixing this? That'd be awesome :-) Thank you!

Grugn0:
I've tried this second scenario and seems that the hypotesis is correct.
Now I try to make a PR (I'm not very practical about git, I've always used svn / cvs but I can do it)

ystros:
Fixed by #1788. Thanks again!

PR #1788, opened 2018-03-01
Reset the number of limited cards played by each player at the start of each round

Grugn0:
Reset the number of limited cards played by each player at the start of each round (Plot phase) and not at the start of the marshal phase.


Grugn0:
#1787

ystros:
Good catch! 👍

PR #1793, opened 2018-03-03
implement The High Sparrow (TFM)

zacchiro:


ystros:
As a suggestion on how to go about fixing the overwriting problem, take a look at `MinMaxProperty` and how it's being used on `Player` and w/ effects. May require something separate but similar, may be able to just use it outright - a bit weird since there's no real notion of "minimum amount of gold / cards drawn" but workable.

zacchiro:
In addition to @ystros' remarks, this PR also still lacks adding checks (calling `canDraw()`) on whether one can draw cards or not as a condition to trigger abilities. @DukeTax you had some cards in mind, please feel free to name them here! I'll do a more general pass to find instances whether that is needed, but specific examples would be helpful nonetheless :)

DukeTax:
For draw: Regular draw phase draw, Insight, Secret Schemes, Pleasure Barge, Red Keep, The Mander.
For gold: Regular marshalling phase gold collection, Tyrion Lannister (Core), In Dorans Name, Plunder, Bear Island, Tower of the Sun.

zacchiro:
@DukeTax pre-emptive checks added in all those places
@ystros switched to use MinMaxProperty for both card and draw limits

hence i've just removed the Needs Work label, but (aside from npm test) I haven't yet been able to test this. Right, I know, what could possibly go wrong?

zacchiro:
OK, i've now also added canDraw() and canGainGold() checks to all card that allow to do either or both of those things. The only missing ones are those where we prompt players to choose between things (e.g., the reader); ideally we should grey out the option to do something that is not currently allowed, but that'd need client support.
This is now as good as I could make it without further review. (And yes, it's also becoming unwieldy…)

server/game/insightkeyword.js
```diff
@@ -16,9 +16,11 @@ class InsightKeyword extends BaseAbility {
 
     executeHandler(context) {
         let {game, challenge, source} = context;
-        let drawn = challenge.winner.drawCardsToHand(1);
-        game.raiseEvent('onInsight', { challenge: challenge, source: source, card: drawn });
-        game.addMessage('{0} draws a card from Insight on {1}', challenge.winner, source);
+        if(challenge.winner.canDraw()) {
```
ystros: This is probably nitpicky, but the `canDraw` check should really go in `meetsRequirements`. If there's ever a weird card that allows cancellation of Insight (e.g. Fear Cuts vs Stealth or Intimidate), it shouldn't fire unless it's actually possible for Insight to initiate.

server/game/player.js
```diff
@@ -200,11 +203,26 @@ class Player extends Spectator {
...
 
     modifyGold(amount) {
-        this.gold += amount;
+        var goldGain = amount;
 
-        if(this.gold < 0) {
-            this.gold = 0;
+        if(amount > 0) {
+            if(!this.canGainGold()) {
+                this.addMessage('{0} cannot gain gold', this);
```
ystros: After the gold encapsulation PR (sorry), this is no longer an appropriate place to check for `canGainGold`. 'Gaining' gold (collecting income during marshalling, Tyrion's ability, etc) is different than 'moving' gold (taking gold off of Lannisport Treasury, stealing gold with Meager Contribution, etc). Both `addGold` and `transferGold` use `modifyGold` in their underlying implementations, so this change will also inappropriately prevent moving gold.

Thus, most of this code should be moved to `Game.addGold` instead.

server/game/insightkeyword.js
zacchiro: Done.

server/game/player.js
zacchiro: Ah, I was out of luck then, I moved it to the wrong place when rebasing on top of your gold encapsulation, rather than leaving where it already was :-) Moved back now.

zacchiro:
I've retested everything, and it seems to work.
There's one remaining issue: `drawCardsToHand(1)` returns a single element whereas, say, `drawCardsToHand(3)` returns a list, because behind the scene `.first()` is used. This means we cannot consistently use `.length` in code like the following (e.g., from Funeral Pyre):
```
let cards = this.controller.drawCardsToHand(3).length;
this.game.addMessage('{0} uses {1} to draw {2} {3}',
    this.controller, this, cards, cards > 1 ? 'cards' : 'card');
```
I'm not sure what's the best way to go about this. Can I make `drawCardsToHand` consistently return a list, even if it's for a single card?

zacchiro:
> There's one remaining issue: drawCardsToHand(1) returns a single element whereas, say, drawCardsToHand(3) returns a list, because behind the scene .first() is used. This means we cannot consistently use .length in code like the following (e.g., from Funeral Pyre):

Thanks to @ystros feedback in chat this is now fixed: `drawCardsToHand` now always returns an array. I've also simplified `addGold` relying on the hypothesis that it will never be passed negative amounts.

PR #1794, opened 2018-03-03
implement "The Song of the Seven" (TFM)

zacchiro:


server/game/cards/08.5-TFM/TheSongOfTheSeven.js
```diff
@@ -0,0 +1,38 @@
+const DrawCard = require('../../drawcard.js');
+
+class TheSongOfTheSeven extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            max: ability.limit.perRound(1),
+            when: {
+                onDominanceDetermined: event =>
+                    this.controller !== event.winner &&
+                    event.winner.faction.power >= 2 &&
```
DukeTax: If event.winner is undefined (in case of a dominance tie), this will likely crash, so you'll need to check for that first. Also, this card should still trigger if the opponent has a single power on their house card.

server/game/cards/08.5-TFM/TheSongOfTheSeven.js
```diff
@@ -0,0 +1,38 @@
...
+                    this.hasTheSeven()
```
DukeTax: This is unneeded as the target API already makes sure the card won't trigger if there's no legal target.

server/game/cards/08.5-TFM/TheSongOfTheSeven.js
```diff
@@ -0,0 +1,38 @@
...
+            target: {
+                cardCondition: card =>
+                    card.location === 'play area' &&
+                    card.controller === this.controller &&
+                    card.getType() === 'character' &&
+                    card.hasTrait('The Seven')
+            },
+            handler: context => {
+                let otherPlayer = context.event.winner;
+                let power = 2;
```
DukeTax: This needs to be 1 if ```otherPlayer.faction.power === 1``` and 2 if more than that. movePower guards against not being able to move more power than actually possible, but setting this variable incorrectly would mess up the message.

server/game/cards/08.5-TFM/TheSongOfTheSeven.js
```diff
@@ -0,0 +1,38 @@
...
+                this.game.movePower(otherPlayer.faction, context.target, power);
+
+                this.game.addMessage('{0} uses {1} to move {2} power from {3}\'s faction to {4}',
```
DukeTax: uses --> plays

server/game/cards/08.5-TFM/TheSongOfTheSeven.js
zacchiro: I guess eventually, if you keep repeating this to me, I'll learn about this, but no promises :-)
(all comments addressed)

PR #1800, opened 2018-03-06
Implement The Night's Watch cards from Sands of Dorne

DukeTax:
Spare Boot: https://www.wardensofthemidwest.com/wp-content/uploads/2018/03/GT30_Cards_COMPOSITE-32-211x300.png

Turncloak: https://www.wardensofthemidwest.com/wp-content/uploads/2018/03/GT30_Cards_COMPOSITE-33-211x300.png

### DIFF
--- a/server/game/cards/10-SoD/SpareBoot.js
+++ b/server/game/cards/10-SoD/SpareBoot.js
@@ -1,0 +1,33 @@
+const DrawCard = require('../../drawcard.js');
+
+class SpareBoot extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Stand and move attachment',
+            phase: 'challenge',
+            limit: ability.limit.perPhase(1),
+            target: {
+                activePromptTitle: 'Select an attachment',
+                cardCondition: card => card.location === 'play area' && card.getType() === 'attachment' &&
+                                       card.parent && card.parent.controller === this.controller &&
+                                       (card.parent !== this || card.kneeled) &&
+                                       card.controller.canAttach(card, this)
+            },
+            handler: context => {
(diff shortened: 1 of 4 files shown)


DukeTax:
I took the card codes from the corresponding entries for NW cards in House of Thorns, but that could be wrong.

server/game/cards/10-SoD/SpareBoot.js
```diff
@@ -0,0 +1,33 @@
...
+        this.action({
+            title: 'Stand and move attachment',
+            phase: 'challenge',
+            limit: ability.limit.perPhase(1),
+            target: {
+                activePromptTitle: 'Select an attachment',
+                cardCondition: card => card.location === 'play area' && card.getType() === 'attachment' &&
+                                       card.parent && card.parent.controller === this.controller &&
+                                       (card.parent !== this || card.kneeled) &&
+                                       this.controller.canAttach(card, this)
```
ystros: This needs to check `card.controller.canAttach(card, this)` instead. Otherwise I believe moving an "opponent only" attachment will fail because it will use the current player as the potential controller and expect it to be attached to a player other than the controller of Spare Boot.

server/game/cards/10-SoD/SpareBoot.js
```diff
@@ -0,0 +1,33 @@
...
+            },
+            handler: context => {
+                if(context.target.parent !== this) {
+                    context.player.attach(context.player, context.target, this);
```
ystros: Similarly, the first parameter of the method here is the new controller for the attachment. This will thus take control of the attachment, which isn't correct. Should be `context.player.attach(context.target.controller, ...)` instead.

PR #1802, opened 2018-03-06
Implement 'Each Player' targeting mode

DukeTax:
- Adds a target mode that satisfies a target for each player, making such cards compatible with upcoming melee functionality and single-player.
- Instead of sequentially prompting for each separate player, all targets are selected in a single prompt.

### DIFF
--- a/server/game/CardSelectors/BaseCardSelector.js
+++ b/server/game/CardSelectors/BaseCardSelector.js
@@ -32,7 +32,7 @@ class BaseCardSelector {
         return context.game.allCards.filter(card => this.canTarget(card, context));
     }
 
-    hasEnoughSelected(selectedCards) {
+    hasEnoughSelected(selectedCards, numPlayers) { // eslint-disable-line no-unused-vars
         return selectedCards.length > 0;
     }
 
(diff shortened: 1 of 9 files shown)


ystros:
💯

server/game/CardSelectors/BaseCardSelector.js
```diff
@@ -32,7 +32,7 @@ class BaseCardSelector {
         return context.game.allCards.filter(card => this.canTarget(card, context));
     }
 
-    hasEnoughSelected(selectedCards) {
+    hasEnoughSelected(selectedCards, numPlayers) { // eslint-disable-line no-unused-vars
```
cryogen: Not super keen on these placeholder variables that do nothing

ystros: I think it's appropriate for base classes (which this is) to better document parameters that while not used here can be used in inheriting classes. Could be removed, but then should be documented in some other way.

cryogen: Probably makes more sense to use comments instead.  What we really need is abstract methods, but those don't exist.  Ideally we should also be "throw"'ing in base methods to make it more obvious that that's what this is

PR #1805, opened 2018-03-07
Use ability contexts directly for immunity

ystros:
Previously, immunity / cannot restrictions looked at an object that
contained the card and the current stage of resolution for the ability.
This was referred to as a "ability context" which overlapped with the
name given to the context that tracks the player using the ability and
cost / target resolution.

To reduce confusion and gain access to this additional information, the
actual AbilityContext object is tracked directly and used within the
immunity system.

### DIFF
--- a/server/game/AbilityContext.js
+++ b/server/game/AbilityContext.js
@@ -9,6 +9,7 @@ class AbilityContext {
         this.costs = {};
         this.costValues = {};
         this.targets = new ResolvedTargets();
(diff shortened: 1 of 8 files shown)


ystros:
This is a simplification Ringteki made a long time ago, and I'm finding that it would be very convenient if I were able to access `context.player` for Iron Bank (in order to tell whether a call to `transferGold` is a player "spending" their own gold or if an opponent is stealing the gold)

PR #1806, opened 2018-03-08
Use ability contexts directly for immunity

ystros:
Previously, immunity / cannot restrictions looked at an object that
contained the card and the current stage of resolution for the ability.
This was referred to as a "ability context" which overlapped with the
name given to the context that tracks the player using the ability and
cost / target resolution.

To reduce confusion and gain access to this additional information, the
actual AbilityContext object is tracked directly and used within the
immunity system.

### DIFF
--- a/server/game/baseability.js
+++ b/server/game/baseability.js
@@ -60,7 +60,33 @@ class BaseAbility {
      * @returns {Boolean}
      */
     canPayCosts(context) {
-        return _.all(this.cost, cost => cost.canPay(context));
+        return this.executeWithTemporaryContext(context, 'cost', () => _.all(this.cost, cost => cost.canPay(context)));
+    }
+
+    /**
+     * Executes the specified callback using the passed ability context and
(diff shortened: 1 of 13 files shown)


server/game/baseability.js
```diff
@@ -60,7 +60,20 @@ class BaseAbility {
      * @returns {Boolean}
      */
     canPayCosts(context) {
-        return _.all(this.cost, cost => cost.canPay(context));
+        return this.withContext(context, 'cost', () => _.all(this.cost, cost => cost.canPay(context)));
+    }
+
+    withContext(context, stage, callback) {
```
cryogen: What 'withContext'?  (Naming Is Hard, I know).  This method probably needs to state what it's doing with the context (executeWithContext, resolveWithContext etc)

ystros: Renamed.

cryogen: Thanks, much clearer!

PR #1807, opened 2018-03-08
Better encapsulate gold usages

ystros:
Introduces methods for spending gold and returning gold to treasury, as well as updates `transferGold` to have a less confusing interface and handle the following scenarios:
* player => player
* player => card
* card => player
* card => card

### DIFF
--- a/server/game/cards/02.4-NMG/RedCloaks.js
+++ b/server/game/cards/02.4-NMG/RedCloaks.js
@@ -5,11 +5,9 @@ class RedCloaks extends DrawCard {
         this.action({
             title: 'Move 1 gold from your gold pool to this card',
             limit: ability.limit.perPhase(1),
-            condition: () => this.controller.gold > 0,
+            condition: () => this.controller.hasEnoughGold(1),
             handler: () => {
-                this.modifyToken('gold', 1);
-                this.game.addGold(this.controller, -1);
-
+                this.game.transferGold({ from: this.controller, to: this, amount: 1 });
                 this.game.addMessage('{0} moves 1 gold from their gold pool to {1}', this.controller, this);
             }
         });
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -399,11 +399,7 @@ class Game extends EventEmitter {
             return;
         }
 
-        player.gold += gold;
-
-        if(player.gold < 0) {
-            player.gold = 0;
-        }
+        player.modifyGold(gold);
     }
 
     movePower(fromCard, toCard, power) {
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -419,6 +419,28 @@ class Game extends EventEmitter {
     }
 
     /**
+     * Spends a specified amount of gold for a player. "Spend" refers to any
+     * usage of gold that returns gold to the treasury as part of an ability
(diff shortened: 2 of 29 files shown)


server/game/cards/02.4-NMG/RedCloaks.js
```diff
@@ -5,11 +5,9 @@ class RedCloaks extends DrawCard {
         this.action({
             title: 'Move 1 gold from your gold pool to this card',
             limit: ability.limit.perPhase(1),
-            condition: () => this.controller.gold > 0,
+            condition: () => this.controller.hasEnoughGold(1),
```
cryogen: Do we have making actions support costs on the todo list anywhere?

server/game/game.js
```diff
@@ -422,13 +418,70 @@ class Game extends EventEmitter {
...
+     * @param {(BaseCard|Player)} transferParams.from
+     * The source object from which gold is being moved
+     * @param {(BaseCard|Player)} transferParams.to
+     * The target object to which gold is being moved
+     */
+    transferGold(transferParams) {
+        let {from, to, amount} = transferParams;
+        let appliedGold = Math.min(from.gold, amount);
+
+        if(from.getGameElementType() === 'player') {
```
cryogen: Any reason we're not using 'real' type checking here?

server/game/player.js
```diff
@@ -195,6 +195,18 @@ class Player extends Spectator {
         return this.plotDiscard.size() + this.usedPlotsModifier;
     }
 
+    hasEnoughGold(gold) {
```
cryogen: This seems unnecessarily verbose, was 'if(foo >= bar)' really that hard to understand?

server/game/cards/02.4-NMG/RedCloaks.js
DukeTax: They do, but Red Cloaks has no cost associated with its ability. Gold needs to be checked to make sure triggering the ability has the potential to change the game state.

server/game/player.js
ystros: It's not that it's hard to understand, it's the fact that `player.gold >= gold` becomes incorrect as soon as we introduce Iron Bank. This was an intermediate step. If you look at the Iron Bank PR, this method has a more complicated implementation that checks multiple gold sources.

Now, there may be merit in doing this:
```javascript
if(player.availableGold() >= gold)

// vs

if(player.hasEnoughGold(gold))
```

but it needs to be a method in both cases in order to take the optional parameters sent in the Iron Bank PR:

```javascript
if(player.availableGold({ playingType: 'marshal' }) >= gold)

// vs

if(player.hasEnoughGold(gold, { playingType: 'marshal' })
```

(after writing this I'm sort of leaning more toward the hypothetical `availableGold` above, but let me know your thoughts)

server/game/game.js
ystros: We could here, game element type is mostly used for the effects engine where there are arrays that contain both card and player objects. It's also useful for writing tests (can send a mock instead of a real player object), but we don't have any tests around this currently. Your call. 🤷‍♀️

cryogen: javascript is annoying.

PR #1808, opened 2018-03-08
Fully implement Iron Bank and Sweet Cersei

### LINKED ISSUE #1628: implement The Iron Bank
The Iron Bank from Archmaester's Key isn't implemented and it should be.

As per previous discussion, the most reasonable UI approach to implement that card seems to be having a checkbox element on the card, defaulting to unchecked. Users can check it and, while it is checked, gold on The Iron Bank is used *first* to satisfy gold needs in the marshaling phase.

A bare bone implementation of this might be no visible clue on the card, but a menu option when clicking on the card that toggles using/not using it. (Simpler to implement, but worse than the previous option in terms of UX.)

ystros:
~**DO NOT MERGE until #1806 and #1807 are merged**~

For review purposes until the other 2 PRs are merged, you can look at https://github.com/ystros/throneteki/compare/b630c5f...ystros:iron-bank-final?expand=1

This works by explicitly prompting the user when they want to spend gold during the eligible phase. If there aren't multiple ways to combine the gold in their pool and the gold on the card, then the gold will be spent automatically without prompting. Examples:

* 4 gold in pool, 1 on Iron Bank. Marshal a 5 cost character. All gold spent automatically, no prompting.
* 1 gold in pool, 4 on Iron Bank. Marshal a 5 cost character. All gold spent automatically, no prompting.
* 4 gold in pool, 2 on Iron Bank. Marshal a 5 cost character. Player is prompted whether they want to spend 1 or 2 gold from the Iron Bank. The remainder is spent from their pool.
* 5 gold in pool, 5 on Iron Bank. Marshal a 5 cost character. Player is prompted whether they want to spend 0 to 5 gold from the Iron Bank. The remainder (if any) is spent from their pool.
* 0 gold in pool, 1 gold on Iron Bank. Play a 1 cost event. Gold spent automatically from Iron Bank, no prompting.

I'm sure there are still bugs (anywhere checking player.gold directly is suspect), and there's probably some quality of life things we can add (like saying "I don't want to spend from Iron Bank this round" to reduce prompting while you let gold build up), but this should get us 99% there.

Fixes #1628

### DIFF
--- a/server/game/GoldSource.js
+++ b/server/game/GoldSource.js
@@ -1,0 +1,28 @@
+class GoldSource {
+    constructor(playerOrCard, allowSpendingFunc = () => true) {
+        this.playerOrCard = playerOrCard;
+        this.allowSpendingFunc = allowSpendingFunc;
+    }
+
+    get gold() {
+        return this.playerOrCard.gold;
+    }
+
+    get name() {
+        if(this.playerOrCard.getGameElementType() === 'player') {
+            return `${this.playerOrCard.name}'s gold pool`;
+        }
+
+        return this.playerOrCard.name;
+    }
+
+    allowSpendingFor(spendParams) {
+        return this.allowSpendingFunc(spendParams);
+    }
+
+    modifyGold(amount) {
+        this.playerOrCard.modifyGold(amount);
+    }
+}
+
+module.exports = GoldSource;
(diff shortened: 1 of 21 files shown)


cryogen:
Probably easier to review this once the other two are in

cryogen:
This conflicts now because of the changes i asked for - rebasing should fix it though

PR #1811, opened 2018-03-09
Allow select card prompts to reveal facedown targets & Implement False Spring

DukeTax:
- Adds a flag to select card prompts that reveals selectable cards that are otherwise facedown, for example cards in an opponent's hand.
- Allows cards like Seen in Flames and His Viper Eyes to use a select card prompt, instead of menu buttons.
- This mirrors earlier changes that converted cards targeting the faction card to switch from menu buttons to target prompts.

### DIFF
--- a/server/game/drawcard.js
+++ b/server/game/drawcard.js
@@ -507,6 +507,7 @@ class DrawCard extends BaseCard {
             inDanger: this.inDanger,
             kneeled: this.kneeled,
             power: this.power,
+            revealWhenHiddenTo: this.revealWhenHiddenTo,
             saved: this.saved,
             strength: this.getStrength(),
             stealth: this.stealth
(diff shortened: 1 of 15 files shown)


server/game/drawcard.js
```diff
@@ -458,7 +459,19 @@ class DrawCard extends BaseCard {
         this.saved = false;
     }
 
+    showFacedownTarget() {
+        this.revealWhenHidden = true;
```
ystros: This approach will not work with Melee because it's all or nothing - it will show the card to all opponents, not just the one currently viewing cards. That works fine with False Spring, which explicitly reveals the cards, but not with His Viper's Eyes or Seen in Flames, as those are explicitly "look at" opponent's hand (only current player can see) not "reveal" opponent's hand (everyone can see)

DukeTax:
Rebased to instead record the player's name the card is revealed to on the card(s) that are being revealed.

ystros:
I think ultimately we need to rework how card visibility is handled in general (make it more centralized, probably move the individual cards off of the Player objects at the same time), but this should work well for now. 👍

PR #1814, opened 2018-03-10
implement Sacrifice to the Red God (TFM)

zacchiro:
French spoiler: https://i.redditmedia.com/0uqmMRzViESffdRtQi59pPVqXFJde4cd88eXUA_VObA.jpg?w=495&s=2bfddfedf849902ed3ec0fa4aea3cf8e

### DIFF
--- a/server/game/cards/08.5-TFM/SacrificeToTheRedGod.js
+++ b/server/game/cards/08.5-TFM/SacrificeToTheRedGod.js
@@ -1,0 +1,39 @@
+const DrawCard = require('../../drawcard.js');
+
+class SacrificeToTheRedGod extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Search for a character',
+            phase: 'marshal',
+            cost: ability.costs.sacrifice(card => card.getType() === 'character'),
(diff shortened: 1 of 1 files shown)


server/game/cards/08.5-TFM/SacrifieAuDieuRouge.js
```diff
@@ -0,0 +1,39 @@
...
+                    onCancel: player => this.doneSelecting(player),
+                    source: this
+                });
+            }
+        });
+    }
+
+    cardSelected(player, card) {
+        player.moveCard(card, 'hand');
+        this.game.addMessage('{0} uses {1} to search their deck and add {2} to their hand',
```
DukeTax: Very minor, but these messages would usually be phrased as '{Player} plays {event} and sacrifices {card} to...' (same for the message in the cancel method). Looks good otherwise.

zacchiro: Thanks, you're right, fixed.
(Looks like we have a bunch of other deck search events that uses the bad wording, probably inherited from non-event tutoring cards…)

zacchiro:
this can now be merged, the official english card name has been spoiled on the Second Sons podcast and I've just updated the PR accordingly

PR #1818, opened 2018-03-11
implement House Yronwood Knight (TFM)

zacchiro:
french spoiler: https://www.instagram.com/p/BfQd_yCBTn0/

### DIFF
--- a/server/game/cards/08.5-TFM/HouseYronwoodKnight.js
+++ b/server/game/cards/08.5-TFM/HouseYronwoodKnight.js
@@ -1,0 +1,22 @@
+const DrawCard = require('../../drawcard.js');
+
+class HouseYronwoodKnight extends DrawCard {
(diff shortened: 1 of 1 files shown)


zacchiro:
the name of the card should be ok, but we don't have the card code yet

DukeTax:
The code is pretty much guaranteed to be 08095, based on previous packs, but the name could be 'Knight of House Yronwood' as well. :)

zacchiro:
On Sun, Mar 11, 2018 at 04:58:33PM +0000, DukeTax wrote:
> The code is pretty much guaranteed to be 08095, based on previous
> packs, but the name could be 'Knight of House Yronwood' as well. :)

I've added the code. I think it's unlikely that it will be called
"knight of", as all other house knights are "House Bla Knight". But it's
not a big deal to wait for merging it :-)

PR #1822, opened 2018-03-14
Use throneteki-json-data for card data

ystros:
* Uses the new `throneteki-json-data` data format for card data instead of `thronesdb-json-data`. This applies to server-side in-game data, client-side deck builder data, and the data used for running tests.
* Since the data formats are slightly different, you can no longer import data from ThronesDB when fetching data. It would have to be converted and there's already converting / syncing functions built into the `throneteki-json-data` repo.
* We still need to run the `fetchdata` script, as we still store packs + cards in Mongo. There is some slight denormalization done to the cards when this is done (adds `packCode` in order to check release dates as well as `label` (name + pack code for cards with multiple versions)). Future updates may be able to remove Mongo storage, but there are things that need to be thought through (e.g. adding new cards to the deck builder without also having to restart the nodes).
* I haven't thought through all the implications but I imagine this will be tricky to deploy. Nodes will need to be restarted to have the code running w/ the new data format, but the lobby will also need to be restarted because it sends the data to the game nodes and we want the formats to be the same.


.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "thronesdb-json-data"]
-	path = thronesdb-json-data
-	url = https://github.com/Alsciende/thronesdb-json-data.git
+[submodule "throneteki-json-data"]
+	path = throneteki-json-data
+	url = https://github.com/ystros/throneteki-json-data
```
ystros: If we want to move the repo to a `throneteki` organization, we can. I'm not sure how much sense that makes unless 1. we're moving the main repo over there as well, 2. we anticipated more repos in the future - 2 repos seems kind of inconsequential for moving things around, 3. if we ever continue on generalizing the engine, it probably doesn't make sense for those repos to be under the name "throneteki".

client/DeckSummary.jsx
```diff
@@ -49,7 +49,8 @@ class DeckSummary extends React.Component {
         let combinedCards = _.union(this.props.deck.plotCards, this.props.deck.drawCards);
 
         _.each(combinedCards, (card) => {
-            let type = card.card.type_name;
+            let typeCode = card.card.type;
+            let type = typeCode[0].toUpperCase() + typeCode.slice(1);
```
cryogen: I hate javascript sometimes, what does this do?

server/game/basecard.js
```diff
@@ -64,12 +65,14 @@ class BaseCard {
 
         this.abilities = { actions: [], reactions: [], persistentEffects: [], playActions: [] };
         this.parseKeywords(cardData.text || '');
-        this.parseTraits(cardData.traits || '');
+        for(let trait of cardData.traits || []) {
```
cryogen: this one change alone makes the whole thing worthwhile ❤️

server/scripts/fetchdata/JsonCardSource.js
```diff
@@ -3,54 +3,54 @@ const fs = require('fs');
 const _ = require('underscore');
 
 class JsonCardSource {
-    getCards() {
-        let files = fs.readdirSync('thronesdb-json-data/pack');
-        let totalCards = [];
-        let types = JSON.parse(fs.readFileSync('thronesdb-json-data/types.json'));
-        let factions = JSON.parse(fs.readFileSync('thronesdb-json-data/factions.json'));
+    constructor() {
```
ystros: Now that the JSON files are the only source for card data, there's an implication that the submodule must be updated every time we want to release new card implementations. Previously, it only needed to be updated if we were writing tests for a new card (since the tests do not use Mongo and need a deterministic data source).

If this is undesirable or we want more options, I could update this to take a directory path so that you can specify a version of the data repo other than what's in the submodule directory. That would probably be more flexible but also more error prone (need to remember to pull in the separate repo then run the script vs just pulling latest source here and running the script)

.gitmodules
cryogen: I'm happy keeping it wherever makes more sense, if there were to be a new organisation I'd move throneteki there as well, but it being where it is, is not a problem for me

server/scripts/fetchdata/JsonCardSource.js
cryogen: That's fine, One Day(tm) there will be a build script that takes care of all that. In the mean time, I'll just put it on my list of things to remember

client/DeckSummary.jsx
ystros: Capitalizes the first letter of the type. "character" => "Character", "location" => "Location", etc. I could introduce a method (either extend the prototype of String, or add a specific method for converting type to its name) but it seemed overkill for just 2 places.

.gitmodules
ystros: Since neither of us have strong opinions about it, it'll stay where it is for now. I'll make everyone collaborators on the repo.

PR #1824, opened 2018-03-16
Refactor navigation

cryogen:


client/routes.js
```diff
@@ -0,0 +1,45 @@
...
+    { path: '/', action: () => <Lobby /> },
+    { path: '/login', action: () => <Login /> },
+    { path: '/logout', action: () => <Logout />},
+    { path: '/register', action: () => <Register /> },
+    { path: '/decks', action: () => <Decks /> },
+    { path: '/decks/add', action: () => <AddDeck /> },
+    { path: '/decks/edit/:id([a-f\\d]{24})', action: context => <EditDeck deckId={ context.params.id }/> },
+    { path: '/play', action: context => (context.currentGame && context.currentGame.started) ? <GameBoard /> : <GameLobby />},
+    { path: '/how-to-play', action: () => <HowToPlay /> },
+    { path: '/about', action: () => <About /> },
```
ystros: Might be nice to sort these in alphabetical order.

client/NavBar.jsx
```diff
@@ -28,40 +28,71 @@ class InnerNavBar extends React.Component {
     }
 
     renderMenuItem(menuItem) {
+        let active = menuItem.path === this.props.path ? 'active' : '';
+        let title = this.props.user ? menuItem.title.replace('%CURRENTUSER%', this.props.user.username) : menuItem.title;
```
ystros: This replacement seems slightly strange since we're also already special casing the only time it occurs (`showProfilePic`). It's fine, but it introduces 2 different independent keying conditions that are unlikely to ever be used independently.

client/Router.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import toRegex from 'path-to-regexp';
```
ystros: Why am I surprised this is a package that exists.

cryogen: Surprisingly useful :)

PR #1828, opened 2018-03-17
Implement remaining Martell characters from Sands of Dorne

DukeTax:
Implements all remaining Martell characters from Sands of Dorne, with the exception of Obella Sand.

https://imgur.com/a/pjdvw

### DIFF
--- a/server/game/cards/10-SoD/MaesterKedry.js
+++ b/server/game/cards/10-SoD/MaesterKedry.js
@@ -1,0 +1,24 @@
+const DrawCard = require('../../drawcard.js');
+
+class MaesterKedry extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Reduce next event by 2',
+            cost: ability.costs.kneelSelf(),
+            handler: context => {
+                this.untilEndOfPhase(ability => ({
+                    targetType: 'player',
+                    targetController: 'current',
+                    effect: ability.effects.reduceNextPlayedCardCost(2, card => card.getType() === 'event')
+                }));
+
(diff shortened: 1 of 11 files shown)


server/game/cards/10-SoD/MaesterKedry.js
```diff
@@ -0,0 +1,24 @@
...
+class MaesterKedry extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Reduce next event by 2',
+            cost: ability.costs.kneelSelf(),
+            handler: context => {
+                this.untilEndOfPhase(ability => ({
+                    targetType: 'player',
+                    targetController: 'current',
+                    effect: ability.effects.reduceNextPlayedCardCost(2, card => card.getType() === 'event')
```
ystros: Checking the card type should (theoretically) be redundant since only events can be 'played'. Is there a situation I'm not thinking of or a general bug in the reducer system?

server/game/cards/10-SoD/NymeriasGuard.js
```diff
@@ -0,0 +1,25 @@
+const DrawCard = require('../../drawcard.js');
+
+class NymeriasGuard extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Put Nymeria\'s Guard into play',
+            location: 'hand',
+            phase: 'challenges',
```
ystros: 'challenge' not 'challenges' (I ought to put some validation on this)

server/game/cards/10-SoD/MaesterKedry.js
DukeTax: Not that I know of, should be good.

PR #1837, opened 2018-03-20
Extract components for Decks page

ystros:


client/ViewDeck.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import ConfirmedButton from './FormComponents/ConfirmedButton';
+import DeckSummary from './DeckSummary';
+import Panel from './SiteComponents/Panel';
+
+class ViewDeck extends React.Component {
```
ystros: I'm fairly ambivalent on this component. It does help separate another large chunk out of Decks rendering wise. It also ensures that each component deals with things at the same level of abstraction - Decks only deals with events on decks (selecting, editing, deleting) whereas ViewDeck deals with the lower level UI concerns (clicking on the edit and delete button and translating them to the proper deck-oriented events).

It's a subtle distinction that I think makes the overall code cleaner and more focused - but it's also a bunch of boilerplate code. So overall I think it's a win but not sure if it's such a slam-dunk win to justify itself. 🤷‍♀️

client/DeckList.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import DeckRow from './DeckRow';
+
+class DeckList extends React.Component {
+    render() {
+        let { activeDeck, className, decks, onSelectDeck } = this.props;
```
cryogen: Argh, the deconstruction, it burns! 🙊

client/DeckRow.jsx
```diff
@@ -3,9 +3,19 @@ import PropTypes from 'prop-types';
...
 
 class DeckRow extends React.Component {
+    constructor() {
+        super();
+
+        this.handleDeckClick = this.handleDeckClick.bind(this);
+    }
+
+    handleDeckClick() {
+        this.props.onSelect(this.props.deck);
```
cryogen: Probably sensible to check this exists, especially since the prop isn’t marked as required (even if it was it would only log a warning)

client/Decks.jsx
```diff
@@ -110,10 +57,12 @@ class InnerDecks extends React.Component {
                     <div className='col-sm-5 full-height'>
                         <Panel title='Your decks'>
                             <Link className='btn btn-primary' href='/decks/add'>New Deck</Link>
-                            <div className='deck-list'>{ !this.props.decks || this.props.decks.length === 0 ? 'You have no decks, try adding one.' : deckList }</div>
+                            <DeckList className='deck-list' activeDeck={ this.props.selectedDeck } decks={ this.props.decks } onSelectDeck={ this.props.selectDeck } />
                         </Panel>
                     </div>
-                    { deckInfo }
```
cryogen: Undefined and null have different meanings to react, what does this evaluate to if the and fails?

PR #1839, opened 2018-03-21
Implement remaining neutral plots from Sands of Dorne

### LINKED ISSUE #1801: Walder Frey not reduced by Noble Cause
Was the only character I could play so it definitely wasn't missing that I had already played a Lord or Lady that turn. Haven't seen this with any other Lord or Lady. Other plot was a valar.

DukeTax:
Fixes #1801.

https://i.imgur.com/iCLXszW.jpg

### DIFF
--- a/server/game/cards/10-SoD/PeaceAndProsperity.js
+++ b/server/game/cards/10-SoD/PeaceAndProsperity.js
@@ -6,9 +6,9 @@ class PeaceAndProsperity extends PlotCard {
             targetType: 'player',
             targetController: 'current',
             effect: [
-                ability.effects.reduceNextMarshalledCardCost(3, card => card.getType() === 'character'),
-                ability.effects.reduceNextMarshalledCardCost(2, card => card.getType() === 'location'),
-                ability.effects.reduceNextMarshalledCardCost(1, card => card.getType() === 'attachment')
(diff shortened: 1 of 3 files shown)


server/game/cards/10-SoD/PeaceAndProsperity.js
```diff
@@ -0,0 +1,19 @@
...
+
+class PeaceAndProsperity extends PlotCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetType: 'player',
+            targetController: 'current',
+            effect: [
+                ability.effects.reduceNextMarshalledCardCost(3, card => card.getType() === 'character'),
+                ability.effects.reduceNextMarshalledCardCost(2, card => card.getType() === 'location'),
+                ability.effects.reduceNextMarshalledCardCost(1, card => card.getType() === 'attachment')
```
ystros: `reduceNextMarshalledCardCost` has a limit of 1 time fixed, so if the plot is recycled and replayed, the effect will no longer be active. Using `reduceFirstMarshalledCardCostEachRound` instead should properly reset the effect each round (and thus keep it active if recycled).

DukeTax: Thanks! You've just fixed #1801 as well. :)

ystros: Ah, good catch!

PR #1840, opened 2018-03-22
Validate against the restricted list

ystros:
In the deck editor:
screen shot 2018-03-21 at 8 32 30 pm
screen shot 2018-03-21 at 8 34 14 pm
screen shot 2018-03-21 at 8 44 04 pm
screen shot 2018-03-21 at 8 44 16 pm

In a pending game:
screen shot 2018-03-21 at 8 32 05 pm

### DIFF
--- a/client/DeckStatus.jsx
+++ b/client/DeckStatus.jsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import DeckStatusSummary from './DeckStatusSummary';
+import StatusPopOver from './StatusPopOver.jsx';
+
+class DeckStatus extends React.Component {
+    render() {
+        let { status } = this.props;
+        let statusName;
+        let className = 'deck-status';
+
+
+        if(this.props.className) {
+            className += ' ' + this.props.className;
+        }
+
+        if(!status.basicRules) {
+            statusName = 'Invalid';
+            className += ' invalid';
+        } else if(!status.faqJoustRules || !status.noUnreleasedCards) {
+            statusName = 'Casual play only';
+            className += ' casual-play';
+        } else {
+            statusName = 'Valid';
+            className += ' valid';
+        }
+
+        return (
+            <span className={ className }>
+                <StatusPopOver status={ statusName } show>
+                    <div>
+                        <DeckStatusSummary status={ status } />
+                        { status.extendedStatus && status.extendedStatus.length !== 0 &&
+                            <ul className='deck-status-errors'>
+                                { status.extendedStatus.map((error, index) => <li key={ index }>{ error }</li>) }
+                            </ul>
+                        }
+                    </div>
+                </StatusPopOver>
+            </span>);
+    }
+}
+
+DeckStatus.propTypes = {
+    className: PropTypes.string,
+    status: PropTypes.object.isRequired
+};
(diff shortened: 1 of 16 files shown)


ystros:
As part of this review, someone should spot check the structure of the associated data (which I didn't actually PR): https://github.com/ystros/throneteki-json-data/blob/master/restricted-list.json

Is this sufficient to represent the restricted list or should we worry about future updates that go beyond "restricted." I know Netrunner has a "removed" list as well. For example, let's say that FFG decides that Ser Pounce is overpowered and bans it from all decks. Presumably this would be done with an FAQ update, so it should be tied to version / date like the restricted cards. We could add it as another key in the data (e.g. `joustBannedCards`) but that structure seems a bit funky and it's unclear if the name "restricted list" is appropriate at that point.

Is this worth thinking about right now, or should we cross that bridge when FFG burns it out from under us?

client/DeckStatus.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+import DeckStatusSummary from './DeckStatusSummary';
+import StatusPopOver from './StatusPopOver.jsx';
+
+class DeckStatus extends React.Component {
+    render() {
+        let { status } = this.props;
+        let statusName;
+        let className = 'deck-status';
+
```
cryogen: Surprised that eslint doesn't pick up this double space, will have to see if I can find a rule for that

client/DeckStatusSummary.jsx
```diff
@@ -0,0 +1,30 @@
...
+            { title: `FAQ v${faqVersion} restricted list`, value: faqJoustRules },
+            { title: 'Only released cards', value: noUnreleasedCards }
+        ];
+
+        return (
+            <ul className='deck-status-summary'>
+                { items.map((item, index) => (
+                    <li className={ item.value ? 'valid' : 'invalid' } key={ index }>
+                        <span className={ item.value ? 'glyphicon glyphicon-ok' : 'glyphicon glyphicon-remove' } />
+                        { ' ' }
```
cryogen: Would this not be better as {` ${item.title}`} ?

cryogen: Or even &nbsp;{ item.title }

server/services/CardService.js
```diff
@@ -52,6 +54,10 @@ class CardService {
             logger.info(err);
         });
     }
+
+    getRestrictedList() {
+        return JSON.parse(fs.readFileSync(path.join(__dirname, '../../throneteki-json-data/restricted-list.json')));
```
cryogen: This has the potential to block the entire server if this read goes wrong for any reason, though I'm not sure what the alternative is.  Last I read the async file stuff wasn't that great.

ystros: Originally I had this as a simple `require` at the top of the file. However, doing so forces you to restart the lobby server anytime data gets updated. Not a big deal in terms of the restricted list, but if we move to loading packs / cards directly from file we'll probably want hot-reload in general.

I can change this method to return a promise instead and use the normal async `readFile` instead if that will alleviate potential problems.

cryogen:
Format seems good to me, and it should give enough flexibility whatever they throw at us, so lets go with this and review if and when they introduce something new!

PR #1841, opened 2018-03-22
Upgrade to react 16

cryogen:
Add error boundaries.  One for the global site, one for the current page.  Also add redux debug data to error reports


cryogen:
image

this is what it looks like when an error happens

PR #1845, opened 2018-03-23
Client reorg

cryogen:
Moved all of the components into a proper folder structure
removed .js/.jsx from imports as it's not needed
removed the 'inner' pattern that's also not needed
updated eslint rules to only allow 1 blank line
fixed existing violation of the above rule


ystros:
There are merge conflicts now.

cryogen:
Fixed

PR #1847, opened 2018-03-27
Disallow triggering action abilities when no valid opponents

### LINKED ISSUE #1497: Cards that use `chooseOpponent` should not trigger if no valid opponents
**Scenario**
Opponent has equal or higher reserve than the player.
Player clicks 'Night Gathers' in hand.

*Currently:*
The 1 gold for Night Gathers is paid, and it is placed in the discard pile.

*Desired:*
Nothing should happen because the condition for the ability hasn't been met.

I think this may only be broken on Action cards, but Interrupt / Reaction cards should be checked as well.

### LINKED ISSUE #1791: Seen in Flames illegal plays during Annals of Castle Black
Currently you can **play Seen in Flames** vs an opponent on **Annals turn** when the **hand is empty** (in order to trigger Core Mel), this is an illegal play as covered on: 
tDB - https://thronesdb.com/card/01064
cgdb - http://www.cardgamedb.com/forums/index.php?/topic/27997-seen-in-flames-on-an-empty-hand/

_Have not tried to do it on a normal turn, the issue might be there then aswell._

DukeTax:
Fixes a bug where Seen in Flames could be played when no opponents had cards in hand to trigger Melisandre.

Fixes #1497.
Fixes #1791.

### DIFF
--- a/server/game/cardaction.js
+++ b/server/game/cardaction.js
@@ -130,7 +130,7 @@ class CardAction extends BaseAbility {
             return false;
         }
 
-        return this.canPayCosts(context) && this.canResolveTargets(context);
+        return this.canResolveOpponents(context) && this.canPayCosts(context) && this.canResolveTargets(context);
     }
 
     execute(player, arg) {


DukeTax:
Responding to the comment in the issue: TriggeredAbility already has this check so reactions/interrupts should be fine.

PR #1852, opened 2018-03-29
add transitions for navigating between pages

cryogen:
will add a gif that shows this in action later...


cryogen:
http://www.giphy.com/gifs/2xPJvpbkNeXrv9hVFo

PR #1854, opened 2018-03-30
Implement You Win or You Die

hrnciar:
My first implemented card, please check whether it’s ok. I choosed the
easiest one for beginning :)


DukeTax:
Card implementation looks good. 👍 

You seem to have committed package-lock.json as well though, could you remove that from your commit?

@cryogen @ystros Should we even track the package files? They show up in my git status from time to time as well.

DukeTax:
Btw we have a throneteki dev discord you could join here: https://discord.gg/Rye4X2

hrnciar:
Yeah, sorry for that. It should be removed now.
Thank you, I will join.

DukeTax:
👍

PR #1858, opened 2018-03-30
Construct className props declaratively using `classnames`

ystros:


client/Components/Profile/CardSizeOption.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+function CardSizeOption(props) {
```
ystros: @cryogen You had mentioned you were warming up to the idea of "proper stateless components" so I used the function syntax, let me know if you want me to convert back to class syntax.

cryogen: I think it's fine.  I'm not keen on the deconstructing as I like being able to see at a glance which are props and which aren't but not sure I care enough to suggest changing it

PR #1862, opened 2018-04-02
Implement Fear Cuts Deeper Than Swords & Someone Always Tells

DukeTax:


server/game/intimidatekeyword.js
```diff
@@ -4,30 +4,26 @@ class IntimidateKeyword extends BaseAbility {
     constructor() {
         super({});
         this.title = 'Intimidate';
+        this.targets = super.buildTargets(this.getTargetProperty());
```
ystros: It would probably be better if inheritors of BaseAbility didn't call this method directly. I believe it was mostly extracted into its own method for readability purposes, not to allow inheritors to use it. If anything significant changes in the base class around how targets are loaded or stored, we'd have to track down these references and fix them.

Instead, try passing the target definition directly into the constructor and let the base class handle it. e.g.

```javascript
super(this.getTargetProperty());

// or

super({
  target: {
    // ...
  }
})
```

PR #1863, opened 2018-04-02
refactor drag drop 

cryogen:
to use react-dnd and generally tidy the whole thing up

### DIFF
--- a/client/Components/GameBoard/Droppable.jsx
+++ b/client/Components/GameBoard/Droppable.jsx
@@ -1,0 +1,129 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { DropTarget } from 'react-dnd';
+import classNames from 'classnames';
+
+import { ItemTypes } from '../../constants';
+
+const validTargets = {
+    hand: [
+        'play area',
+        'discard pile',
+        'draw deck',
+        'dead pile',
+        'out of game',
+        'conclave'
+    ],
+    'play area': [
+        'discard pile',
+        'hand',
+        'draw deck',
+        'dead pile',
+        'out of game',
+        'conclave'
+    ],
+    'discard pile': [
+        'dead pile',
+        'hand',
+        'draw deck',
+        'play area',
+        'out of game',
+        'conclave'
+    ],
+    'dead pile': [
+        'hand',
+        'draw deck',
+        'play area',
+        'discard pile',
+        'out of game',
+        'conclave'
+    ],
+    'draw deck': [
+        'hand',
+        'discard pile',
+        'dead pile',
+        'play area',
+        'out of game',
+        'conclave'
+    ],
+    'plot deck': [
+        'revealed plots',
+        'scheme plots',
+        'out of game'
+    ],
+    'revealed plots': [
+        'plot deck',
+        'scheme plots',
+        'out of game'
+    ],
+    'out of game': [
+        'plot deck',
+        'revealed plots',
+        'scheme plots',
+        'draw deck',
+        'play area',
+        'discard pile',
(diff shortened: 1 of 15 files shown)


client/Components/GameBoard/Droppable.jsx
```diff
@@ -0,0 +1,129 @@
...
+        'dead pile',
+        'out of game',
+        'conclave'
+    ],
+    'play area': [
+        'discard pile',
+        'hand',
+        'draw deck',
+        'dead pile',
+        'out of game',
```
ystros: It should be noted (but probably not worth fixing) that this and the other out-of-game drag combinations can lead to invalid locations. Example: dragging a Rains Scheme plot from out-of-game to the play area, or an out-of-game character to the plot deck.

This will already be prevented on the server-side implementation for drag-drop, so it won't lead to invalid game states. It just may lead to the UI displaying a drag as okay and then not following through with it.

client/Components/GameBoard/PlayerRow.jsx
```diff
@@ -154,33 +155,43 @@ class PlayerRow extends React.Component {
...
+                </Droppable>
+                <Droppable onDragDrop={ this.props.onDragDrop } source='dead pile'>
+                    <CardPile className='dead' title='Dead' source='dead pile' cards={ this.props.deadPile }
+                        onMouseOver={ this.props.onMouseOver } onMouseOut={ this.props.onMouseOut } onCardClick={ this.props.onCardClick }
+                        popupLocation={ this.props.isMe || this.props.spectating ? 'top' : 'bottom' }
+                        orientation='kneeled' size={ this.props.cardSize } />
+                </Droppable>
+
+                <Droppable onDragDrop={ this.props.onDragDrop } source='out of game'>
+                    { this.getOutOfGamePile() }
```
ystros: This will lead to a drag target for out-of-game pile when the pile isn't displayed (likely resulting in a 0x0 px drag target).

less/cards.less
```diff
@@ -456,4 +456,47 @@
...
+
+.overlay {
+  position: absolute;
+  padding: inherit;
+  margin: inherit;
+  top: 0;
+  left: 0;
+  height: 100%;
+  width: 100%;
+  z-index: 9999;
```
ystros: May be worth pulling this into a z-index variable in variables.less. 9999 is already defined as `@layer-top`, probably worth looking at what else is using that and seeing if the drag interactions need to be even higher.

client/Components/GameBoard/PlayerRow.jsx
```diff
@@ -154,33 +155,43 @@ class PlayerRow extends React.Component {
                     size={ this.props.cardSize } />
                 { this.getAgenda() }
```
ystros: Need to make Conclave be draggable as well (but probably not other agenda types). I imagine that it's still setting the now unused `onDragDrop` property in this method since I don't see a diff.

client/Components/GameBoard/Droppable.jsx
cryogen: It does, and also things like dragging attachments to the dead pile.  It's only intended as a rudimentary guide.  It could be tightened up to be 100% accurate, but it makes it a lot more complex

less/cards.less
cryogen: right, knew there was something i was supposed to go back and change..

client/Components/GameBoard/Droppable.jsx
ystros: Yeah. Like I said, worth acknowledging, but probably not really worth fixing.

PR #1865, opened 2018-04-03
Do not reveal strength / icons for facedown cards

ystros:
Returning information about a card's strength and the icons while it's
facedown can make it guessable what card it may be. This is a potential
problem during setup, for cards in hand, and cards in the upcoming
Shadow zone.


cryogen:
Holy cow that's an old bug!  Nice catch :)

PR #1866, opened 2018-04-03
Only allow Forced March to initiate twice

ystros:
The post-then effect of Forced March reads as "initiate this effect
again." The "this effect" refers to the pre-then effect, not to the
ability as a whole. This means initiating the post-then effect does not
lead to another instance of the "then", and thus you can only initiate
the effect a total of two times.

See: https://thronesdb.com/card/10048#review-441

### DIFF
--- a/server/game/cards/10-SoD/ForcedMarch.js
+++ b/server/game/cards/10-SoD/ForcedMarch.js
@@ -6,6 +6,7 @@ class ForcedMarch extends PlotCard {
     setupCardAbilities() {
         this.whenRevealed({
             handler: () => {
+                this.initiatedPostThen = false;
                 this.initiate();
             }
         });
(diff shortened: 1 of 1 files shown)


DukeTax:
Do you or scantrell have anything to back up that ruling? I feel like as long as it doesn't specify what part of the effect, it is the whole effect, and so should be able to be reinitiated as many times as both players have standing MIL icons still.

cryogen:
"If the effect text of an ability includes the word "then," the text preceding the word "then" must be successfully resolved in full (i.e. the game state changes to reflect the intent of the pre-then aspect of the effect in its entirety) before the remainder of the effect described after the word "then" can be resolved."

To me that pretty clearly states that both parts are considered two aspects of a single effect. 

https://thronesdb.com/rulesreference#The_word_Then

cryogen:
(The above is a quite from Facebook, not my words)

ystros:
There now seems to be consensus that it should be the whole effect, including Then, as previously implementing. Closing!

PR #1871, opened 2018-04-04
Abstract user

cryogen:
Convert the user object to be a class, so that the intent is much clearer as the data is passed from place to place
Fix a bug where the user data was not sent across the wire from the game node on a sync, so settings/blocklist data were lost causing the block list not to work

### DIFF
--- a/server/lobby.js
+++ b/server/lobby.js
@@ -690,6 +690,11 @@ class Lobby {
         _.each(games, game => {
             let owner = game.players[game.owner];
 
+            if(!owner) {
+                logger.error('Got a game where the owner wasn\'t a player', game.owner);
+                return;
+            }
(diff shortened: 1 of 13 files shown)


server/lobby.js
```diff
@@ -694,7 +688,9 @@ class Lobby {
 
     onNodeReconnected(nodeName, games) {
         _.each(games, game => {
-            let syncGame = new PendingGame({ username: game.owner }, { spectators: game.allowSpectators, name: game.name });
+            let ownerUser = game.players[game.owner];
```
ystros: Wouldn't normally comment on naming, but this is a bit confusing when followed by `ownerUser.user`. Really, this is a player object.

PR #1873, opened 2018-04-04
Remove the scheme plots pile

cryogen:
Change rains to use effects to restrict the available plot selection
Update the plot selection to be able to show valid targets
Add the ability to tell the client to group cards and change the client to group said cards
Make the target select cursor conditional and only apply to the board element so that it doesnt stick if a game breaks

### DIFF
--- a/client/Components/GameBoard/CardPile.jsx
+++ b/client/Components/GameBoard/CardPile.jsx
@@ -19,6 +19,17 @@ class CardPile extends React.Component {
         };
     }
 
+    componentWillReceiveProps(props) {
+        let hasNewSelectableCard = props.cards && props.cards.some(card => card.selectable);
+        let didHaveSelectableCard = this.props.cards && this.props.cards.some(card => card.selectable);
+
+        if(!didHaveSelectableCard && hasNewSelectableCard) {
+            this.setState({ showPopup: true });
+        } else if(didHaveSelectableCard && !hasNewSelectableCard) {
+            this.setState({ showPopup: false });
+        }
+    }
+
     onCollectionClick(event) {
         event.preventDefault();
 
@@ -84,11 +95,14 @@ class CardPile extends React.Component {
         }
     }
 
-    getPopup() {
-        let popup = null;
+    getCardList(cards) {
         let cardIndex = 0;
 
-        let cardList = _.map(this.props.cards, card => {
+        if(!cards) {
+            return null;
+        }
+
+        let cardList = cards.map(card => {
             let cardKey = card.uuid || cardIndex++;
             return (<Card key={ cardKey } card={ card } source={ this.props.source }
                 disableMouseOver={ this.props.disableMouseOver }
@@ -99,6 +113,28 @@ class CardPile extends React.Component {
                 orientation={ this.props.orientation === 'kneeled' ? 'vertical' : this.props.orientation }
                 size={ this.props.size } />);
         });
+
+        return cardList;
+    }
+
+    getPopup() {
+        let popup = null;
+
+        let cardList = [];
+
+        if(this.props.cards && this.props.cards.some(card => card.group)) {
+            let cardGroup = _.groupBy(this.props.cards, card => card.group);
+            for(const [type, cards] of Object.entries(cardGroup)) {
+                cardList.push(
+                    <div key={ type }>
+                        <div className='group-title'>{ `${type} (${cards.length})` }</div>
+                        { this.getCardList(cards) }
+                    </div>
+                );
+            }
+        } else {
+            cardList = this.getCardList(this.props.cards);
+        }
 
         if(this.props.disablePopup || !this.state.showPopup) {
             return null;
@@ -211,8 +247,7 @@ CardPile.propTypes = {
     popupMenu: PropTypes.array,
     size: PropTypes.string,
     source: PropTypes.oneOf(['hand', 'discard pile', 'play area', 'dead pile', 'draw deck', 'plot deck',
-        'revealed plots', 'selected plot', 'attachment', 'agenda', 'faction', 'additional',
-        'scheme plots', 'conclave']).isRequired,
+        'revealed plots', 'selected plot', 'attachment', 'agenda', 'faction', 'additional', 'conclave']).isRequired,
     title: PropTypes.string,
     topCard: PropTypes.object
 };
--- a/client/Components/GameBoard/CardPile.jsx
+++ b/client/Components/GameBoard/CardPile.jsx
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 import _ from 'underscore';
 
 import Card from './Card';
+import CardTiledList from './CardTiledList';
 import Droppable from './Droppable';
 
 class CardPile extends React.Component {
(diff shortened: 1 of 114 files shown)


cryogen:
Plot deck with rains:
image

Select plot prompt (plot auto opens)
image

Rains trigger
image

server/game/cards/10-SoD/AtPrinceDoransBehest.js
```diff
@@ -24,10 +24,10 @@ class AtPrinceDoransBehest extends PlotCard {
...
     }
 
-    getPlotSourceForPhase(card) {
-        return (this.game.currentPhase !== 'plot') ? 
-            ['plot deck', 'scheme plots'].includes(card.location) :
-            card.location === 'plot deck';
+    checkPlotForPhase(card) {
+        return (this.game.currentPhase !== 'plot') ?
+            card.location === 'plot deck' :
+            card.location === 'plot deck' && !card.hasTrait('scheme');
```
ystros: This will prevent Behest from revealing a Scheme during the plot phase, which we'd only want to prevent if the player is using Rains as their agenda. Checking also that they can select schemes should solve it.

server/game/cards/agendas/TheRainsOfCastamere.js
```diff
@@ -33,11 +31,20 @@ class TheRainsOfCastamere extends AgendaCard {
...
             },
             handler: context => this.trigger(context)
         });
+
+        this.persistentEffect({
+            targetType: 'player',
+            targetController: 'current',
+            effect: [
+                ability.effects.cannotSelectSchemes(),
+                ability.effects.groupCardPile('plot deck')
```
ystros: This feels like the server knowing too much about the UI presentation. Shouldn't this check be done on the client as it was previously when the client would check for the agenda and then conditionally render a separate schemes pile?

It feels a lot like the `additionalPiles` mechanism (ironically, it was used for the scheme pile) we used to have that we took out because it had too much knowledge about the UI.

server/game/effects.js
```diff
@@ -1141,6 +1141,26 @@ const Effects = {
                 game.skipPhase[name] = false;
             }
         };
+    },
+    cannotSelectSchemes: function() {
+        return {
+            apply: function(player) {
+                player.cannotSelectSchemes = true;
```
ystros: This should work fine for now, but given all the things that are coming that interact with the plot deck I'm wondering if a more general solution would be justified. Don't have a list of those memorized so not sure if it's beneficial to anything coming. But you could do this more generally as:
```
this.persistentEffect({
  condition: () => this.game.currentPhase === 'plot',
  effect: ability.effects.cannotRevealPlot(plot => plot.hasTrait('Scheme')
});
```

I also worry somewhat about the 'Scheme cards are not considered to be in your plot deck' aspect during the plot phase in terms of future interactions with other cards, but we can cross that bridge when we come to it. I couldn't even begin to guess what the future implications are.

client/Components/GameBoard/CardPile.jsx
```diff
@@ -19,6 +19,17 @@ class CardPile extends React.Component {
         };
     }
 
+    componentWillReceiveProps(props) {
+        let hasNewSelectableCard = props.cards && props.cards.some(card => card.selectable);
+        let didHaveSelectableCard = this.props.cards && this.props.cards.some(card => card.selectable);
+
+        if(!didHaveSelectableCard && hasNewSelectableCard) {
+            this.setState({ showPopup: true });
```
ystros: It feels like this should be controlled at a higher level somehow, to handle the (admittedly rare) cases where cards are selectable in multiple piles. Per @DukeTax example in chat, Summer will potentially cause both the discard and dead piles to open simultaneously and overlapping. Ideally either only one would open or neither would open, leaving the player to click the one they want to choose from.

server/game/effects.js
DukeTax: A possible complication in this regard could arise when we convert The Crow is a Tricksy Bird to use a target prompt, as it would show even the Scheme plots of the Rains player, which it shouldn't. We could obviously work around that by checking if the opponent has Rains as their agenda on The Crow, but that seems a little iffy...

server/game/gamesteps/plot/selectplotprompt.js
```diff
@@ -50,9 +51,30 @@ class SelectPlotPrompt extends AllPlayerPrompt {
...
+        player.clearSelectableCards();
 
         this.game.addMessage('{0} has selected a plot', player);
     }
+
+    highlightSelectableCards(player) {
+        let selectableCards = this.game.allCards.filter(card => card.getType() === 'plot' &&
+                                                                card.location === 'plot deck' &&
+                                                                card.controller === player &&
+                                                                (player.cannotSelectSchemes && !card.hasTrait('scheme')));
```
ystros: This check is causing all plots to be displayed as 'unselectable' when playing with a non-Rains agenda. It still lets the player to select the plot, so it's strictly a UI bug.

server/game/cards/agendas/TheRainsOfCastamere.js
cryogen: I sort of agree, but on the other hand the client is supposed to be somewhat dumb about the rules engine and just do what the server tells it.  I plan to use this mechanism, or something very like it, to enable the rookery display at the start of the game.

If anything, this is more of a hint that there's a new property of the data.  It's still up to the client to decide how to display that change.

Happy to consider alternatives that don't involve the client checking for specific cards or agendas

server/game/effects.js
cryogen: Funny, before I read this, but after our earlier discussions, I was thinking about changing it to be exactly that :)

I'll see what I can do

client/Components/GameBoard/CardPile.jsx
cryogen: I think I disagree with this somewhat.  Mainly because even in the case of multiple selection targets, how do we decide which should take priority?  Moving it up a level will involve a whole lot more passing around of state and components that have no business knowing about it being educated about it.

The other alternative is to drive it from the server, but I'm not sure that's a good idea either.

I personally think popping up the dead pile and discard pile is fine in the case of summer.  If the piles get in the way of each other than we can fix that with styling.

server/game/gamesteps/plot/selectplotprompt.js
cryogen: Oops.

server/game/cards/10-SoD/AtPrinceDoransBehest.js
cryogen: I think the solution to the below comment will fix this by default

server/game/player.js
```diff
@@ -682,7 +683,7 @@ class Player extends Spectator {
     }
 
     recyclePlots() {
-        if(this.plotDeck.isEmpty()) {
+        if(this.plotDeck.isEmpty() || (this.groupedPiles['plot deck'] && this.plotDeck.all(plot => plot.hasTrait('Scheme')))) {
```
cryogen: @ystros not amazingly happy with this, do you have a better idea?

cryogen:
Better screenshots over here: https://github.com/cryogen/throneteki/pull/1874

server/game/effects.js
cryogen: @DukeTax my changes don't fix your "Being able to see your opponents schemes" issue, but let's worry about that if and when we convert that to a target prompt :)

server/game/player.js
ystros: The rules accurate version of this would be to have a proper "is not considered in the plot deck" effect applied to the Schemes and then check how many cards *are* considered in the plot deck currently. If it's 0, then you recycle.

You could try to fake it w/ the current `cannotReveal` effect as we are currently, but I think that may cause Tricksy Crow to cause early recycles (assuming the effect doesn't wear off at the proper time). Worth looking at, perhaps.

Barring that, I'd honestly be more comfortable with checking that the agenda is Rains here rather than the presentational `groupedPiles` property. How things are grouped in the UI should not be affecting rules.

PR #1874, opened 2018-04-04
Update styling for pile popup groups

ystros:
* Scales the size of the popup based on the card size selected (previously stuck on normal sized calculations, so you'd get more wrapping if you set card size to large or extra large)
* Updated styling for grouped cards in the popup.

Normal:
screen shot 2018-04-04 at 3 39 03 pm

Select plot:
screen shot 2018-04-04 at 3 39 19 pm

Select Rains:
screen shot 2018-04-04 at 3 39 37 pm

### DIFF
--- a/client/Components/GameBoard/CardPile.jsx
+++ b/client/Components/GameBoard/CardPile.jsx
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 import _ from 'underscore';
 
 import Card from './Card';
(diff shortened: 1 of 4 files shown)


cryogen:
I'm not sure what value adding the CardTiledList component has, given how many props it has to be passed, I think it was fine as a regular HTML element.

That being said, I'll take it as-is

PR #1884, opened 2018-04-07
Implement I Am No One

DukeTax:


ystros:
For now, I'm going to assume that the lose traits / factions aspect of this card means "lose 1 instance of each trait / faction" and not "this card no longer has traits / factions". The latter is more thematic but the former is more likely. If the card text suggests otherwise when we get it, we can update this card.

DukeTax:
@ystros Oh right, I meant to add this but I forgot. I asked mplain and he referred me here: http://www.cardgamedb.com/index.php/agameofthrones2ndedition/a-game-of-thrones-2nd-edition-cards/_/lions-of-casterly-rock/sweet-donnel-hill-r543?st=0#comment_65056

"In first edition and in Conquest we treated “all keywords” as “one instance of each keyword.”
 
I think it’s best to remain consistent with that interpretation here.
 
So that, if a character has stealth, loses all keywords, and is gaining stealth, it is
at +1 stealth (1 for printed, -1 for loses all, and +1 for the gain).
 
Thanks!
 
Nate"

PR #1896, opened 2018-04-09
Refactor the deck builder again

### LINKED ISSUE #1825: Deck builder caches unsaved changes
### Scenario:
* Edit an existing deck
* Add some cards to the deck or delete some existing ones
* Without clicking Save, navigate to the Decks page

### Currently:
* When viewing the deck, whatever cards were in the deck after editing it are listed

### Expected:
* When viewing the deck, it should have the same cards it had before editing it.

cryogen:
Fixes #1825

### DIFF
--- a/client/Components/Decks/DeckEditor.jsx
+++ b/client/Components/Decks/DeckEditor.jsx
@@ -1,7 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import _ from 'underscore';
-import $ from 'jquery';
 import { connect } from 'react-redux';
 
 import Input from '../Form/Input';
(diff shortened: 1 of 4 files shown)


client/Components/Decks/DeckEditor.jsx
```diff
@@ -1,54 +1,92 @@
...
 
-    componentWillMount() {
-        if(!this.props.deck.faction && this.props.factions) {
-            let deck = this.copyDeck(this.state.deck);
+        if(props.deck) {
+            this.state.deckId = props.deck._id;
+            this.state.deckName = props.deck.name;
+            this.state.plotCards = props.deck.plotCards;
+            this.state.drawCards = props.deck.drawCards;
+            this.state.bannerCards = props.deck.bannerCards;
```
ystros: Should copies of these arrays be made instead of direct references? From the looks of the code flow, this will result in the deck in the store being updated as well, so I'm not 100% confident it fixes #1825.

cryogen: The deck is now owned by the add/edit components and is not in redux at all, so it definitely fixes it :)

PR #1910, opened 2018-04-12
convert last remaining usages of findCardByUuid and then remove it

cryogen:


server/game/gamesteps/decksearchprompt.js
```diff
@@ -67,8 +67,8 @@ class DeckSearchPrompt extends UiPrompt {
     }
 
     buttons() {
-        let uniqueCardsByTitle = _.uniq(this.searchCards(), card => card.cardData.label);
-        let buttons = _.map(uniqueCardsByTitle, card => {
+        this.uniqueCardsByTitle = _.uniq(this.searchCards(), card => card.cardData.label);
```
cryogen: it was a toss up between putting this and 'buttons' on the class, but since there's a method already called buttons(), this won

PR #1912, opened 2018-04-13
Various styling fixes

cryogen:
shrink the nav bar
shrink the player stat rows
add horizontal scrollbar to play area
fix firefox to not grow the play area vertically beyond the screen with no way of scrolling
extend the chat window full height
make the card zoom overlay
fix card row spacing

### DIFF
--- a/less/gameboard.less
+++ b/less/gameboard.less
@@ -4,7 +4,7 @@
   bottom: 0;
   left: 0;
   right: 0;
-  top: 40px;
+  top: @navbar-height;
   display: flex;
   justify-content: space-between;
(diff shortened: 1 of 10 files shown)


less/gameboard.less
```diff
@@ -4,7 +4,7 @@
   bottom: 0;
   left: 0;
   right: 0;
-  top: 50px;
+  top: 40px;
```
ystros: This has shown up a few times, is it the height of the nav bar? May be worth pulling into a variable (e.g. `@nav-bar-height`)

PR #1922, opened 2018-04-16
Implement Septon Cellador

DukeTax:
Character, Night's Watch, The Seven, Loyal
Cost: 4
STR: 2
Icons: Intrigue
Challenge action: kneel Septon Cellador to choose an opponent. This player must choose and discard a character from his hand unless he reveals a hand containing no character.

### DIFF
--- a/server/game/cards/08.6-SAT/SeptonCellador.js
+++ b/server/game/cards/08.6-SAT/SeptonCellador.js
@@ -1,0 +1,68 @@
+const DrawCard = require('../../drawcard.js');
+
+class SeptonCellador extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Opponent discards character',
+            phase: 'challenge',
+            cost: ability.costs.kneelSelf(),
+            chooseOpponent: opponent => !opponent.hand.isEmpty(),
+            handler: context => {
+                if(context.opponent.hand.any(card => card.getType() === 'character')) {
+                    this.promptForCharacterDiscard(context);
+                    return;
+                }
+
+                this.revealHandPrompt(context);
+            }
+        });
+    }
+
+    promptForCharacterDiscard(context) {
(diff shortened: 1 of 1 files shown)


server/game/cards/08.6-SAT/SeptonCellador.js
```diff
@@ -0,0 +1,68 @@
...
+
+    revealHandPrompt(context) {
+        let cards = [];
+        context.opponent.hand.each(card => cards.push(card));
+
+        this.game.addMessage('{0} kneels {1} to have {2} reveal {3} as their hand',
+            context.player, this, context.opponent, cards);
+
+        this.game.promptForSelect(context.player, {
+            activePromptTitle: 'Select a card or click done to stop looking',
```
ystros: I just realized that in addition to this not being great in Melee (theoretical), it also won't reveal the cards to spectators (non-theoretical). Is this a problem we care about?

DukeTax: Eehh, I don't think so? Spectators won't see the cards flipped but they see the chatmessage all the same. Regarding melee: we could even check to see if the game is a melee game, and disable the (fake) prompt altogether in that case, to keep things equal/fair among all players. But I don't think it matters much.

ystros: I somehow missed that there was a chat message 2 lines above. Which definitely works for spectators, but makes me wonder if there's any value added by this fake prompt. As I said in chat, I like how it mirrors what would happen in a real game but I was thinking you meant it as a replacement for the chat message.

DukeTax: The value is there solely for the joust players who like the visual aspect better than the message, but you're right in that it's limited, not really necessary and I don't like misusing prompts for this purpose myself, so I've gotten rid of it. Also removed the underscore usage in anticipation of #1924 .

PR #1923, opened 2018-04-17
Blank traits

cryogen:


server/game/basecard.js
```diff
@@ -295,7 +299,7 @@ class BaseCard {
     }
 
     hasTrait(trait) {
-        return this.traits.contains(trait);
+        return !this.isBlank(true) && this.traits.contains(trait);
```
cryogen: Will this work, or will traits need to be converted to effects? *shudder*

jeremylarner: Looks like it should work to me...

server/game/PropertyTypes/BlankingProperty.js
```diff
@@ -0,0 +1,19 @@
...
+
+    addBlankingSource(source, includingTraits = false) {
+        this.blanks[source.uuid] = { source: source, includingTraits: includingTraits };
+    }
+
+    removeBlankingSource(source) {
+        delete this.blanks[source.uuid];
+    }
+
+    isBlank(includingTraits = false) {
```
ystros: Could we split this into two separate methods (e.g. `isTextBlank()` and `areTraitsBlank()`)? Passing boolean params to a getter is hard to understand unless you jump back and forth between usage and definition. I did a double take when I saw `isBlank(true)`.

server/game/basecard.js
ystros: This is technically incorrect. Consider - a trait-blanked character with the Knighted attachment attached. Traits from the character's text box should be gone, but they should still get the Knight trait from the attachment.

However, since the only thing that blanks traits is White Shadows, I'm not sure it really matters.

server/game/effects.js
```diff
@@ -497,13 +497,15 @@ const Effects = {
...
         ];
     },
-    blank: {
-        apply: function(card) {
-            card.setBlank();
-        },
-        unapply: function(card) {
-            card.clearBlank();
-        }
+    blank: function(includingTraits = false) {
```
ystros: As above, `.blank(true)` doesn't read very well. Either a separate effect name (e.g. `.blankIncludingTraits()` can use the same underlying implementation) or sending a human readable flag (`.blank({ includeTraits: true })`) may be better.

server/game/PropertyTypes/BlankingProperty.js
```diff
@@ -0,0 +1,19 @@
+class BlankingProperty {
+    constructor() {
+        this.blanks = {};
+    }
+
+    addBlankingSource(source, includingTraits = false) {
+        this.blanks[source.uuid] = { source: source, includingTraits: includingTraits };
```
ystros: I understand how you're using `source` here but is it really necessary? It's only being used for removal tracking, but you could do the same thing by tracking the number of "blank (except traits)" and "blank (including traits)". Removing the specific source isn't important here (though may be in other scenarios), just removing 1 of the matching types of blanks.

PR #1927, opened 2018-04-18
Add rookery support to the deck builder

cryogen:
This only adds support to the deck builder, it does not add server side support, or allow rookery games to be created

### DIFF
--- a/client/deck-validator.js
+++ b/client/deck-validator.js
@@ -299,13 +299,13 @@ class DeckValidator {
                 {
                     message: 'More than 2 plot cards in rookery',
                     condition: deck => {
-                        return !deck.rookeryCards || deck.rookeryCards.filter(card => card.card.type === 'plot').length <= 2;
+                        return !deck.rookeryCards || getDeckCount(deck.rookeryCards.filter(card => card.card.type === 'plot')) <= 2;
                     }
                 },
                 {
                     message: 'More than 10 draw cards in rookery',
                     condition: deck => {
-                        return !deck.rookeryCards || deck.rookeryCards.filter(card => card.card.type !== 'plot').length <= 10;
+                        return !deck.rookeryCards || getDeckCount(deck.rookeryCards.filter(card => card.card.type !== 'plot')) <= 10;
                     }
                 }
             ]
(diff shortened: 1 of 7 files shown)


client/deck-validator.js
```diff
@@ -298,19 +325,20 @@ class DeckValidator {
...
-        let mayIncludeFuncs = _.compact(_.pluck(validators, 'mayInclude'));
-        let cannotIncludeFuncs = _.compact(_.pluck(validators, 'cannotInclude'));
-        let combinedRules = _.reduce(validators, (rules, validator) => rules.concat(validator.rules || []), []);
+        let mayIncludeFuncs = validators.map(validator => validator.mayInclude).filter(v => !!v);
+        let cannotIncludeFuncs = validators.map(validator => validator.cannotInclude).filter(v => !!v);
+        let combinedRules = validators.reduce((rules, validator) => rules.concat(validator.rules || []), []);
         let combined = {
-            mayInclude: card => _.any(mayIncludeFuncs, func => func(card)),
-            cannotInclude: card => _.any(cannotIncludeFuncs, func => func(card)),
+            mayInclude: card => mayIncludeFuncs.every(func => func(card)),
```
cryogen: This raises red flags with me as I've changed the logic and I don't think I should have had to, would appreciate extra scrutiny here

ystros: What situation led you to change it? I believe the previous logic should have been correct. This will fail on banner decks - the faction rules `mayInclude` requires a card be the faction or neutral. The banner `mayInclude` requires a card be the banner faction and non-loyal. It can't ever be both unless you do something weird like Lannister bannered w/ Lannister.

cryogen: Putting a non banner non faction card in the deck wouldn't trigger a validation error until I changed it.  I'm really not sure why..

client/deck-validator.js
```diff
@@ -281,9 +285,32 @@ class DeckValidator {
...
         let agendaRules = this.getAgendaRules(deck);
-        return this.combineValidationRules([standardRules, factionRules].concat(agendaRules));
+        let rookeryRules = this.getRookeryRules(deck);
+
+        return this.combineValidationRules([standardRules, factionRules, rookeryRules].concat(agendaRules));
+    }
+
+    getRookeryRules() {
+        return {
+            mayInclude: () => true,
```
ystros: This shouldn't be here. It's explicitly allowing every single card in the full deck, regardless of other `mayInclude`s.

It also might be nice to add a rule below that the player doesn't include agendas in the rookery.

PR #1934, opened 2018-04-25
Click cards to trigger interrupt / reaction abilities

ystros:
* When prompted for an interrupt or reaction window, players now click on the card they want to trigger directly instead of clicking menu buttons.
* If the same card can trigger on multiple simultaneous events (e.g. Marya Seaworth when multiple characters are bypassed from stealth), the player is then prompted to click on the card associated with the event.
* If the card has multiple choices (e.g. LoCR Tyrion), the choice is presented in a button menu after the card is chosen.
* Action abilities continue to be triggered through card-level menus during the appropriate action window.
* When revealed and forced interrupt / reaction order continues to be chosen by button prompts for now.

I highly recommend pulling this locally to try it out rather than reviewing solely based on code changes.

### DIFF
--- a/client/Components/GameBoard/Card.jsx
+++ b/client/Components/GameBoard/Card.jsx
@@ -66,6 +66,12 @@ class InnerCard extends React.Component {
     }
 
     isAllowedMenuSource() {
+        // Explicitly disable menus on agendas when they're selectable during a
+        // card select prompt.
+        if(this.props.source === 'agenda' && this.props.card.selectable) {
+            return false;
+        }
+
         return this.props.source === 'play area' || this.props.source === 'agenda' || this.props.source === 'revealed plots';
     }
 
--- a/client/Components/GameBoard/CardPile.jsx
+++ b/client/Components/GameBoard/CardPile.jsx
@@ -77,7 +77,7 @@ class CardPile extends React.Component {
             return;
         }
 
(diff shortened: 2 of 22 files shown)


client/Components/GameBoard/Card.jsx
```diff
@@ -66,6 +66,12 @@ class InnerCard extends React.Component {
     }
 
     isAllowedMenuSource() {
+        // Explicitly disable menus on agendas when they're selectable during a
```
cryogen: We probably need a better, more generic way of doing this, but this is fine for now

client/Components/GameBoard/CardPile.jsx
```diff
@@ -88,6 +88,14 @@ class CardPile extends React.Component {
         this.setState({ showPopup: !this.state.showPopup });
     }
 
+    get isTopCardSelectable() {
```
cryogen: Super minor, but I think this doesn't make sense as a 'get' property

client/Components/GameBoard/Card.jsx
ystros: Agreed. Originally I was going to see about making menus more dynamic on the server-side for this purpose, but it wasn't 100% clear if that'd be useful for other things and the PR was already getting pretty large, so I went with the quicker fix.

PR #1942, opened 2018-04-30
implement Driftwood Cudgel (SAT)

zacchiro:


zacchiro:
french spoiler: https://i62.servimg.com/u/f62/13/09/73/62/4uyk6w11.jpg

not to be merged yet, as the english card name is still unknown

also, moving attachments is kinda black magic for me, so this is likely buggy as hell…

server/game/cards/08.6-SAT/GourdinDeBoisFlotte.js
```diff
@@ -0,0 +1,33 @@
...
+    setupCardAbilities(ability) {
+        this.attachmentRestriction({ trait: 'Drowned God' });
+
+        this.whileAttached({
+            effect: ability.effects.dynamicStrength(() => this.controller.deadPile.length)
+        });
+
+        this.interrupt({
+            when: {
+                onCharacterKilled: event => event.cardStateWhenKilled === this.parent
```
DukeTax: This won't work as cardStateWhenKilled is not a perfect copy of the original card object.

But, since this is an interrupt, simply using `onCharacterKilled: event => event.card === this.parent` works.

server/game/cards/08.6-SAT/GourdinDeBoisFlotte.js
```diff
@@ -0,0 +1,33 @@
...
+            },
+            target: {
```
DukeTax: This needs an added `type: 'select'` so the opponent won't get informed of the new parent before deciding whether to cancel.

server/game/cards/08.6-SAT/GourdinDeBoisFlotte.js
```diff
@@ -0,0 +1,33 @@
...
+                cardCondition: card =>
+                    card.location === 'play area' && card.getType() === 'character'
+                    && this.controller.canAttach(this, card)
```
DukeTax: This needs an extra `card !== this.parent` check so the ability does not prompt in case there are no other legal parents.

zacchiro:
Thanks @DukeTax ! all done, including the rename to the official card name.
(FWIW the renames have now been done also for all the other pending SAT PRs.)

PR #1957, opened 2018-05-09
Fix timing for At Prince Doran's Behest

### LINKED ISSUE #1864: At Prince Doran's Behest and multiple "When Revealed" triggers
So this seems to be an issue specific to this plot at the moment (but is technically a timing issue for plots in general). Basically if player A reveals At Prince Doran's Behest and player B reveals any other plot with a "when revealed", and then player A chooses another plot with a "when revealed" (from At Prince Doran's Behest) then the first player should have the control over the order the remaining plots trigger.

Example:
Player A flips At Prince Doran's Behest, Player B flips Marched to the Wall.
Player B chooses to be first player and chooses to resolve At Prince Doran's Behest first.
Player A reveals Valar Morghulis.

With the current implementation the first player chooses which player resolves their "when reveals" first, rather than choosing which plots trigger first. So Player A resolves all of their plot effects at once, so will immediately trigger the Valar. However I believe what should happen is the first player should be able to trigger Marched to the Wall prior to the Valar if they choose to.

The exact same thing will happen if both players flip At Prince Doran's Behest, just with an extra iteration.

Relevant ruling.
http://www.cardgamedb.com/forums/index.php?/topic/38115-at-prince-dorans-behest-questions/

ystros:
* Ensures that first player gets to choose plot order after a new plot has been revealed by Behest
* Adds the capability to have a separate interrupt window with combined reactions, allowing proper interrupt / reaction timing after a new plot has been revealed by Behest.
* Fixes timing on ability copiers like Varys's Riddle to resolve the copied ability in the context of their own "When Revealed" event (necessary w/ Behest, otherwise it generates a nested sequence)

Fixes #1864

### DIFF
--- a/test/server/cards/04.2-CtA/SummerHarvest.spec.js
+++ b/test/server/cards/04.2-CtA/SummerHarvest.spec.js
@@ -26,22 +26,6 @@ describe('SummerHarvest', function() {
             it('should calculate the gold amount properly', function() {
                 expect(this.summerHarvest.getIncome()).toBe(7);
             });
-
-            describe('when playing it again after going through all plots', function() {
-                beforeEach(function() {
-                    this.completeMarshalPhase();
-                    this.completeChallengesPhase();
-
-                    // Move Summer Harvest back to the plot deck so it's eligible to be picked again.
-                    this.summerHarvest.controller.moveCard(this.summerHarvest, 'plot deck');
-
-                    this.selectFirstPlayer(this.player1);
-                });
-
-                it('should still properly calculate the gold amount properly', function() {
-                    expect(this.summerHarvest.getIncome()).toBe(7);
-                });
-            });
         });
 
         describe('when played against Varys\'s Riddle', function() {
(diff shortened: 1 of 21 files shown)


test/server/cards/04.2-CtA/SummerHarvest.spec.js
```diff
@@ -26,22 +26,6 @@ describe('SummerHarvest', function() {
...
-                    this.completeChallengesPhase();
-
-                    // Move Summer Harvest back to the plot deck so it's eligible to be picked again.
-                    this.summerHarvest.controller.moveCard(this.summerHarvest, 'plot deck');
-
-                    this.selectFirstPlayer(this.player1);
-                });
-
-                it('should still properly calculate the gold amount properly', function() {
-                    expect(this.summerHarvest.getIncome()).toBe(7);
```
ystros: This test was very strange and inconsistent:
* The deck only has one plot in it, so it's unclear if we have defined behaviour for that after recycling.
* To get around that, `moveCard` was called manually instead of the more appropriate `dragCard` to simulate UI interaction.
* The assertion itself was wrong - this ends up being Summer Harvest vs Varys's Riddle, and given all the above it wasn't prompting properly for that interaction and instead just set income to 7.

Since the bug it was protecting against is long gone - in the original version, it was adding to the plot's gold modifier directly each round instead of setting a base income - it didn't seem worth the trouble to rework the test.

PR #1990, opened 2018-05-24
Server side privacy / avatar changes

cryogen:
Do not send emailhash anywhere
add an 'enablegravatar' flag which lets users decide whether or not to turn on avatars.  Default off
download avatar from gravatar if this option is turned on

### DIFF
--- a/server/game/gamechat.js
+++ b/server/game/gamechat.js
@@ -22,8 +22,8 @@ class GameChat {
         args = _.reduce(args, (argList, arg) => {
             if(arg instanceof Spectator) {
                 argList.push(arg.name);
-            } else if(arg && arg.name) {
-                argList.push({ name: arg.name, argType: 'player' });
+            } else if(arg && arg.name && arg.argType === 'player') {
+                argList.push({ name: arg.name, argType: arg.argType });
             } else {
                 argList.push(arg);
             }


server/api/account.js
```diff
@@ -604,8 +632,27 @@ module.exports.init = function(server) {
...
+
+        await downloadAvatar(user);
+
+        res.send({ success: true });
+    }));
 };
 
+async function downloadAvatar(user) {
+    let avatar = await util.httpRequest(`https://www.gravatar.com/avatar/${user.emailHash}?d=identicon&s=24`, { encoding: null });
+    await writeFile(`public/img/avatar/${user.username}.png`, avatar, 'binary');
```
ystros: Is our username validation in a state such that the username will always be a valid filename? E.g. if slashes are allowed in the username ("Guns/Roses") then it will look for things under a non-existent sub directory.

server/api/account.js
```diff
@@ -471,6 +489,16 @@ module.exports.init = function(server) {
             user.password = await hashPassword(userToSet.password, 10);
         }
 
+        if(userToSet.enableGravatar && !user.enableGravatar) {
+            user.emailHash = crypto.createHash('md5').update(user.email).digest('hex');
```
ystros: Would it be better to do this calculation as a local variable within the `downloadAvatar` method instead of storing it on the user object? We don't need it for anything else and not storing it reduces the risk of it leaking out accidentally in other areas.

server/api/account.js
```diff
@@ -166,16 +181,19 @@ module.exports.init = function(server) {
...
-            emailHash: crypto.createHash('md5').update(req.body.email).digest('hex'),
+            emailHash: req.body.enableGravatar ? crypto.createHash('md5').update(req.body.email).digest('hex') : DefaultEmailHash,
+            enableGravatar: req.body.enableGravatar,
             verified: false,
             activiationToken: activiationToken,
             activiationTokenExpiry: formattedExpiration,
             registerIp: req.get('x-real-ip')
         };
 
+        await downloadAvatar(newUser);
```
ystros: In other projects I would have done this download (as well as sending email) in some sort of background job queue. That way, if there's latency between us and Gravatar, the user isn't stuck waiting until it finally clears or it times out.

Perhaps we can achieve something similar here without a job queue. What would the consequences be if we didn't `await` the download? Just fire it off, and forget it?

server/api/account.js
```diff
@@ -471,6 +489,16 @@ module.exports.init = function(server) {
             user.password = await hashPassword(userToSet.password, 10);
         }
 
+        if(userToSet.enableGravatar && !user.enableGravatar) {
+            user.emailHash = crypto.createHash('md5').update(user.email).digest('hex');
+        } else if(!userToSet.enableGravatar && user.enableGravatar) {
+            user.emailHash = DefaultEmailHash;
```
ystros: Since we use the same image for all users without Gravatar enabled, would it make more sense to download that once locally, and then copy it over to the new location as updates are made? The current flow re-downloads it from Gravatar each time.

server/game/gamechat.js
```diff
@@ -22,8 +22,8 @@ class GameChat {
         args = _.reduce(args, (argList, arg) => {
             if(arg instanceof Spectator) {
                 argList.push(arg.name);
-            } else if(arg && arg.emailHash) {
-                argList.push({ name: arg.name, emailHash: arg.emailHash, noAvatar: arg.user.settings.disableGravatar });
+            } else if(arg && arg.name) {
+                argList.push({ name: arg.name });
```
ystros: I get the feeling this will try to format Card arguments as players / avatars, since card objects also have `name`. It might be useful to also add some sort of indicator for the client that it should format this as a name + avatar, rather than just as a object with name. E.g.
```javascript
argList.push({ name: arg.name, argType: 'player' });
```

ystros: Having thought a bit more, it may also be useful to send the URL to the client directly, especially if the rules about transforming usernames to valid filenames (strip this character out, escape this character, handling case sensitivity, etc) becomes complex.

```javascript
argList.push({ name: arg.name, avatarUrl: arg.avatarUrl, argType: 'player' });
```

server/api/account.js
cryogen: usernames must match /^[A-Za-z0-9_-]+$/ - so all good there

server/api/account.js
cryogen: It is slightly wasteful, but at the volumes we're using I think it will be fine

I plan to run a script to update the existing users en masse

server/game/gamechat.js
```diff
@@ -22,8 +22,8 @@ class GameChat {
         args = _.reduce(args, (argList, arg) => {
             if(arg instanceof Spectator) {
                 argList.push(arg.name);
-            } else if(arg && arg.emailHash) {
-                argList.push({ name: arg.name, emailHash: arg.emailHash, noAvatar: arg.user.settings.disableGravatar });
+            } else if(arg && arg.name) {
```
ystros: Just so it isn't lost with the other closed out issues - this check is still overly broad and will cause the client to format cards as if they were players:
screen shot 2018-05-24 at 4 24 09 pm

PR #2006, opened 2018-06-02
Implement more shadow city cards

cryogen:


server/game/cards/11.1-TSC/StormlandsFiefdom.js
```diff
@@ -0,0 +1,23 @@
+const DrawCard = require('../../drawcard.js');
+
+class StormlandsFiefdom extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
```
ystros: Need to also add a call to the `plotModifiers` method to set up the +1 Income on this card (like Roseroad, Northern Armory, etc)

PR #2029, opened 2018-06-09
Implement saves using singular events

ystros:
Previously, we used fake plural events such as onCharactersKilled to
simulate simultaneous resolution. Since then, support has been added for
simultaneous events as well as better UI for choosing which event you
want to match when triggering an ability. It is now possible to
implement saves directly on the singular onCharacterKilled event.

### DIFF
--- a/server/game/cards/03-WotN/JoryCassel.js
+++ b/server/game/cards/03-WotN/JoryCassel.js
@@ -20,7 +20,7 @@ class JoryCassel extends DrawCard {
 
                 context.event.saveCard();
 
-                if(toKill.this.canGainPower && this.game.anyPlotHasTrait('Winter')) {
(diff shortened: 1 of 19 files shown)


server/game/cards/03-WotN/JoryCassel.js
```diff
@@ -5,28 +5,22 @@ class JoryCassel extends DrawCard {
...
+                let message = '{0} uses {1} to save {2}';
+                let toKill = context.event.card;
 
-                context.event.saveCard(toKill);
+                context.event.saveCard();
 
-                if(this.game.getPlayers().some(player => {
-                    return player.activePlot.hasTrait('Winter');
-                }) && toKill.canGainPower()) {
+                if(toKill.this.canGainPower && this.game.anyPlotHasTrait('Winter')) {
```
DukeTax: Just a typo here.

PR #2042, opened 2018-06-16
Implement Growing Ambition

DukeTax:
https://images-cdn.fantasyflightgames.com/filer_public/46/01/4601c97b-a4a4-4065-a69e-3e20cc774561/gt33_a1_growing-ambition.png

### DIFF
--- a/server/game/cards/11.3-SoKL/GrowingAmbition.js
+++ b/server/game/cards/11.3-SoKL/GrowingAmbition.js
@@ -1,0 +1,62 @@
+const DrawCard = require('../../drawcard');
+
+class GrowingAmbition extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Search deck',
+            phase: 'challenge',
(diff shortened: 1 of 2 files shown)


server/game/cards/11.3-SoKL/GrowingAmbition.js
```diff
@@ -0,0 +1,62 @@
...
+            context.player, this, context.opponent, this.chosenCards);
+
+        for(let card of this.chosenCards) {
+            context.player.moveCard(card, 'discard pile');
+        }
+
+        this.game.promptForSelect(context.opponent, {
+            activePromptTitle: `Select ${context.xValue} card(s)`,
+            cardCondition: card => card.controller === this.controller && card.location === 'discard pile',
+            numCards: context.xValue,
```
ystros: This should also be an 'exactly' prompt, shouldn't it? The opponent must select the same number of cards discarded, not just up to the number discarded.

PR #2045, opened 2018-06-20
Implement rival bonuses for Melee

ystros:
Progress toward #1237


cryogen:
This is conflicting now, sorry

PR #2067, opened 2018-06-28
Announce setup cards + clean up pluralization

ystros:


server/game/cards/04.5-GoH/RitualOfRhllor.js
```diff
@@ -15,7 +16,7 @@ class RitualOfRhllor extends DrawCard {
                     mode: 'exactly',
                     numCards: xValue,
                     gameAction: 'gainPower',
-                    activePromptTitle: 'Select ' + (xValue === 1 ? 'a' : xValue) + ' character' + (xValue === 1 ? '' : 's'),
+                    activePromptTitle: `Select ${TextHelper.count(xValue, 'character')}`,
```
ystros: It should be noted that this will change such prompts from 'Select a character' to 'Select 1 character' in the singular case. If we want to keep the original prompt title, I can add a separate helper to distinguish them, but I wasn't sure how much it may matter.

PR #2082, opened 2018-07-04
Implement Serve. Obey. Protect.

DukeTax:
http://lcg-cdn.fantasyflightgames.com/got2nd/GT32_36.jpg

### DIFF
--- a/server/game/cards/11.2-TMoW/ServeObeyProtect.js
+++ b/server/game/cards/11.2-TMoW/ServeObeyProtect.js
@@ -1,0 +1,29 @@
+const DrawCard = require('../../drawcard.js');
+
+class ServeObeyProtect extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onChallengeInitiated: event => event.challenge.defendingPlayer === this.controller &&
+                                               this.controller.getNumberOfChallengesLost('defender') >= 1
+            },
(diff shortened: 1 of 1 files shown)


server/game/cards/11.2-TMoW/ServeObeyProtect.js
```diff
@@ -0,0 +1,29 @@
...
+            when: {
+                onChallengeInitiated: event => event.challenge.defendingPlayer === this.controller &&
+                                               this.controller.getNumberOfChallengesLost('defender') >= 1
+            },
+            cost: ability.costs.returnToHand(card => card.isFaction('martell') && card.getType() === 'character'),
+            target: {
+                cardCondition: (card, context) => card.location === 'hand' && card.controller === this.controller &&
+                                                  card.isFaction('martell') && card.getType() === 'character' &&
+                                                  card.getPrintedCost() <= 5 && 
+                                                  (!context.costs.returnToHand || card !== context.costs.returnToHand)
```
ystros: Should "different character" mean a card with a different name, rather than a card other than the one returned?

DukeTax: Not sure yet, it's been asked to the designers.

We've never had a true definition of different, and there's a bunch of other cards it impacts as well, like "Steel Rain" and the jumping Direwolf attachments like Lady.

ystros: k, will merge for now and we can revisit when there's an official ruling.

PR #2092, opened 2018-07-07
Limit Northern Armory's stand effect to cards in play

DukeTax:


ystros:
Going to be so nice once I get traction w/ game actions and 90% of these checks become obsolete...

PR #2108, opened 2018-07-15
Track which targets have effects applied

### LINKED ISSUE #2098: Lasting effects persisting despite card "leaving play" and returning
Seeing this mainly in Targ decks.

Burn card effects persist on the card despite them leaving play (killed) and being brought back to the board using "Fire and Blood".

Link to video with this interaction: https://youtu.be/TaHtndqcGuY?t=14m46s

Description: Viserion (Core) was killed using burn card, and returned to the board using "Fire and Blood". When declared for a challenge, Viserion was immediately killed again due to 0 Str + terminal burn effect.

Rules from FFG: 

> The phrase “leaves play” refers to any time a card
> makes a transition from an in play state to an out of
> play state.
> If a card leaves play, the following consequences occur
> simultaneously with the card leaving play:
> xxAll tokens on the card are returned to the treasury.
> xxAll duplicates on the card are discarded.
> xxAll attachments that are on the card are returned
> to their owners’ hands.
> **xxAll lasting effects and/or delayed effects affecting**
> **the card while it is in play expire for that card.**

ystros:
Previously, if a target for an effect had immunity to that effect, it
simply wasn't added as a target. In order to handle cases where immunity
wears off after the effect is already applied, it will now be added as a
target but not have the effect applied to it.

Additionally, gaining immunity after a lasting effect has already been
applied will no longer unapply the effect, per RRG rules.

Fixes #2098

### DIFF
--- a/server/game/effect.js
+++ b/server/game/effect.js
@@ -50,6 +50,7 @@ class Effect {
         this.effect = this.buildEffect(properties.effect);
         this.gameAction = this.effect.gameAction || 'genericEffect';
         this.targets = [];
+        this.appliedTargets = new Set();
         this.context = { game: game, source: source };
         this.active = !source.facedown;
         this.isConditional = !!properties.condition || this.targetType === 'player' && _.isFunction(properties.match);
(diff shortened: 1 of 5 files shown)


ystros:
This is more of a bandaid than anything else - the bug was introduced when the interaction between TFM Dany and burn effects was implemented. To summarize:

* If Dany is targeted by Dracarys!, she receives the "is killed if STR is 0" effect, but is immune to the STR reduction. But it is considered to have been applied to her - it just does nothing.
* If she's then blanked, then STR reduction becomes active and she is killed.

To implement this, compound effects were split to individual effects within the effects engine.

This is not strictly correct - if a character is immune to the effect as a whole, they can't be targeted for the effect. But if they're only partially immune, the effect as a whole is applied but the sub-component effect is not applied.

Because the effects were split up into individual components, the engine will thus allow a fully immune target to have a lasting effect applied to them, which could then be activated improperly if immunity is lost. In theory, the target API should prevent most of these situations, but in practice I can't guarantee it.

ystros:
This messes up lasting effects that apply to the currently revealed plot. Back to the drawing board.

PR #2138, opened 2018-07-28
Fix message log for you win or you die

cryogen:


server/game/cards/08.6-SAT/YouWinOrYouDie.js
```diff
@@ -6,7 +6,7 @@ class YouWinOrYouDie extends PlotCard {
         this.whenRevealed({
             handler: () => {
                 let cards = this.controller.drawCardsToHand(2).length;
-                this.game.addMessage('{0} uses {1} to draw {2} to hand',
+                this.game.addMessage('{0} uses {1} to draw {2}',
```
ystros: It might be nicer if it were 'draw {2} cards' so that the card icon appears. We haven't been 100% consistent about it, though, so there are other ability messages that would need to be updated as well.

cryogen: It already says that, using your helper :)

ystros: Oh, right. :-)

(Can't merge cause Jenkins is stuck?)

PR #2142, opened 2018-08-04
Implement characters from Kings of the Isles announcement article

DukeTax:
https://www.fantasyflightgames.com/en/news/2018/7/30/king-of-the-isles/

### DIFF
--- a/server/game/cards/12-KotI/BalonGreyjoy.js
+++ b/server/game/cards/12-KotI/BalonGreyjoy.js
@@ -1,0 +1,29 @@
+const DrawCard = require('../../drawcard.js');
+
+class BalonGreyjoy extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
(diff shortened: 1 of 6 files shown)


server/game/cards/12-KotI/BalonGreyjoy.js
```diff
@@ -0,0 +1,29 @@
+const DrawCard = require('../../drawcard.js');
+
+class BalonGreyjoy extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Put card into play',
+            cost: ability.costs.kneelFactionCard(),
+            target: {
+                cardCondition: card => card.location === 'discard pile' && card.controller !== this.controller &&
+                                       this.controller.canPutIntoPlay(card)
```
ystros: As discussed in chat, this may not actually be the case, but this should be fine until we get a clear ruling one way or the other.

PR #2144, opened 2018-08-08
Add card implementations for first half of Streets of King's Landing

DukeTax:
http://www.cardgamedb.com/index.php/agameofthrones2ndedition/a-game-of-thrones-2nd-edition-cards/_/dance-of-shadows-cycle/streets-of-king-s-landing/


ystros:
Try rebasing vs master since Jenkins never came back when the PR was opened. The only conflict I think will be on the submodule, which should be easy to resolve (your change has the full data vs the one in master just has Valyrian Crew).

PR #2158, opened 2018-08-17
Fix pending game chat

cryogen:
Set the arg type for lobby players so that pending game chat works properly


ystros:
Feels like there should be a more elegant fix here, but that would require serious delving into the message system and this should just work.

PR #2167, opened 2018-08-21
Check maximums for non-event abilities

ystros:
Previously, only abilities on event cards were being checked for
maximums, and it was being done as a cost. This led to a bug where the
max on character / attachment abilities would not be properly enforced.
The timing would also have been incorrect in the case of the cost being
cancelled (Drowned God Fanatic).

Now, maximums are enforced when requirements for the ability are
checked, and marked as used at the same time as limits. This should
ensure that Drowned God Fanatic is properly limited to once per round.


mplain:
Please note that limits and maxima should be marked as used only if the initiation sequence gets to step 6 "initiate effects". If an ability is aborted during the payment of costs or choosing targets, the limits and maxima are not used.

ystros:
Correct - limits are already marked as used at step 6 in the code. This PR moves using maxima to the same place. So saving vs costs + limits or maxima will have the correct behavior (once we allow saves vs costs)

PR #2175, opened 2018-08-29
Drowned God Fanatic: ensure only character can be killed as costs

### LINKED ISSUE #2173: Drowned God Fanatic kill cost
When I trigger Fanatic in my dead pile, it highlights my locations in play as potential targets for his kill cost.

zacchiro:
Fixes #2173


server/game/cards/11.3-SoKL/DrownedGodFanatic.js
```diff
@@ -7,7 +7,7 @@ class DrownedGodFanatic extends DrawCard {
             title: 'Put Drowned God Fanatic into play',
             condition: () => this.controller.canPutIntoPlay(this),
             location: 'dead pile',
-            cost: ability.costs.kill(),
+            cost: ability.costs.kill(card => card.getType() === 'character'),
```
ystros: This should be fixed by modifying `/server/game/costs/KillCost.js` instead, since you cannot ever kill locations. Otherwise, we'll need to specify this on every single usage of the kill cost in the future.

zacchiro: *nod*.

FWIW I did ask about that on discord/IRC, but sent the PR this way as it's consistent with how the only other card needing this (Given to the Drowned God) is currently doing.

I'll update the PR to fix `KillCost`.

PR #2179, opened 2018-08-30
Manual dupes

### LINKED ISSUE #992: prompt before saving a character using a duplicate
Just wanted to put in a request for a prompt on whether or not a dupe is used to save a character. A relevant scenario would be if I have a duped Jorah with 2 betrayal tokens on him and I used Waking the Dragon to stand him during the challenge phase. At the end of the phase I actually want the Waking the Dragon to send him back to my hand. Thanks!

### LINKED ISSUE #1774: Dupes on cards that are controlled by an opponent should still be usable by the owner
as per the rules

cryogen:
Make use of the player setting for whether or not dupes are prompted instead of being automatic

Fixes #992
Fixes #1774

### DIFF
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -149,6 +149,10 @@ const Costs = {
         };
     },
     /**
+     * Cost that requires discarding a duplicate.
+     */
+    discardDuplicate: () => CostBuilders.discardDuplicate.self(),
+    /**
      * Cost that requires discarding from hand the card that initiated the ability.
      */
     discardSelfFromHand: () => CostBuilders.discardFromHand.self(),
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -57,6 +57,10 @@ const Costs = {
      */
     sacrificeSelf: () => CostBuilders.sacrifice.self(),
     /**
+     * Cost that will sacrifice a specific card
+     */
+    sacrificeCard: match => CostBuilders.sacrifice.specific(match),
+    /**
      * Cost that requires sacrificing a card that matches the passed condition
      * predicate function.
      */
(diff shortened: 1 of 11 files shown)


server/game/costs.js
```diff
@@ -144,6 +144,10 @@ const Costs = {
             }
         };
     },
+    /**
+     * Cost that requires discarding a duplicate.
+     */
+    discardDuplicate: () => CostBuilders.discardDuplicate.self(),
```
ystros: Naming wise this cost name is somewhat inconsistent - it would normally end with 'self'. However, since it's likely only ever going to be used in this single case, and there probably won't be a general "choose and discard a duplicate to X" cost, I'm not too bothered.

server/game/drawcard.js
```diff
@@ -58,9 +61,32 @@ class DrawCard extends BaseCard {
...
         clone.traits = this.traits.clone();
+
         return clone;
     }
 
+    setupDuplicateAbility(ability) {
+        let dupeCondition = event => event.card === this.parent && this.parent.canBeSaved() && event.allowSave && this.parent.owner.promptDupes;
+
+        this.interrupt({
+            canCancel: true,
```
ystros: Forgot one thing in my review - this interrupt also needs `cannotBeCanceled: true` to prevent things like Treachery or Fear Cuts Deeper Than Swords from triggering inappropriately.

server/game/costs.js
cryogen: I can change it, but it I'm not sure what I'd change it to :)

cryogen:
It's actually easy to do what you've suggested, all you need to do is remove the check altogether.  If it can be saved automatically it will be and then won't be prompted.

However, this breaks the tests, and I ran out of time trying to fix them :)

I'll take another look at some point

PR #2187, opened 2018-09-04
Implement Stark and Tyrell cards for Music of Dragons

ystros:


server/game/cards/11.4-MoD/MeeraReed.js
```diff
@@ -0,0 +1,41 @@
...
+
+class MeeraReed extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onSacrificed: event => this.isControlledStarkCharacter(event.cardStateWhenSacrificed)
+            },
+            handler: context => {
+                this.game.addMessage('{0} uses {1} to return {1} to shadows', context.player, this);
+                context.player.putIntoShadows(this, false);
```
DukeTax: I struggle to think of a reason why, but I think you should be able to save versus this if you want. Unless this is on purpose to diminish (likely useless) prompts, which is fine with me.

server/game/cards/11.4-MoD/MeeraReed.js
```diff
@@ -0,0 +1,41 @@
...
+            }
+        });
+
+        this.reaction({
+            when: {
+                onCardOutOfShadows: event => event.card === this
+            },
+            target: {
+                activePromptTitle: 'Select a character',
```
DukeTax: Just an FYI: this is the default.

server/game/cards/11.4-MoD/BondsOfChivalry.js
```diff
@@ -0,0 +1,49 @@
+const DrawCard = require('../../drawcard');
+
+class BondsOfChivalry extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Stand and remove Knight',
+            condition: () => this.game.isDuringChallenge(),
+            target: {
+                cardCondition: card => card.controller === this.controller && card.getType() === 'character' && card.isParticipating()
```
DukeTax: Needs to be a Knight.

server/game/cards/11.4-MoD/MeeraReed.js
ystros: The closest scenario I can think of is if you duped Meera, then the opponent steals Meera, and tries to trigger this ability. You might (maybe??) want to save in that scenario. But since she would be returned to your shadows, the opponent probably doesn't want to do that in the first place.

Mostly the cannot save here is to prevent automatic dupes from keeping Meera in play after the player explicitly triggers an ability to get her out of play.

server/game/cards/11.4-MoD/MeeraReed.js
ystros: Thanks! I actually looked it up while implementing Baratheon cards earlier. I had thought "Select a card" was the default for some reason. Gotta shake off that rust in my brain after not implementing cards for so long. 😄

PR #2193, opened 2018-09-05
Add greyjoy cards from MoD

cryogen:
Also update card data submodule

### DIFF
--- a/server/game/cards/11.4-MoD/GreyGhost.js
+++ b/server/game/cards/11.4-MoD/GreyGhost.js
@@ -1,0 +1,63 @@
+const DrawCard = require('../../drawcard');
+
+class GreyGhost extends DrawCard {
+    constructor(owner, cardData) {
+        super(owner, cardData);
+
+        this.tracker = new ShadowsTracker(this.game, this);
+    }
+
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Choose character(s)',
+            cost: ability.costs.kneelSelf(),
+            handler: () => {
+                this.game.promptForSelect(this.controller, {
+                    numCards: this.tracker.hasComeOutOfShadowsThisPhase ? 2 : 1,
+                    activePromptTitle: this.tracker.hasComeOutOfShadowsThisPhase ? 'Select 2 characters' : 'Select a character',
+                    source: this,
+                    cardCondition: card => card.location === 'play area' && card.getType() === 'character',
+                    onSelect: (player, cards) => this.onSelect(player, cards),
+                    onCancel: (player) => this.cancelSelection(player)
+                });
+            }
+        });
+    }
+
+    onSelect(player, cards) {
+        if(!Array.isArray(cards)) {
+            cards = [cards];
+        }
+
+        this.untilEndOfPhase(ability => ({
+            match: card => cards.some(c => card === c),
+            targetController: 'opponent',
+            effect: ability.effects.cannotBeDeclaredAsDefender()
+        }));
+
+        this.game.addMessage('{0} kneels {1} to make {2} unable to be declared as {3}',
+            this.controller, this, cards, cards.length === 1 ? 'a defender' : 'defenders');
+        return true;
+    }
+
+    cancelSelection(player) {
+        this.selections.push({ player: player, cards: [] });
+        this.proceedToNextStep();
+    }
+}
+
+class ShadowsTracker {
+    constructor(game, card) {
+        this.hasComeOutOfShadowsThisPhase = false;
+        game.on('onCardOutOfShadows', event => {
+            if(event.card === card) {
+                this.hasComeOutOfShadowsThisPhase = true;
+            }
+        });
+        game.on('onPhaseEnded', () => this.hasComeOutOfShadowsThisPhase = false);
+    }
+}
+
+GreyGhost.code = '11072';
+
+module.exports = GreyGhost;
--- a/server/game/cards/11.4-MoD/GreyGhost.js
+++ b/server/game/cards/11.4-MoD/GreyGhost.js
@@ -13,6 +13,7 @@ class GreyGhost extends DrawCard {
             cost: ability.costs.kneelSelf(),
             handler: () => {
                 this.game.promptForSelect(this.controller, {
+                    multiselect: true,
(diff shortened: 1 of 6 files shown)


server/game/cards/11.4-MoD/GreyGhost.js
```diff
@@ -0,0 +1,63 @@
...
+class GreyGhost extends DrawCard {
+    constructor(owner, cardData) {
+        super(owner, cardData);
+
+        this.tracker = new ShadowsTracker(this.game, this);
+    }
+
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Choose character(s)',
```
DukeTax: According to convention, titles are after the effect, so something like 'Make character(s) unable to defend', or maybe something less awkward than that. :)

server/game/cards/11.4-MoD/GreyGhost.js
```diff
@@ -0,0 +1,63 @@
...
+            cost: ability.costs.kneelSelf(),
+            handler: () => {
+                this.game.promptForSelect(this.controller, {
```
DukeTax: This should be done via the target API (which makes sure that 1) the ability can't be initiated if there are no valid targets, and 2) that targets are chosen and communicated to the opponent before they decide whether to cancel).

server/game/cards/11.4-MoD/GreyGhost.js
```diff
@@ -0,0 +1,63 @@
...
+                    cardCondition: card => card.location === 'play area' && card.getType() === 'character',
+                    onSelect: (player, cards) => this.onSelect(player, cards),
+                    onCancel: (player) => this.cancelSelection(player)
+                });
+            }
+        });
+    }
+
+    onSelect(player, cards) {
+        if(!Array.isArray(cards)) {
```
DukeTax: If you add `multiselect: true` to the target definition, `cards` (or `context.target` if done via the target API) will be an array irrespective of the number of targets chosen.

server/game/cards/11.4-MoD/GreyGhost.js
```diff
@@ -0,0 +1,63 @@
...
+            targetController: 'opponent',
+            effect: ability.effects.cannotBeDeclaredAsDefender()
+        }));
+
+        this.game.addMessage('{0} kneels {1} to make {2} unable to be declared as {3}',
+            this.controller, this, cards, cards.length === 1 ? 'a defender' : 'defenders');
+        return true;
+    }
+
+    cancelSelection(player) {
```
DukeTax: Method needs to contain an alert message of some sort instead of this.

server/game/cards/11.4-MoD/VictarionGreyjoy.js
```diff
@@ -0,0 +1,72 @@
...
+                });
+            }
+        });
+    }
+
+    kneelLocation() {
+        this.game.promptForSelect(this.controller, {
+            activePromptTitle: 'Select a location',
+            source: this,
+            cardCondition: card => card.getType() === 'location',
```
DukeTax: Needs an added `card.location === 'play area'`.

server/game/cards/11.4-MoD/VictarionGreyjoy.js
```diff
@@ -0,0 +1,72 @@
...
+            onCancel: (player) => this.cancelSelection(player)
+        });
+
+        return true;
+    }
+
+    standVictarion(player) {
+        this.controller.standCard(this);
+
+        this.game.addMessage('{0} uses {1} to stand {1}', player, this);
```
DukeTax: Per convention, messages contain costs, so in this case `{0} uses {1} and kneels {2} to stand {1}`, with `{2}` being `this.context.costs.kneel`.

server/game/cards/11.4-MoD/VictarionGreyjoy.js
```diff
@@ -0,0 +1,72 @@
...
+
+        return true;
+    }
+
+    gainRenown() {
```
DukeTax: Needs a message.

server/game/cards/11.4-MoD/VictarionGreyjoy.js
```diff
@@ -0,0 +1,72 @@
...
+            effect: ability.effects.addKeyword('renown')
+        }));
+
+        return true;
+    }
+
+    onLocationSelected(player, card) {
+        card.controller.kneelCard(card);
+
+        this.game.addMessage('{0} uses {1} to kneel {2}', player, this, card);
```
DukeTax: Same as above.

server/game/cards/11.4-MoD/VictarionGreyjoy.js
```diff
@@ -0,0 +1,72 @@
...
+
+        return true;
+    }
+
+    cancelSelection(player) {
+        this.game.addMessage('{0} cancels the resolution of {1}', player, this);
```
DukeTax: Probably better if this is an alert message.

And not 100% sure but probably needs a `return true;` as well to close the prompt.

server/game/cards/11.4-MoD/VictarionGreyjoy.js
```diff
@@ -0,0 +1,72 @@
+const DrawCard = require('../../drawcard');
+
+class VictarionGreyjoy extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kneel a warship location',
```
DukeTax: Normally titles are after the effect, but in this case this is probably better.

server/game/cards/11.4-MoD/GreyGhost.js
cryogen: I couldn't make it any less awkward, which is why i went this way - open to suggestions

server/game/cards/11.4-MoD/GreyGhost.js
cryogen: Oops, copy/pasta ftw

cryogen: Also, this is horribly inconsistent throughout the card base.. definitely needs to get fixed

PR #2200, opened 2018-09-08
Fix crash + language on Growing Ambition

ystros:
Cleans up language of the prompt for Growing Ambition after converting
it to a card select prompt, as well as adding messages for cancellation
cases and removing the call to a cleanup method that was removed.

Fixes THRONETEKI-2AJ

### DIFF
--- a/server/game/cards/11.3-SoKL/GrowingAmbition.js
+++ b/server/game/cards/11.3-SoKL/GrowingAmbition.js
@@ -12,23 +12,16 @@ class GrowingAmbition extends DrawCard {
 
                 this.game.promptForDeckSearch(this.controller, {
                     numToSelect: context.xValue,
-                    onSelect: (player, cards) => this.selectCards(player, cards, context),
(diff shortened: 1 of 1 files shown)


server/game/cards/11.3-SoKL/GrowingAmbition.js
```diff
@@ -42,11 +35,16 @@ class GrowingAmbition extends DrawCard {
...
             source: this,
-            onSelect: (player, cards) => this.onCardsSelected(player, cards)
+            onSelect: (player, cards) => this.opponentSelectCards(player, cards),
+            onCancel: (player) => this.opponentCancel(player)
         });
     }
 
-    onCardsSelected(player, cards) {
+    selfCancelSearch(context) {
+        this.game.addMessage('{0} play {1} to search their deck, but chooses no cards', context.player, this);
```
cryogen: Typo: play/plays

PR #2204, opened 2018-09-12
Allow dupes when card abilities prevented

ystros:
Using a dupe to save is considered a "game ability" instead of a "card
ability." Since manual dupes are set up as using the ability system,
they were improperly considered a "card ability" when cards like
Winterfell, Core Catelyn, or King in the North were in play.

Now, abilities can be defined as a game ability and thus skip the checks
involved with effects preventing card abilities.

### DIFF
--- a/server/game/PlayActions/MarshalCardAction.js
+++ b/server/game/PlayActions/MarshalCardAction.js
@@ -4,6 +4,7 @@ const Costs = require('../costs');
 class MarshalCardAction extends BaseAbility {
     constructor() {
         super({
+            abilitySourceType: 'game',
             cost: [
                 Costs.payReduceableGoldCost('marshal'),
                 Costs.playLimited()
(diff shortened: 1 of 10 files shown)


server/game/PlayActions/MarshalCardAction.js
```diff
@@ -4,6 +4,7 @@ const Costs = require('../costs');
 class MarshalCardAction extends BaseAbility {
     constructor() {
         super({
+            abilitySourceType: 'game',
```
ystros: I'd be open to alternate names for this property. It's meant to distinguish between 'game abilities' and 'card abilities.' The name `abilityType` is already used to distinguish between the various ability timings (interrupt, reaction, forced interrupt, etc). Perhaps that itself should be renamed, but even so it isn't very descriptive for this situation. RRG doesn't really offer any guidance on language here either (only one mention of "game ability" and it's in reference to duplicates).

PR #2211, opened 2018-09-17
Allow ability messages defined outside of handler

ystros:
Adds a `message` property to triggered abilities in order to allow
messages to be output before the handler for the ability executes. This
will allow messages to be output to the game log before players are
prompted for cancel abilities (Treachery, Hand's Judgment, etc). It will
also enable the game action mechanism in the future, which will require
removing / replacing the existing `handler` property for abilities.

### DIFF
--- a/docs/implementing-cards.md
+++ b/docs/implementing-cards.md
@@ -316,6 +316,35 @@ class SealOfTheHand extends DrawCard {
 }
 ```
 
+#### Ability messages
+
+The `message` property can be used to add a message to the game log outside of the `handler` function. By separating the message from the handler that executes the ability, messages can be added to the game log prior to prompting to cancel abilities (e.g. Treachery, Hand's Judgment, etc).
+
+The `message` property should be a function that takes an ability context object and returns an array with the message string as the first item in the array, followed by any arguments for the message:
+
+```javascript
+this.action({
+    // ...
+    message: context => ['{0} kneels {1} to stand {2}', context.player, this, this.parent],
+    handler: () => {
+        // ...
+    }
+});
+```
+
+It can also output the message directly using the game object without returning a message format:
+```javascript
+this.action({
+    // ...
+    message: context => {
+        this.game.addMessage('{0} kneels {1} to stand {2}', context.player, this, this.parent);
+    },
+    handler: () => {
+        // ...
+    }
+});
+```
+
 #### Checking ability restrictions
 
 Card abilities can only be triggered if they have the potential to modify game state (outside of paying costs). To ensure that the action's play restrictions are met, pass a `condition` function that returns `true` when the restrictions are met, and `false` otherwise. If the condition returns `false`, the action will not be executed and costs will not be paid.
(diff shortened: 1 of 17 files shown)


ystros:
More background around this change: the upcoming game actions will stop using the `handler` property directly. So something like this:
```javascript
        this.action({
            title: 'Stand attached character',
            condition: () => this.parent.kneeled,
            handler: () => {
                this.controller.standCard(this.parent);
                this.game.addMessage('{0} kneels {1} to stand {2}', this.controller, this, this.parent);
            }
        });
```

will look more like:
```javascript
        this.action({
            title: 'Stand attached character',
            condition: () => this.parent.kneeled,
            gameAction: ability.actions.stand(context => ({ card: this.parent }))
        });
```

This obviously removes the usual place we do game log messages for abilities. While I'd like to auto-generate messages, that would require switching *everything* over to game actions, a much larger change than gradually introducing them. Even if auto-generating messages works well in most cases, we'll still probably want to be able to override them in certain cases.

Given the above, it seemed like the best way would be to introduce a new property that allows you to generate the message separate from the handler. e.g.

```javascript
        this.action({
            title: 'Stand attached character',
            condition: () => this.parent.kneeled,
            message: context => ['{0} kneels {1} to stand {2}', context.player, this, this.parent],
            gameAction: ability.actions.stand(context => ({ card: this.parent }))
        });
```

As a bonus for separating out the message, it can now be added to the game log prior to the handler executing, and thus be available in the log when a player is prompted for Treachery / Hand's Judgment / other cancellation abilities.

There is one slight snag that will need to be solved - our messages about gaining gold and drawing cards gives the exact amount gained / drawn rather than what the ability specifies. This seems inconsistent with the rest of our messages. e.g. if an ability says to kill a character, we always add a message saying 'Player 1 uses X to kill Y' even if Y is saved via an ability or dupe. That doesn't work as well for gold gained / cards drawn but we might be able to come up with something ('Player 1 uses Counting Coppers to draw 3 cards.' 'Player 1 only draws 1 card due to reaching card draw maximum')

Thoughts about this new property or the message inconsistency?

docs/implementing-cards.md
```diff
@@ -316,6 +316,35 @@ class SealOfTheHand extends DrawCard {
...
+#### Ability messages
+
+The `message` property can be used to add a message to the game log outside of the `handler` function. By separating the message from the handler that executes the ability, messages can be added to the game log prior to prompting to cancel abilities (e.g. Treachery, Hand's Judgment, etc).
+
+The `message` property should be a function that takes an ability context object and returns an array with the message string as the first item in the array, followed by any arguments for the message:
+
+```javascript
+this.action({
+    // ...
+    message: context => ['{0} kneels {1} to stand {2}', context.player, this, this.parent],
```
cryogen: I'm not a huge fan of it needing to be an array, but I can't think of a better way..

`message: context => { format: 'blah blah', args: [1, 2, 3] }`

*might* be cleaner and more intuitive, but that's debatable.

cryogen:
I started out with messages stating exactly what the card ability said.  It was changed later, to give actual figures, as it was considered helpful.  I'm not against it going back to the original card ability text.  I think it would be messy with these additional clarification messages.

ystros:
@cryogen Updated with a new direction - more structure when there are additional arguments, as well as introducing named arguments within the format for clarity. I like the shorthand (with named arguments) when you don't need any additional arguments. When you need arguments it becomes a bit verbose, but it's a lot clearer what's going on. Not sure whether the verbosity vs clarity trade-off here leans too much in one direction.

Let me know what you think!

PR #2218, opened 2018-09-20
Always prompt for dupes on stolen characters

### LINKED ISSUE #2214: Warded Character with Duplicate Died
I duplicated my Septon Celledor on setup. Turn 1 my opponent attached Ward and took control of him. There were no Military challenges, so he had the duplicate into the 2nd round. When my opponent played Valar Morghulis 2nd round, Celledor died despite having a duplicate.

Per this thread (http://www.cardgamedb.com/forums/index.php?/topic/28627-taking-control-and-duplicates/) the owner of the card still controls the duplicates, so I should've been able to save him (and then return him to my control when I Confiscated the Ward).

ystros:
Prior to implementing manual dupes, there was no way for a player to use
the dupes on a character that was stolen from them. After manual dupes
was implemented, they could save stolen character if they had the manual
dupe option turned on. However, using automatic dupes still left no
option to save stolen characters.

Now, players will be prompted whether they want to use the dupe or pass
on a stolen character, regardless of whether they have the manual dupe
option turned on.

Fixes #2214

### DIFF
--- a/server/game/drawcard.js
+++ b/server/game/drawcard.js
@@ -66,7 +66,7 @@ class DrawCard extends BaseCard {
     }
 
     setupDuplicateAbility(ability) {
-        let dupeCondition = event => event.card === this.parent && this.parent.canBeSaved() && event.allowSave && this.controller.promptDupes;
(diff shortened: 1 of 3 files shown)


cryogen:
Genuinely thought I'd already done this :)

PR #2279, opened 2018-10-12
Implement Cunning Diplomat - Tyrell card from Daggers in the Dark 

hrnciar:
Cunning Diplomat(my translation, may be different); 3 cost; non-unique; non-loyal; Ally, Spy; Power;
Ambush (5), Shadows(3)
Reaction: After Cunning Diplomat enters play during the challenge, choose participating character. Stand that character and remove it from challenge.

### DIFF
--- a/server/game/cards/11.6-DitD/CunningDiplomat.js
+++ b/server/game/cards/11.6-DitD/CunningDiplomat.js
@@ -1,0 +1,25 @@
+class CunningDiplomat extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardEntersPlay: event => (
+                    event.card === this &&
+                    this.game.isDuringChallenge()
+                )
+            },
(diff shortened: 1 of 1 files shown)


server/game/cards/11.6-DitD/CunningDiplomat.js
```diff
@@ -0,0 +1,27 @@
...
+                )
+            },
+            target: {
+                cardCondition: card => card.getType() === 'character' && card.isParticipating()
+            },
+            handler: context => {
+                context.target.controller.standCard(context.target);
+                this.game.currentChallenge.removeFromChallenge(context.target);
+
+                this.game.addMessage('{0} uses {1} to remove {2} from the challenge', context.player, this, context.target);
```
ystros: The message should probably mention that the targeted character is stood, e.g. "to stand and remove {2}"

PR #2282, opened 2018-10-15
Implement Arbor Market - Tyrell card from Daggers in the Dark 

hrnciar:
Arbor Market(my translation, may be different); 2 cost; non-unique; loyal; House Redwyne; 
Challenges phase action: Kneel Arbor Market to choose participating character. That character gains +1 STR until the end of challenge.


mplain:
Might I ask, where is this from?

hrnciar:
@mplain I am the one who wanted to spoil Daggers on FB. I did not do it after reaction and instead I decided to help community with implementing some easy cards so it could be released when the time comes. Only very few people will notice it here.

mplain:
oh =_= got it, sorry

PR #2301, opened 2018-10-27
Add quick join functionality

cryogen:


views/layout.pug
```diff
@@ -20,6 +20,4 @@ html
       })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
       ga('create', 'UA-75043688-2', 'auto');
-      ga('send', 'pageview');
-
-</script>
\ No newline at end of file
+      ga('send', 'pageview');
```
ystros: 👍

server/lobby.js
```diff
@@ -392,6 +392,29 @@ class Lobby {
             return;
         }
 
+        if(gameDetails.quickJoin) {
+            let sortedGames = _.sortBy(this.games, 'createdAt');
+            let gameToJoin = sortedGames.find(game => !game.started && game.gameType === gameDetails.gameType && _.size(game.players) < 2 && !game.password);
```
ystros: This is probably fine for now but it might be nice to encapsulate the logic on the game object (particularly for Melee purposes where that size check wouldn't be correct)

server/lobby.js
```diff
@@ -392,6 +392,29 @@ class Lobby {
...
+
+            if(gameToJoin) {
+                gameToJoin.join(socket.id, socket.user.getDetails(), undefined, (err, message) => {
+                    if(err) {
+                        socket.send('passworderror', message);
```
ystros: This doesn't seem like the proper message to send back since passworded games were explicitly excluded in the `gameToJoin` array.

server/lobby.js
cryogen: Yes, this needs to be game.canJoin(), but that refactor is too big for this small change right now

server/lobby.js
cryogen: There is no generic error mechanism, this check is actually redundant as the only exceptions thrown are related to password checking.  However, I'll leave this in for now as we need a generic error mechanism to prevent things like a bug Zack just reported with joining blocked games

PR #2316, opened 2018-11-06
Implement great wyk

luky92:


ystros:
I'm going to close this one for now since you also included the implementation in #2318.

PR #2318, opened 2018-11-08
Implement when i woke

luky92:
Since i used automatic spacing fix it also fixed the comma thats why the file is there if thats a problem i can create another or for this

### DIFF
--- a/server/game/cards/12-KotI/GreatWyk.js
+++ b/server/game/cards/12-KotI/GreatWyk.js
@@ -1,0 +1,47 @@
+const DrawCard = require('../../drawcard.js');
+
+class GreatWyk extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction ({
+            when: {
+                onCardEntersPlay: event=>(
+                    event.card.getType() === 'character' &&
+                    event.card.controller === this.controller &&
+                    event.originalLocation === 'dead pile'
+                )}
+            ,
+            limit: ability.limit.perPhase(1),
+            handler: () => {
+                 
+                for(let opponent of this.game.getOpponents(this.controller)) {
+                    if(opponent.hand.length > 0) {
+                        this.game.promptForSelect(opponent, {
+                            activePromptTitle: 'Select a card',
+                            source: this,
+                            cardCondition: card => opponent.hand.includes(card),
+                            gameAction: 'discard',
+                            onSelect: (opponent, card) => this.onSelectCard(opponent, card),
+                            onCancel: (opponent) => this.onSelectCard(opponent, null)
+                        });
+                    }
+                }        
+            }
+        });
+
+    }
+    onSelectCard(player,card) {
+        if(card === null) {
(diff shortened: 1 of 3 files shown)


server/game/cards/12-KotI/WhenIWoke.js
```diff
@@ -0,0 +1,40 @@
+const DrawCard = require('../../drawcard.js');
+
+class WhenIWoke extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                afterChallenge: event => event.challenge.challengeType === 'military' && event.challenge.loser !== undefined
```
ystros: Checking that a `loser` is truthy is a little more idiomatic for Javascript. It would also cover any case where `loser` was `null` instead of `undefined`

```suggestion
                afterChallenge: event => event.challenge.challengeType === 'military' && event.challenge.loser
```

or possibly:

```suggestion
                afterChallenge: event => event.challenge.challengeType === 'military' && !!event.challenge.loser
```

server/game/cards/12-KotI/GreatWyk.js
```diff
@@ -0,0 +1,47 @@
...
+
+class GreatWyk extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction ({
+            when: {
+                onCardEntersPlay: event=>(
+                    event.card.getType() === 'character' &&
+                    event.card.controller === this.controller &&
+                    event.originalLocation === 'dead pile'
+                )}
```
ystros: Spacing here is not quite in convention with the rest of the codebase. The closing `}` should come at the same indentation as the `when`. So more like:
```javascript
            when: {
                onCardEntersPlay: event => (
                    event.card.getType() === 'character' &&
                    event.card.controller === this.controller &&
                    event.originalLocation === 'dead pile'
                )
            },
```

server/game/cards/12-KotI/GreatWyk.js
```diff
@@ -0,0 +1,47 @@
...
+                            onSelect: (opponent, card) => this.onSelectCard(opponent, card),
+                            onCancel: (opponent) => this.onSelectCard(opponent, null)
+                        });
+                    }
+                }        
+            }
+        });
+
+    }
+    onSelectCard(player,card) {
```
ystros: Add a new line between the closing of one method and the start of another.

server/game/cards/12-KotI/GreatWyk.js
```diff
@@ -0,0 +1,47 @@
...
+        }
+        
+        player.discardCard(card);
+        this.game.addMessage('{0} chooses {1} for {2}', player, card, this);
+        return true;
+
+    }
+
+}
+GreatWyk.code = '12017';
```
ystros: Other card definitions have a new line between the end of the class and the code being set, and between the code being set and the `module.exports` line.

cryogen:
Yeah, not sure eslint rules exist for those things.  I'll check.  Particularly keen on not having too much vertical space around braces, it's unnecessary.

PR #2320, opened 2018-11-12
Implement We Take Westeros and What is Dead May Never Die

luky92:


server/game/cards/12-KotI/WhatIsDeadMayNeverDie.js
```diff
@@ -0,0 +1,33 @@
+const DrawCard = require('../../drawcard.js');
+
+class WhatIsDeadMayNeverDie extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onDominanceDetermined: event => this.controller === event.winner
+            },
+            target: {
+                activePromptTitle: 'Select character',
```
ystros: This title isn't necessary because it's the default for the target API.

server/game/cards/12-KotI/WhatIsDeadMayNeverDie.js
```diff
@@ -0,0 +1,33 @@
...
+                cardCondition: card => this.cardCondition(card)
+            },
+            handler: context => {
+                context.player.putIntoPlay(context.target);
+                this.game.addMessage('{0} uses {1} to put {2} into play from their dead pile', context.player, this, context.target);
```
ystros: This is a really small thing, but the general convention we've used is `{0} plays {1}` for event cards instead of `{0} uses {1}` which is used for everything else.

server/game/cards/12-KotI/WeTakeWesteros.js
```diff
@@ -0,0 +1,20 @@
...
+class WeTakeWesteros extends PlotCard {
+    setupCardAbilities() {
+        this.whenRevealed({
+            target: {
+                activePromptTitle: 'Select a card',
+                cardCondition: card => card.getType() === 'location' && card.location === 'discard pile' && this.controller.canPutIntoPlay(card)
+            },
+            handler: (context) => {
+                context.player.putIntoPlay(context.target);
+                this.game.addMessage('{0} uses {1} to put {2} into play ',context.player,this,context.target);
```
ystros: While we generally try to avoid verbose messages, it might be useful to add whose discard pile the card came from. i.e.
```suggestion
                this.game.addMessage('{0} uses {1} to put {2} into play from {3}\'s discard pile', context.player, this, context.target, context.target.owner);
```

PR #2336, opened 2018-12-01
Implement some cards from the greyjoy box

cryogen:


server/game/chatcommands.js
```diff
@@ -10,6 +11,7 @@ class ChatCommands {
             '/add-icon': this.addIcon,
             '/add-keyword': this.addKeyword,
             '/add-trait': this.addTrait,
+            '/add-card': this.addCard,
```
ystros: 🎉

PR #2337, opened 2018-12-01
Backport various fixes from keyteki

cryogen:


server/pendinggame.js
```diff
@@ -100,34 +100,19 @@ class PendingGame {
...
-                this.password = hash;
-                this.addPlayer(id, user);
-
-                callback();
-            });
-        } else {
-            this.addPlayer(id, user);
-
-            callback();
+            this.password = crypto.createHash('md5').update(password).digest('hex');
```
ystros: Have to admit, even though this is ephemeral (only lasts as long as the individual game), it makes me nervous (perhaps irrationally). What spurred dropping bcrypt here?

cryogen: To get rid of it being async.  It fixes the 'can join multiple games at once' bug.  It could as well store the plain text, it's really just a shared key between two people, and not meant to be protecting anything secure

PR #2344, opened 2018-12-02
More greyjoy box cards

cryogen:


server/game/cards/12-KotI/DesertRaider.js
```diff
@@ -0,0 +1,52 @@
...
+
+class DesertRaider extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                afterChallenge: event => event.challenge.loser === this.controller && this.controller.canPutIntoPlay(this)
+            },
+            location: 'dead pile',
+            target: {
+                type: 'select',
```
ystros: The card explicitly says choose, so this isn't needed. `type: 'select'` is more for implicit targeting (e.g. "Place a poison token on a character without intrigue icon")

server/game/cards/12-KotI/DesertRaider.js
```diff
@@ -0,0 +1,52 @@
...
+                cardCondition: card => card.location === 'play area' && card.getType() === 'character' &&
+                                       card.isParticipating() && card.canGainPower() && this.game.currentChallenge.winner === card.controller
```
ystros: Card text doesn't require the character to be participating in the challenge, just controlled by the winning opponent. It probably also doesn't need to be able to actually gain the power - it reads like a dual effect - gains 1 power AND loses an icon. As long as one can happen, you should be able to target the character.

server/game/cards/12-KotI/DrownedProphet.js
```diff
@@ -0,0 +1,35 @@
...
+    setupCardAbilities(ability) {
+        this.interrupt({
+            when: {
+                onCharacterKilled: event => event.card === this
+            },
+            limit: ability.limit.perPhase(1),
+            handler: () => {
+                this.game.promptForDeckSearch(this.controller, {
+                    numCards: 5,
+                    cardCondition: card => card.isFaction('greyjoy'),
```
ystros: Specifically needs to be a Greyjoy character.

server/game/cards/12-KotI/GunthorSonOfGurn.js
```diff
@@ -0,0 +1,35 @@
...
+        this.reaction({
+            when: {
+                afterChallenge: event => (
+                    event.challenge.attackingPlayer === this.controller &&
+                    event.challenge.winner === this.controller &&
+                    this.isAttacking()
+                )
+            },
+            handler: context => {
+                this.game.promptForSelect(context.opponent, {
```
ystros: I'm not sure `context.opponent` will automatically be filled in here (though maybe it should). I think it's exclusive to cards that use `chooseOpponent`. Use `context.event.challenge.loser` instead.

server/game/cards/12-KotI/GunthorSonOfGurn.js
```diff
@@ -0,0 +1,35 @@
...
+                    activePromptTitle: 'Select a card',
+                    source: this,
+                    cardCondition: card => card.location === 'hand' && card.controller === context.opponent,
```
ystros: Same note as above re: `context.opponent`.

PR #2345, opened 2018-12-02
Even more Kings of the Isles cards

ystros:


server/game/cards/12-KotI/BronzeYohnRoyce.js
```diff
@@ -0,0 +1,20 @@
+const DrawCard = require('../../drawcard');
+
+class BronzeYohnRoyce extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                afterChallenge: event => event.challenge.winner === this.controller && this.isDefending() && this.kneeled
```
cryogen: Why the kneeled check here, am I missing something?

cryogen: Never mind, missed the "to stand him"

PR #2363, opened 2018-12-18
Make games more disconnect tolerant

ystros:
Keeps games open for a while in cases where both players have
disconnected. This should help people from losing game progress during
server-side network issues.

Additionally, the disconnect message is now delayed. This will give the
player a chance to reconnect in the case of client-side network issues,
and may help discourage other players from leaving the second they see a
disconnect message.

### DIFF
--- a/server/game/ConnectionMessages.js
+++ b/server/game/ConnectionMessages.js
@@ -1,0 +1,66 @@
+const DisconnectMessageDelay = 15 * 1000;
+
+class ConnectionMessages {
+    constructor(gameChat) {
+        this.gameChat = gameChat;
+        this.disconnectedPlayers = new Set();
+    }
+
+    printSpectatorJoined(player) {
+        this.gameChat.addAlert('info', '{0} has joined the game as a spectator', player);
+    }
+
+    printLeft(player) {
+        this.gameChat.addAlert('info', '{0} has left the game', player);
+    }
+
+    printDisconnected(player) {
+        if(player.isSpectator()) {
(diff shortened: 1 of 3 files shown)


ystros:
@cryogen Let me know if you only want certain pieces of this (either just holding the game open, or just the message delay). I tried to pick reasonable defaults for the timeouts - 2 minutes holding a game open on socket disconnect, 15 seconds delay on outputting the 'Player 1 disconnected' message.

cryogen:
We probably need to think about the reconnect and "connected to server" messages as part of this.  This will make those messages go out of sync.  Not the end of the world but worth considering

ystros:
I'm not sure there's any weirdness around the initial connect messages, but this would lead to messages saying 'Player 1 has reconnected' without a corresponding 'Player 1 has disconnected' message ever being printed. Need to think about good ways to work around that.

cryogen:
Here's what I've done for keyteki, which is somewhat similar: https://github.com/keyteki/keyteki/pull/640/files

What do you think about porting that across?

ystros:
The Keyteki version looks reasonable to me. Modifying the message to say the game will try to reconnect for 30 seconds should make it more clear that the disconnected player may be back. And it should prevent dumping the game if there's a network problem and both players are disconnected simultaneously (seems like the biggest problem right now).

I'd maybe pull the 30 second limit into a constant so it isn't repeated 3 times (game server check, actual game level timeout check, message) and is easier to adjust.

Issue #2376, opened 2019-01-03
Implement At the Gates pack


ystros:
https://www.fantasyflightgames.com/en/news/2018/12/7/at-the-gates/

* [x] 13001 - [Alysane Mormont](https://thronesdb.com/card/13001)
* [x] 13002 - [Old Gate](https://thronesdb.com/card/13002)
* [x] 13003 - [The Knight of Flowers](https://thronesdb.com/card/13003)
* [x] 13004 - [Gate of the Gods](https://thronesdb.com/card/13004)
* [x] 13005 - [Aged Craftsman](https://thronesdb.com/card/13005)
* [x] 13006 - [Flea Bottom Alley](https://thronesdb.com/card/13006)
* [x] 13007 - [R'hllor Infiltrator](https://thronesdb.com/card/13007)
* [x] 13008 - [King's Gate](https://thronesdb.com/card/13008)
* [x] 13009 - [Ser Mandon Moore](https://thronesdb.com/card/13009)
* [x] 13010 - [Lion Gate](https://thronesdb.com/card/13010)
* [x] 13011 - [Baelor Blacktyde](https://thronesdb.com/card/13011)
* [x] 13012 - [Iron Gate](https://thronesdb.com/card/13012)
* [x] 13013 - [Viserion](https://thronesdb.com/card/13013)
* [x] 13014 - [Dragon Gate](https://thronesdb.com/card/13014)
* [x] 13015 - [Southron Escort](https://thronesdb.com/card/13015)
* [x] 13016 - [River Gate](https://thronesdb.com/card/13016)
* [x] 13017 - [Littlefinger](https://thronesdb.com/card/13017)
* [x] 13018 - [White Sword Tower](https://thronesdb.com/card/13018)
* [x] 13019 - [City of Secrets](https://thronesdb.com/card/13019)
* [x] 13020 - [At the Gates](https://thronesdb.com/card/13020)

PR #2384, opened 2019-01-06
Implement Old Gate #2376

weber-knecht:
Progress toward #2376

### DIFF
--- a/server/game/cards/13.1-AtG/OldGate.js
+++ b/server/game/cards/13.1-AtG/OldGate.js
@@ -13,7 +13,7 @@ class OldGate extends DrawCard {
             cost: ability.costs.sacrificeSelf(),
             handler: () => {
                 let cards = this.controller.drawCardsToHand(2).length;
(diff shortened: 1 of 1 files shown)


server/game/cards/13.1-AtG/OldGate.js
```diff
@@ -0,0 +1,30 @@
...
+            gold: 1
+        });
+        this.action({
+            title: 'Sacrifice to draw 2 cards',
+            condition: () => this.allCharactersHaveStarkAffiliation(),
+            phase: 'challenge',
+            cost: ability.costs.sacrificeSelf(),
+            handler: () => {
+                let cards = this.controller.drawCardsToHand(2).length;
+                this.game.addMessage('{0} sacrifices {1} to draw {2} cards',
```
ystros: Since you're using the TextHelper already, you don't need the additional 'cards' in the message format. It will end up outputting something like `'Player 1 sacrifices Old Gate to draw 2 cards cards'`

PR #2391, opened 2019-01-12
New cards atg gate of the gods

weber-knecht:
Progress toward #2376

### DIFF
--- a/server/game/cards/13.1-AtG/GateOfTheGods.js
+++ b/server/game/cards/13.1-AtG/GateOfTheGods.js
@@ -1,0 +1,43 @@
+const DrawCard = require('../../drawcard.js');
+const TextHelper = require('../../TextHelper');
+
+class GateOfTheGods extends DrawCard {
+    setupCardAbilities(ability) {
+        this.plotModifiers({
+            gold: 1
+        });
+        this.action({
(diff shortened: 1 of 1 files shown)


server/game/cards/13.1-AtG/GateOfTheGods.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    characterWithHighestStrength() {
+        let charactersInPlay = this.game.filterCardsInPlay(card => card.getType() === 'character');
+        if(charactersInPlay.length === 0) {
+            return false;
+        }
+        let cardWithHighestStrength;
+        let highestStrength = 0;
+        charactersInPlay.forEach(function(card) {
```
ystros: This logic fails if multiple players tie for the highest STR character. e.g. one player controls Core Randyl Tarly (STR 5), another controls Core Balon (STR 5). The player with Gate of the Gods should be able to trigger it even though there's a tie, and the current code will sometimes allow them to but sometimes not, depending on which player created the game.

The logic should be more along the lines of:
```
get the highest STR in play
get a list of characters that have that STR
return true if the current player controls one of those characters
```

As a code style note, we generally prefer using `for` loops instead of the `forEach` method, unless there's a good reason to do so.

PR #2431, opened 2019-03-11
Support simultaneous sacrifices

### LINKED ISSUE #2410: Characters are not being sacrificed simultaneously for Return to the Fields
When you trigger Return to the Fields, I believe the chosen characters are currently being sacrificed sequentially in the order in which you select them, which opens up a reaction window for cards like Wyman Manderly and Catelyn Stark (WotN) even if you choose them to be sacrificed. For instance, if you choose Wyman last, you can trigger him twice and draw two cards before he gets sacrificed himself.

All the characters should be sacrificed simultaneously, and only then should the reaction window open.

ystros:
Fixes #2410


cryogen:
👍😀

PR #2450, opened 2019-03-28
Implement White Sword Tower

ystros:
Progress toward #2376


ystros:
Oops, this is failing because #2442 hasn't been merged yet and it doesn't have the data.

Issue #2461, opened 2019-03-31
Gerris vs The Crow is a Tricksy Bird / Khal Drogo + Aegon's Return


PrseB:
Hello,

Both issues I have experienced (#1) and heard about (#2). The first one is easy, the second seems more difficult to fix.

Issue #1 : Gerris Drinkwater vs The Crow is a Tricksy Bird (TCiaTB)

When the NW player uses TCiaTB, He chooses You Win or You Die (YWoYD). Then I use Gerris Drinkwater to swich a plot in my used pile with YWoYD.
The normal interaction (if I understand rules correctly) is that the game continue like nothing happened.

Next plot phase, I cannot choose my plot and I am forced to "reveal" YWoYD from my used pile (So I still have 6 plots while my opponent has 5 plots). The effect of the plot works correctly (If I remember well).

So my guess is that TCiaTB triggers a variable in the code. Gerris does not reset this variable at zero. So at the moment this variable triggers (next plot phase), she force the decision that was previously defined.

Issue #2 Aegon's Return vs Khal Drogo

I heard about it. It's not my experience. What seems to have happened is :

The player use Aegon's Return in the middle of the challenge phase to return Khal Drogo. But he cannot do two military challenges, he only has the option to do one.

My guess is that the number of challenges you can do is either defined at the start of challenge phase, or when the window to declare challenges open. So once the window for the number of challenges is passed, it does not update anymore.

For what I know about coding, would be incompetent to fix that.

ystros:
Thanks for the detailed report!

Fix is on the way for Tricksy Bird.

As for the other issue, the bug doesn't seem directly related to Unexpected Return. It seems to be more related to Khal Drogo leaving play after using both military challenges but before the challenge phase ends. When he returns to play (re-marshalled if returned to hand, back from discard using Unexpected Return, etc), his ability is still marked as used since he wasn't in play at the end of the challenges phase. The next round he'll give the additional challenge properly.

The easiest thing to do would be to mark the ability as unused when he leaves play, but then there'd be a bug with the following scenario:
* Initiate 2 military challenges
* Khal Drogo leaves play
* Khal Drogo re-enters play during the same phase
* A 3rd military challenge would be improperly allowed

As you say, it will probably be tricky to fix.

PrseB:
Hello ! You're welcome. Your job is excellent.

Apparently, the problem was that Khal Drogo entered player, the played did a first challenge, but was not allowed to do a second.

[...]

PR #2462, opened 2019-03-31
Handle Tricksy Bird vs moving plots

ystros:
Previously, if the plot selected by Tricksy Bird was moved out of the
plot deck by another ability, such as Ser Gerris Drinkwater or Change of
Plans, it would be selected in the following phase from the used pile.

Now, plot selection ensures that the "must reveal" plot is in the plot
deck. Additionally, if you have multiple copies of the same plot, one is
chosen for Tricksy, then moved out, you now properly have to reveal the
other copy instead.

Progress toward #2461


ystros:
Hold off on merging for the moment - though I wrote tests I feel like I may have missed some cases that need a closer look.

Issue #2377, opened 2019-01-03
Implement City of Secrets pack


ystros:
https://www.fantasyflightgames.com/en/news/2019/1/3/city-of-secrets/

* [ ] 13021 - [Smalljon Umber](https://thronesdb.com/card/13021)
* [ ] 13022 - [No Surprises](https://thronesdb.com/card/13022)
* [ ] 13026 - [Shadow of the Wall](https://thronesdb.com/card/13026)
* [ ] 13027 - [Red Keep Gaoler](https://thronesdb.com/card/13027)
* [ ] 13029 - [The Regent's Guard](https://thronesdb.com/card/13029)
* [ ] 13034 - [Shadow of the East](https://thronesdb.com/card/13034)
* [ ] 13036 - [Shadow of the South](https://thronesdb.com/card/13036)
* [ ] 13038 - [Shadow Politics](https://thronesdb.com/card/13038)

Saviouress:
Hey there I'm completely new to github, java script and throneteki.
I tried to implement Ghost of High Heart, but I'm not quite sure how to test it, or if it would even work.

I hope this is at least somewhat helpful:

`const DrawCard = require('../../drawcard.js');

class GhostOfHighHeart extends DrawCard {
	setupCardAbilities(ability) {
		this.persistentEffect({
			condition: () => !this.controller.anyCardsInPlay(card => card.getType() === 'character' && card.isLoyal()),
            effect: ability.effects.modifyStrength(1)
		});
		this.action({
			title: 'Look at a player\'s hand',
			phase: 'challenge',
			choices: {
				'Opponent': () => {
					chooseOpponent: opponent => opponent.hand.length !== 0,
					handler: context => {
						this.game.addMessage('{0} plays {1} to look at {2}\'s hand', context.player, this, context.opponent);
						this.game.promptForSelect(context.player, {
							activePromptTitle: 'Select a card',
							source: this,
							revealTargets: true,
							cardCondition: card => card.location === 'hand' && card.controller === context.opponent,
							onSelect: (player, card) => this.onCardSelected(player, card)
						});
					}
					onCardSelected(player, card) {
						let otherPlayer = card.controller;
						otherPlayer.discardCard(card);
						if(this.controller.canDraw()) {
							this.controller.drawCardsToHand(1);
						}	
						this.game.addMessage('{0} then uses {1} to discard {2} from {3}\'s hand', player, this, card, otherPlayer);
						return true;
					}
				},
				'Self': () => {
[...]

PR #2476, opened 2019-04-23
Implement Nightflyer

ystros:
Progress toward #2377


ystros:
It should be noted this is not 100% a correct implementation. Right now it's always going to say that they paid 0 to bring Nightflyer out of shadows, *then* they'll be prompted for the X amount and pay that. It's close enough that I think it's worth getting a working version of the card out, but will be confusing for people watching the logs carefully.

The true fix is to 100% support X cost for shadows in this way, but this will probably involve overhauling the way all X values are treated, which requires more thought.

PR #2504, opened 2019-05-02
Add promptid to every button

cryogen:
To make sure button clicks aren't processed if the current step does not permit it (ie, when theyve been clicked more than once)

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -75,11 +75,11 @@ class Game extends EventEmitter {
         this.skipPhase = {};
         this.cardVisibility = new CardVisibility(this);
 
-        for(let player of details.players || []) {
+        for(let player of Object.values(details.players || {})) {
             this.playersAndSpectators[player.user.username] = new Player(player.id, player.user, this.owner === player.user.username, this);
         }
 
-        for(let spectator of details.spectators || []) {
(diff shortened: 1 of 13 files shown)


server/game/gamepipeline.js
```diff
@@ -19,7 +19,7 @@ class GamePipeline {
     getCurrentStep() {
         var step = this.pipeline[0];
 
-        if(typeof(step) === 'function') {
+        if(typeof (step) === 'function') {
```
ystros: Do we have eslint rules for this (if this is the formatting we want)?

server/game/game.js
```diff
@@ -75,11 +75,11 @@ class Game extends EventEmitter {
         this.skipPhase = {};
         this.cardVisibility = new CardVisibility(this);
 
-        for(let player of details.players || []) {
+        for(let player of Object.values(details.players || {})) {
```
ystros: Seems fairly unrelated to this change?

cryogen: It is, but the game node server won’t run without this being fixed :)

server/game/gamepipeline.js
cryogen: No, and I’m not even sure it is what we want, I have too many instances of vs code with differing configs.  I will sort that out in my end

PR #2512, opened 2019-05-05
Add a /rematch command

### LINKED ISSUE #1577: add a /reset-game command
[ nothing this down in a stable place, after several devel discussion ]

We should add a /reset-game which restarts the game (same decks, same players) as if the player who created the game just clicked on "Start". The use cases/rationale for this are:

1. rematch. The two players have just completed a game and want to have a rematch without having to go through the game list again (which is annoying, as other players might get in the way and you'll have to tune game options again)

2. setup/deck testing. Single player or player with sparring partner, who want to test setups and/or a few game turns over and over again

To avoid abuses of the feature there should probably be a check that the current game is over and/or a confirmation prompt to the other player(s).

I gave a stab at implementing this by rewinding/undoing game state (see https://github.com/zacchiro/throneteki/tree/feature/reset-game-command), but I got convinced that it's a lost cause: there are too many aspects to reset and there will always be more in the future. Recreating a Game object is probably the better way to implement this.

cryogen:
Fixes #1577

### DIFF
--- a/server/lobby.js
+++ b/server/lobby.js
@@ -587,13 +587,13 @@ class Lobby {
         this.messageService.addMessage(chatMessage);
     }
 
-    onSelectDeck(socket, gameId, deckId, cb) {
+    onSelectDeck(socket, gameId, deckId) {
         let game = this.games[gameId];
         if(!game) {
-            return;
+            return undefined;
         }
 
-        Promise.all([this.cardService.getAllCards(), this.cardService.getAllPacks(), this.deckService.getById(deckId), this.cardService.getRestrictedList()])
+        return Promise.all([this.cardService.getAllCards(), this.cardService.getAllPacks(), this.deckService.getById(deckId), this.cardService.getRestrictedList()])
             .then(results => {
                 let [cards, packs, deck, restrictedList] = results;
                 let formattedDeck = formatDeckAsFullCards(deck, { cards: cards });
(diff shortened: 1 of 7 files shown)


server/lobby.js
```diff
@@ -193,25 +195,30 @@ class Lobby {
...
             });
         }
 
         socket.send('users', filteredUsers);
     }
 
     broadcastGameList(socket) {
-        let sockets = socket ? [socket] : this.sockets;
-        _.each(sockets, socket => {
+        let sockets = {};
```
ystros: Previous logic was to broadcast to all connected sockets if a single socket wasn't passed. This now broadcasts to no sockets

server/lobby.js
```diff
@@ -566,29 +573,29 @@ class Lobby {
...
+
+        this.messageService.addMessage(chatMessage);
     }
 
     onSelectDeck(socket, gameId, deckId) {
-        var game = this.games[gameId];
+        let game = this.games[gameId];
         if(!game) {
-            return;
+            return undefined;
```
ystros: Should this return a rejected Promise instead of nothing?

server/lobby.js
cryogen: This should be functionally equivalent, it just works around the old issue where it was either an array or an object, and underscore Just Dealt With It(tm)

PR #2516, opened 2019-05-09
Implement Eddard Stark (PoS)

Saviouress:
Implementation of Eddard Stark (PoS)

Hey there another try to implement a card, tried myself at the Ghost of High Heart before, this one is an easier one. I'd appreciate feedback if this was the correct way to do things.

Greetings.

### DIFF
--- a/server/game/cards/13.3-PoS/EddardStark.js
+++ b/server/game/cards/13.3-PoS/EddardStark.js
@@ -1,24 +1,29 @@
 const DrawCard = require('../../drawcard.js');
 
 class EddardStark extends DrawCard {
-	setupCardAbilities(ability) {
-			this.action({
-				condition: () => this.allCharactersHaveStarkAffiliation()
-				title: 'Kill a character'
-				phase: 'dominance'
-				cost: ability.costs.kneelSelf()
-				target: {
-					cardCondition: card => card.location === 'play area' && card.getType() === 'character' && card.getPrintedCost() <= 4
-					gameAction: 'kill'
-				},
-				handler: context => {
-					this.game.addMessage('{0} kneels {1} to kill {2}', this.controller, this, context.target);
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kill a character',
+            phase: 'dominance',
+            condition: context => this.allCharactersHaveStarkAffiliation(context.player),
+            cost: ability.costs.kneelSelf(),
+            target: {
+                cardCondition: card => card.location === 'play area' && card.getType() === 'character' && card.getPrintedCost() <= 4,
+                gameAction: 'kill'
+            },
+            message: '{player} kneels {source} to kill {target}',
+            handler: context => {
+                this.game.killCharacter(context.target);
+            },
+            limit: ability.limit.perPhase(1)
+        });
+    }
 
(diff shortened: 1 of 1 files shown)


server/game/cards/13.3-PoS/EddardStark.js
```diff
@@ -0,0 +1,26 @@
+const DrawCard = require('../../drawcard.js');
+
+class EddardStark extends DrawCard {
+	setupCardAbilities(ability) {
+			this.action({
+				condition: () => this.allCharactersHaveStarkAffiliation()
```
ystros: In order for this to work, you will need to define the `allCharactersHaveStarkAffiliation` function inside this class (but outside the `setupCardAbilities` function). You should be able to copy/paste this from similar cards.

server/game/cards/13.3-PoS/EddardStark.js
```diff
@@ -0,0 +1,26 @@
...
+				title: 'Kill a character'
+				phase: 'dominance'
+				cost: ability.costs.kneelSelf()
+				target: {
+					cardCondition: card => card.location === 'play area' && card.getType() === 'character' && card.getPrintedCost() <= 4
+					gameAction: 'kill'
+				},
+				handler: context => {
+					this.game.addMessage('{0} kneels {1} to kill {2}', this.controller, this, context.target);
```
ystros: We've started transitioning to a new message format. So while this is correct, we've been trying to implement messages for newer cards like this:
```javascript
message: '{player} kneels {source} to kill {target}',
handler: context => {
  this.game.killCharacter(context.target);
}
```

server/game/cards/13.3-PoS/EddardStark.js
```diff
@@ -0,0 +1,26 @@
+const DrawCard = require('../../drawcard.js');
+
+class EddardStark extends DrawCard {
+	setupCardAbilities(ability) {
```
ystros: The indentation for this file doesn't match up with the way we enforce code style. It would probably be productive to install an editor on your machine (we tend to use Visual Studio Code but others should work fine too), pull the code locally, and make sure your editor is set up to check the code style using `eslint`.

server/game/cards/13.3-PoS/EddardStark.js
Saviouress: Ah that looks a lot easier to understand when you read the code.

server/game/cards/13.3-PoS/EddardStark.js
Saviouress: Thanks for the feedback, just joined discord.

server/game/cards/13.3-PoS/EddardStark.js
Saviouress: I did this in Notepad++ but I guess a proper editor for Java Script would be a good idea to get.  
You do Spaces instead of Tabs?

server/game/cards/13.3-PoS/EddardStark.js
ystros: Oops, you also need a comma after this line before the title.

ystros:
I've cleaned this up, so it should be ready to merge. Thanks for the initial submission!

Issue #2511, opened 2019-05-04
Black Market Merchant allows attaching to opponents characters


dokempf:
The text explicitly states "to a character you control".

ystros:
Thanks for the report!

gunthergun:
Gifts for the Widow uses the same code it seems. I will look into the function anyCardsInPlay.

PR #2517, opened 2019-05-11
Fixed Black Market Merchant Bug #2511

gunthergun:
Fixes the Issue #2511
The problem was that the card was just put into play with no restrictions. This made it so you could target any card. I used the card 'Gifts for the Widow' as a reference as that seemed to work without the bug. To keep everything the same i changed a bit of the naming, so that it is in line with the other card.


ystros:
Thanks! Wish there was a cleaner way of doing this, but this is definitely how it's been done elsewhere.

PR #2548, opened 2019-05-30
 Don't reveal hand early for Ghost of High Heart

### LINKED ISSUE #2514: Ghost of High Heart and cancels
There was a slight problem when I tried to use Ghost of High Heart against Greyjoy and my opponent had Drowned God Fanatic in hand. I was able to look at his hand and select a card I wanted to discard, and only then did my opponent get the opportunity to use the Fanatic and cancel Ghost's ability. The interrupt window should open before the hand is revealed.

I assume Begging Brother or Treachery would have the same problem when interacting with GoHH, but I haven't tested it.

ystros:
Fixes #2514 

Will either need to rebase #2546 or this when merged.

### DIFF
--- a/server/game/GameActions/AbilityAdapter.js
+++ b/server/game/GameActions/AbilityAdapter.js
@@ -1,3 +1,5 @@
+const ThenAbilityAction = require('./ThenAbilityAction');
+
 /**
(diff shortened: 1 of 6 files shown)


cryogen:
Looks like we got away without rebasing

PR #2553, opened 2019-06-02
Convert game actions

ystros:


ystros:
Merging this directly because I keep needing them for different cards.

PR #2591, opened 2019-06-05
Implement "A Cask of Ale"

ystros:
Progress towards #2404

### DIFF
--- a/server/game/GameActions/DiscardPower.js
+++ b/server/game/GameActions/DiscardPower.js
@@ -6,7 +6,7 @@ class DiscardPower extends GameAction {
     }
 
     canChangeGameState({ card }) {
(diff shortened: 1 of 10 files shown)


server/game/GameActions/DiscardPower.js
```diff
@@ -6,7 +6,7 @@ class DiscardPower extends GameAction {
     }
 
     canChangeGameState({ card }) {
-        return ['faction', 'play area'].includes(card.location) && card.power > 0;
+        return ['active plot', 'faction', 'play area'].includes(card.location) && card.power > 0;
```
cryogen: Feels like this array is duplicated a lot and could be a return from a function so you wouldn't need to change it in so many places, maybe one for the future

Issue #2416, opened 2019-02-23
Implement Pit of Snakes pack


ystros:
* [x] 13041 - [Eddard Stark](https://thronesdb.com/card/13041)
* [x] 13042 - [Malleon's Tome](https://thronesdb.com/card/13042)
* [x] 13043 - [Maester Gormon](https://thronesdb.com/card/13043)
* [x] 13044 - [Unexpected Guile](https://thronesdb.com/card/13044)
* [x] 13045 - [Maester Mullin](https://thronesdb.com/card/13045)
* [x] 13046 - [Guard Duty](https://thronesdb.com/card/13046)
* [x] 13047 - [Ser Eldon Estermont](https://thronesdb.com/card/13047)
* [x] 13048 - [Shadow of the Throne](https://thronesdb.com/card/13048)
* [x] 13049 - [Chiswyck](https://thronesdb.com/card/13049)
* [x] 13050 - ["Hands of Gold"](https://thronesdb.com/card/13050)
* [x] 13051 - [King's Landing Proselyte](https://thronesdb.com/card/13051)
* [x] 13052 - [Outfitted for War](https://thronesdb.com/card/13052)
* [x] 13053 - [Jon Connington](https://thronesdb.com/card/13053)
* [x] 13054 - [The Dragonpit](https://thronesdb.com/card/13054)
* [x] 13055 - [Arianne Martell](https://thronesdb.com/card/13055)
* [x] 13056 - [A Pit of Snakes](https://thronesdb.com/card/13056)
* [x] 13057 - [The Most Devout](https://thronesdb.com/card/13057)
* [x] 13058 - [Statue of Baelor](https://thronesdb.com/card/13058)
* [x] 13059 - [The Faith's Decree](https://thronesdb.com/card/13059)
* [x] 13060 - [Manning the City Walls](https://thronesdb.com/card/13060)

PR #2622, opened 2019-06-19
Implement malleons tome

cryogen:
Progress towards #2416

### DIFF
--- a/server/game/GameActions/LookAtHand.js
+++ b/server/game/GameActions/LookAtHand.js
@@ -6,7 +6,7 @@ class LookAtHand extends GameAction {
     }
 
     canChangeGameState({ opponent }) {
-        return opponent.hand.length > 0;
+        return opponent && opponent.hand.length > 0;
     }
 
     createEvent({ player, opponent, context }) {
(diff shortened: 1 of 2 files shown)


server/game/GameActions/LookAtHand.js
```diff
@@ -6,7 +6,7 @@ class LookAtHand extends GameAction {
     }
 
     canChangeGameState({ opponent }) {
-        return opponent.hand.length > 0;
+        return opponent && opponent.hand.length > 0;
```
ystros: Hmm, kind of annoying, but I see why it's currently necessary.

server/game/cards/13.3-PoS/MelleonsTome.js
```diff
@@ -0,0 +1,39 @@
...
+        this.whileAttached({
+            effect: ability.effects.addIcon('intrigue')
+        });
+
+        this.reaction({
+            when: {
+                afterChallenge: event =>
+                    event.challenge.attackingPlayer === this.controller &&
+                    event.challenge.winner === this.controller &&
+                    this.parent.isAttacking()
```
ystros: Might need an explicit `this.parent &&` here, since it only should trigger when attached.

(Unfortunately can't auto-check easily because things like Hotah's Axe have abilities that trigger while not attached)

PR #2628, opened 2019-06-20
implement unexpected guile

cryogen:
Progress towards #2416

### DIFF
--- a/server/game/cards/13.3-PoS/UnexpectedGuile.js
+++ b/server/game/cards/13.3-PoS/UnexpectedGuile.js
@@ -2,7 +2,7 @@ const DrawCard = require('../../drawcard.js');
 
 class UnexpectedGuile extends DrawCard {
     setupCardAbilities(ability) {
-        this.attachmentRestriction(card => card.isShadow() && card.controller === this.controller);
+        this.attachmentRestriction({ controller: 'current', shadow: true });
         this.whileAttached({
             effect: [
                 ability.effects.modifyStrength(2),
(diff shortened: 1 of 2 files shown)


server/game/cards/13.3-PoS/UnexpectedGuile.js
```diff
@@ -0,0 +1,29 @@
+const DrawCard = require('../../drawcard.js');
+
+class UnexpectedGuile extends DrawCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction(card => card.isShadow() && card.controller === this.controller);
```
ystros: By itself, this would allow it to be attached to non-character cards. Unfortunately there's no way to fill in defaults when it's a restriction function instead of a restriction matcher object (which gets the defaults merged in).

server/game/cards/13.3-PoS/UnexpectedGuile.js
```diff
@@ -0,0 +1,29 @@
...
+        this.whileAttached({
+            effect: [
+                ability.effects.modifyStrength(2),
+                ability.effects.addKeyword('Insight')
+            ]
+        });
+
+        this.forcedInterrupt({
+            when: {
+                afterChallenge: event => event.challenge.isParticipating(this.parent)
```
ystros: afterChallenge corresponds to a winner / loser being chosen. onChallengeFinished is the event we want. This seems to be the first card that uses it!

PR #2629, opened 2019-06-20
implement guard duty

cryogen:


server/game/cards/13.3-PoS/GuardDuty.js
```diff
@@ -0,0 +1,22 @@
+const DrawCard = require('../../drawcard.js');
+
+class GuardDuty extends DrawCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction(
+            { faction: 'thenightswatch' }
+        );
+        this.reaction({
+            when: {
+                onDeclaredAsDefender: event => event.card === this.parent
```
ystros: Missing check that 1. the parent can stand, 2. the parent is currently knelt (which will probably need a check that there is a parent)

server/game/cards/13.3-PoS/GuardDuty.js
```diff
@@ -0,0 +1,22 @@
...
+            },
+            cost: ability.costs.kneelSelf(),
+            handler: () => {
+                this.controller.standCard(this.parent);
```
ystros: Missing game message

Issue #2639, opened 2019-06-22
Implement Beneath the Red Keep pack


ystros:
* [x] 13061 - [White Harbor Watchman](https://thronesdb.com/card/13061)
* [x] 13062 - [Tower of the Hand](https://thronesdb.com/card/13062)
* [x] 13063 - [Alyce Graceford](https://thronesdb.com/card/13063)
* [x] 13064 - [The Maidenvault](https://thronesdb.com/card/13064)
* [x] 13065 - [Forest Patrol](https://thronesdb.com/card/13065)
* [x] 13066 - [Blackbird](https://thronesdb.com/card/13066)
* [x] 13067 - [Castle Guard](https://thronesdb.com/card/13067)
* [x] 13068 - [Public Execution](https://thronesdb.com/card/13068)
* [x] 13069 - [Doubting Septa](https://thronesdb.com/card/13069)
* [x] 13070 - [Tunnels of the Red Keep](https://thronesdb.com/card/13070)
* [x] 13071 - [Gorold Goodbrother](https://thronesdb.com/card/13071)
* [x] 13072 - [Shadow of the Isles](https://thronesdb.com/card/13072)
* [x] 13073 - [Shadowblack Assassin](https://thronesdb.com/card/13073)
* [x] 13074 - [Dragon Skull](https://thronesdb.com/card/13074)
* [x] 13075 - [Sarella Sand](https://thronesdb.com/card/13075)
* [x] 13076 - [Meadowlark](https://thronesdb.com/card/13076)
* [x] 13077 - [Advisor to the Crown](https://thronesdb.com/card/13077)
* [x] 13078 - [Knights of the Mind](https://thronesdb.com/card/13078)
* [x] 13079 - [Kingdom of Shadows](https://thronesdb.com/card/13079)
* [x] 13080 - [City of Spiders](https://thronesdb.com/card/13080)

PR #2642, opened 2019-06-22
implemented Tower Of The Hand (BtRK)

EvadLhorg:
First contribution to the project. I implemented Tower Of The Hand by copying & pasting the relevant parts from Oldtown Undercity, Maidens Bane and Jory Cassel. I would appreciate any feedback!

### DIFF
--- a/server/game/cards/13.4-BtRK/TowerOfTheHand.js
+++ b/server/game/cards/13.4-BtRK/TowerOfTheHand.js
@@ -13,7 +13,10 @@ class TowerOfTheHand extends DrawCard {
                 cardCondition: card => card.location === 'play area' && card.getType() === 'character' && card.isFaction('stark'),
                 gameAction: 'stand'
             },
-            message: '{player} kneels {source} to stand {target}'
+            message: '{player} kneels {source} to stand {target}',
+            handler: context => {
+                context.player.standCard(context.target);
+            }
         });
     }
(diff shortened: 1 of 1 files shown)


EvadLhorg:
Progress towards #2639

server/game/cards/13.4-BtRK/TowerOfTheHand.js
```diff
@@ -0,0 +1,26 @@
...
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onCardOutOfShadows: event => 
+                    event.card.controller === this.controller &&
+                    this.game.anyPlotHasTrait('Winter')
+            },
+            limit: ability.limit.perPhase(1),
+            target: {
+                cardCondition: card => card.location === 'play area' && card.getType() === 'character' && card.isFaction('stark'),
```
ystros: This will currently let you choose a character that is already standing. You'll need to add a `&& card.kneeled` check here.

(in the future it may not be required to explicitly check this, but for now it is)

EvadLhorg: Thank you for your feedback. I added the check for the target card in the commit a2d776fd4ff4be69d8134c71819e4691c15bf9ad

ystros:
Thanks again!

PR #2649, opened 2019-06-26
implement tunnels of the red keep

EvadLhorg:
Progress towards #2639

### DIFF
--- a/server/game/cards/13.4-BtRK/TunnelsOfTheRedKeep.js
+++ b/server/game/cards/13.4-BtRK/TunnelsOfTheRedKeep.js
@@ -5,13 +5,14 @@ class TunnelsOfTheRedKeep extends DrawCard {
         this.action({
             title: 'Kneel and return to shadows',
             cost: ability.costs.kneelSelf(),
-            limit: ability.limit.perPhase(1),
+            max: ability.limit.perPhase(1),
             handler: context => {
                 context.player.putIntoShadows(this, false);
-                this.game.addMessage('{0} kneels {1} to return it to shadows.', this.controller, this);
+                this.game.addMessage('{0} kneels {1} to return it to shadows.', context.player, this);
+                let targetCharacters = context.player.filterCardsInPlay(card => card.getType() === 'character');
                 this.untilEndOfPhase(ability => ({
-                    match: card => card.getType() === 'character' && card.controller === this.controller,
-                    effect: ability.effects.dynamicStrength(() => this.controller.shadows.length)
+                    match: targetCharacters,
+                    effect: ability.effects.dynamicStrength(() => context.player.shadows.length)
                 }));
             }
         });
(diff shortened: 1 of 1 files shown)


server/game/cards/13.4-BtRK/TunnelsOfTheRedKeep.js
```diff
@@ -0,0 +1,23 @@
+const DrawCard = require('../../drawcard.js');
+
+class TunnelsOfTheRedKeep extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kneel and return to shadows',
+            cost: ability.costs.kneelSelf(),
+            limit: ability.limit.perPhase(1),
```
ystros: Should be `max: ability.limit.perPhase(1)` instead of limit here (per the card).

server/game/cards/13.4-BtRK/TunnelsOfTheRedKeep.js
```diff
@@ -0,0 +1,23 @@
...
+        this.action({
+            title: 'Kneel and return to shadows',
+            cost: ability.costs.kneelSelf(),
+            limit: ability.limit.perPhase(1),
+            handler: context => {
+                context.player.putIntoShadows(this, false);
+                this.game.addMessage('{0} kneels {1} to return it to shadows.', this.controller, this);
+                this.untilEndOfPhase(ability => ({
+                    match: card => card.getType() === 'character' && card.controller === this.controller,
+                    effect: ability.effects.dynamicStrength(() => this.controller.shadows.length)
```
ystros: Everywhere in this handler should use `context.player` instead of `this.controller`. This is in case a player takes control of Tunnels, when the ability is activated and `putIntoShadows` finishes, it will go into its owner's shadow area, but the effect should apply to the player that stole and triggered it.

server/game/cards/13.4-BtRK/TunnelsOfTheRedKeep.js
```diff
@@ -0,0 +1,23 @@
...
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kneel and return to shadows',
+            cost: ability.costs.kneelSelf(),
+            limit: ability.limit.perPhase(1),
+            handler: context => {
+                context.player.putIntoShadows(this, false);
+                this.game.addMessage('{0} kneels {1} to return it to shadows.', this.controller, this);
+                this.untilEndOfPhase(ability => ({
+                    match: card => card.getType() === 'character' && card.controller === this.controller,
```
ystros: Lasting effects only apply to the cards that were in play at the time it began. So you'll need to create an array of these cards and then pass the array into `match` instead of a function. Look at [Nymeria Sand](https://github.com/throneteki/throneteki/blob/master/server/game/cards/02.2-TRtW/NymeriaSand.js#L15-L24) as an example (but use `context.player` instead of `this.controller`).

EvadLhorg: Thank you for your great feedback. I committed the requested changes

EvadLhorg: Just asking for clarity on the workflow here: Am i supposed to resolve the conversation now after i committed the changes or do you do that as the reviewer? Thanks!

ystros: I don't pay much attention to whether the conversation is resolved, but you could click it after you've pushed up requested changes.

PR #2650, opened 2019-06-26
implement the maidenvault

EvadLhorg:
Progress towards #2639

### DIFF
--- a/server/game/cards/13.4-BtRK/TheMaidenvault.js
+++ b/server/game/cards/13.4-BtRK/TheMaidenvault.js
@@ -1,0 +1,37 @@
+const DrawCard = require('../../drawcard.js');
+const GameActions = require('../../GameActions');
+
+class TheMaidenvault extends DrawCard {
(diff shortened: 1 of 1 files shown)


server/game/cards/13.4-BtRK/TheMaidenvault.js
```diff
@@ -0,0 +1,37 @@
...
+            }
+        });
+        this.interrupt({
+            when: {
+                onPhaseEnded: event => event.phase === 'challenge'
+            },
+            cost: ability.costs.kneelSelf(),
+            handler: context => {
+                context.player.putIntoShadows(this, false);
+                this.game.addMessage('{0} kneels {1} to return it to shadows.', this.controller, this);
```
ystros: Use `context.player` here to make sure the message matches the right player (in case of take-control)

EvadLhorg: Changed it in the commit below. Thanks!

PR #2651, opened 2019-06-26
implement shadow of the isles

EvadLhorg:
Progress towards #2639

as inspiration i used "shadow of the Throne" and "nothing burns like the cold"

### DIFF
--- a/server/game/cards/13.4-BtRK/ShadowOfTheIsles.js
+++ b/server/game/cards/13.4-BtRK/ShadowOfTheIsles.js
@@ -1,0 +1,36 @@
+const DrawCard = require('../../drawcard');
+const GameActions = require('../../GameActions');
+const Messages = require('../../Messages');
+
+class ShadowOfTheIsles extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Choose and discard locations',
+            targets: {
+                choosingPlayer: 'each',
+                activePromptTitle: 'Select a location',
+                cardCondition: (card, context) => card.location === 'play area' && card.controller === context.choosingPlayer && card.getType() === 'location' && !card.isLimited(),
+                gameAction: 'discard',
+                messages: Messages.eachPlayerTargetingForCardType('locations')
+            },
+            message: '{player} plays {source} to have each player choose and discard a non-limited location',
+            handler: context => {
+                let actions = context.targets.map(card => GameActions.discardCard({ card }));
+
+                this.game.resolveGameAction(
+                    GameActions.simultaneously(actions),
+                    context
+                );
+
+                if(this.game.anyPlotHasTrait('War')) {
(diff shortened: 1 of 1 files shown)


server/game/cards/13.4-BtRK/ShadowOfTheIsles.js
```diff
@@ -0,0 +1,36 @@
+const DrawCard = require('../../drawcard');
+const GameActions = require('../../GameActions');
+const Messages = require('../../Messages');
+
+class ShadowOfTheIsles extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Choose and discard locations',
+            targets: {
```
ystros: Since there's only a single *type* of target (non-limited location), this should use `target` (singular). Nothing Burns uses `targets` (plural) because it's selecting multiple types of targets. The keys nested under it are the target name ("attachment"), then the value is the target definition (prompt title, card condition, etc for choosing attachments).

server/game/cards/13.4-BtRK/ShadowOfTheIsles.js
```diff
@@ -0,0 +1,36 @@
...
+                choosingPlayer: 'each',
+                activePromptTitle: 'Select a location',
+                cardCondition: (card, context) => card.location === 'play area' && card.controller === context.choosingPlayer && card.getType() === 'location' && !card.isLimited(),
+                gameAction: 'discard',
+                messages: Messages.eachPlayerTargetingForCardType('locations')
+            },
+            message: '{player} plays {source} to have each player choose and discard a non-limited location',
+            handler: context => {
+                let actions = context.targets.map(card => GameActions.discardCard({ card }));
```
ystros: Since this is an "each player" targeting card, you'll need to get the list of cards the same way Nothing Burns and similar cards do: 
```javascript
let cards = context.targets.selections.map(selection => selection.value).filter(card => !!card);
```

(plural here is correct).

Then you can do a `map` on those cards into discardCard actions. 

We probably should have a better mechanism for doing this.

EvadLhorg: Thanks for the explanation, i think i understand the difference in comparison to "normal" targeting, but could you explain what the filtering `.filter(card => !!card)` does? I don´t understand the !! operator. Is it a double negation? Please have a look at the changes in the commit 1b688dd2aea5b1a7f04d7468b36aede6331bac7e

ystros: Yes, `!!` is a double negation, which is a somewhat typical Javascript convention to transform an arbitrary object into a boolean. e.g. `!!(null) === false`, `!!(object) === true`. So it's filtering out any selections where the player did not choose a card. I think it's more relevant for "if able" abilities (since the player may not have cards to choose), but I think it may still be needed here.

Issue #2640, opened 2019-06-22
Implement The Blackwater pack


ystros:
* [x] 13081 - [Smallfolk Mob](https://thronesdb.com/card/13081)
* [x] 13082 - [Shadow of the North](https://thronesdb.com/card/13082)
* [x] 13083 - [Highgarden Jester](https://thronesdb.com/card/13083)
* [x] 13084 - [Shadow of the Rose](https://thronesdb.com/card/13084)
* [x] 13085 - [Yoren](https://thronesdb.com/card/13085)
* [x] 13086 - [Bound for the Wall](https://thronesdb.com/card/13086)
* [x] 13087 - [Ser Preston Greenfield](https://thronesdb.com/card/13087)
* [x] 13088 - [Blackwater Rush](https://thronesdb.com/card/13088)
* [x] 13089 - [Tyrion Lannister](https://thronesdb.com/card/13089)
* [x] 13090 - [A Very Large Shadow](https://thronesdb.com/card/13090)
* [x] 13091 - [Maester Kerwin](https://thronesdb.com/card/13091)
* [x] 13092 - ["The Bloody Cup"](https://thronesdb.com/card/13092)
* [x] 13093 - [Covert Loyalist](https://thronesdb.com/card/13093)
* [x] 13094 - [Music of Dragons](https://thronesdb.com/card/13094)
* [x] 13095 - [Dagos Manwoody](https://thronesdb.com/card/13095)
* [x] 13096 - [The Mountain's Skull](https://thronesdb.com/card/13096)
* [x] 13097 - [Sparrows](https://thronesdb.com/card/13097)
* [x] 13098 - [White Cloak](https://thronesdb.com/card/13098)
* [x] 13099 - [The White Book](https://thronesdb.com/card/13099)
* [x] 13100 - [King's Landing Riot](https://thronesdb.com/card/13100)

PR #2655, opened 2019-06-30
implement White Cloak

EvadLhorg:
Progress towards #2640

### DIFF
--- a/server/game/cards/13.5-TB/WhiteCloak.js
+++ b/server/game/cards/13.5-TB/WhiteCloak.js
@@ -1,0 +1,32 @@
+const DrawCard = require('../../drawcard.js');
+
+class WhiteCloak extends DrawCard {
+    setupCardAbilities(ability) {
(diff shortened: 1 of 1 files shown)


server/game/cards/13.5-TB/WhiteCloak.js
```diff
@@ -0,0 +1,32 @@
...
+            ]
+        });
+        this.interrupt({
+            canCancel: true,
+            when: {
+                onCharacterKilled: event => event.allowSave && event.card.canBeSaved() && this.isKingOrQueen(event.card)
+            },
+            cost: ability.costs.kneelParent(),
+            handler: context => {
+                let parent = context.cardStateWhenInitiated.parent;
```
ystros: Just FYI, this probably wasn't strictly necessary. It's only important when the parent character is killed or sacrificed as a cost (e.g. Glamor), and thus becomes unattached before the effect of the ability resolves. But this should still work.

ystros:
Looks good, thanks!

PR #2657, opened 2019-07-02
fix Lionstar target cardCondition to only target printedCost 4 or lower characters

EvadLhorg:


EvadLhorg:
Yesterday i noticed that Lionstar lets me target Queen Cersei and Tywin, so i decided to look up the card and fix the target cardCondition.

ystros:
Thanks! I think someone reported this in the lobby one time, but I lost track of it.

PR #2658, opened 2019-07-03
Refactor and fix the blocklist to update when the api changes a user's blocklist

cryogen:


ystros:
Looks reasonable, seems like there's some additional mocking that needs to be done in tests now.

PR #2661, opened 2019-07-06
add attachmentRestriction for Ruby of R'hllor

### LINKED ISSUE #2659: Ruby of R'hllor does not check for R'hllor trait when marshalled
You can attach the Ruby of R'hllor to any character when it should be restricted to R'hllor.  A highly competitive and frequently played card, so definitely a high priority issue.

EvadLhorg:
The card was missing the attachmentRestriction so it can only be attached to R'hllor characters
Closes #2659


ystros:
Thanks!

PR #2662, opened 2019-07-07
Implement Yoren #2655

weber-knecht:
Progress towards #2640

### DIFF
--- a/server/game/cards/13.5-TB/Yoren.js
+++ b/server/game/cards/13.5-TB/Yoren.js
@@ -17,7 +17,8 @@ class Yoren extends DrawCard {
                     },
                     match: context.target,
                     effect: ability.effects.takeControl(this.controller)
-                }));            }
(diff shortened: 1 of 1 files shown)


server/game/cards/13.5-TB/Yoren.js
```diff
@@ -0,0 +1,28 @@
+const DrawCard = require('../../drawcard.js');
+
+class Yoren extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardEntersPlay: event => event.card === this
+            },
+            target: {
+                cardCondition: card => card.getType() === 'character' && card.owner !== this.controller && card.getPrintedCost() <= 5
```
ystros: You'll need to check that the location is 'play area', otherwise it'll allow players to take control of cards in the opponent discard / dead piles.

PR #2670, opened 2019-07-13
implement Alyce Graceford

EvadLhorg:
do not merge yet, card code needs to be fixed

Progress towards #2639

### DIFF
--- a/server/game/cards/13.4-BtRK/AlyceGraceford.js
+++ b/server/game/cards/13.4-BtRK/AlyceGraceford.js
@@ -3,7 +3,7 @@ const DrawCard = require('../../drawcard.js');
 class AlyceGraceford extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
-            title: 'Kneel to stand character',
+            title: 'Give renown',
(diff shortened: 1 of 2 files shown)


server/game/cards/13.4-BtRK/AlyceGraceford.js
```diff
@@ -0,0 +1,26 @@
+const DrawCard = require('../../drawcard.js');
+
+class AlyceGraceford extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Kneel to stand character',
+            cost: ability.costs.kneelSelf(),
+            target: {
+                activePromptTitle: 'Select another The Seven character',
```
ystros: Generally we don't do specific prompt messages like this, unless it would be confusing otherwise. Removing `activePromptTitle` will revert it to the default ("Select a character"), which should be sufficient.

EvadLhorg: Thank you for the review, i removed the activePromptTitle

PR #2673, opened 2019-07-13
implement shadowblack assassin

EvadLhorg:
card code needs to be fixed

Progress towards #2639

### DIFF
--- a/server/game/cards/13.4-BtRK/ShadowblackAssassin.js
+++ b/server/game/cards/13.4-BtRK/ShadowblackAssassin.js
@@ -19,6 +19,6 @@ class ShadowblackAssassin extends DrawCard {
     }
 }
 
-ShadowblackAssassin.code = 'TODO';
+ShadowblackAssassin.code = '13073';
 
 module.exports = ShadowblackAssassin;
--- a/server/game/cards/13.4-BtRK/ShadowblackAssassin.js
+++ b/server/game/cards/13.4-BtRK/ShadowblackAssassin.js
@@ -1,0 +1,24 @@
+const DrawCard = require('../../drawcard');
+
+class ShadowblackAssassin extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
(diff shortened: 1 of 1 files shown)


server/game/cards/13.4-BtRK/ShadowblackAssassin.js
```diff
@@ -0,0 +1,24 @@
+const DrawCard = require('../../drawcard');
+
+class ShadowblackAssassin extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardOutOfShadows: event => event.card === this
+            },
+            target: {
+                type: 'select',
```
ystros: The card text says 'choose' so this isn't needed. `type: 'select'` is only used for abilities that have implicit targeting For example, "place a token on a character" like Tears of Lys uses it because it doesn't say to choose the character.

server/game/cards/13.4-BtRK/ShadowblackAssassin.js
```diff
@@ -0,0 +1,24 @@
...
+
+class ShadowblackAssassin extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            when: {
+                onCardOutOfShadows: event => event.card === this
+            },
+            target: {
+                type: 'select',
+                activePromptTitle: 'Select a character with STR 1',
```
ystros: No need for the specific prompt title.

server/game/cards/13.4-BtRK/ShadowblackAssassin.js
```diff
@@ -0,0 +1,24 @@
...
+                cardCondition: card => card.location === 'play area' && card.controller !== this.controller && card.getType() === 'character' && card.getStrength() === 1
```
ystros: The card text says "STR 1 or lower", so your comparison should be <= 1. Additionally, you should be able to kill your own characters, so no need for the controller check.

PR #2674, opened 2019-07-13
implement dragon skull

EvadLhorg:
card code needs to be fixed

Progress towards #2639

### DIFF
--- a/server/game/cards/13.4-BtRK/DragonSkull.js
+++ b/server/game/cards/13.4-BtRK/DragonSkull.js
@@ -19,6 +19,6 @@ class DragonSkull extends DrawCard {
     }
 }
 
-DragonSkull.code = 'TODO';
+DragonSkull.code = '13074';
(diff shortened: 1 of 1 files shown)


server/game/cards/13.4-BtRK/DragonSkull.js
```diff
@@ -0,0 +1,24 @@
...
+            effect: [
+                ability.effects.addKeyword('Intimidate'),
+                ability.effects.addKeyword('Renown')
+            ]
+        });   
+    }
+
+    areNoDragonsInPlay() {
+        return !this.game.getPlayers().some(player => {
+            return player.anyCardsInPlay(card => card.getType() === 'character' && card.hasTrait('Dragon'));
```
ystros: This works but for future reference, you could also do `!this.game.anyCardsInPlay(card => ...)` instead.

PR #2680, opened 2019-07-15
Implement Shadow Of the North #2655

weber-knecht:
Progress towards #2640


server/game/cards/13.5-TB/ShadowOfTheNorth.js
```diff
@@ -0,0 +1,29 @@
+const DrawCard = require('../../drawcard.js');
+
+class ShadowOfTheNorth extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Raise claim by 1',
+            max: ability.limit.perChallenge(1),
+            cardCondition: () => this.game.isDuringChallenge(),
```
ystros: This should just be `condition`. `condition` is used at the ability level, `cardCondition` is used at the targeting level.

PR #2682, opened 2019-07-15
implement knights of the mind

EvadLhorg:
card code needs to be fixed
i would love feedback on the handler implementation as i am not sure if i did that correctly when there is no "choose" part in the card text

### DIFF
--- a/server/game/cards/13.4-BtRK/KnightsOfTheMind.js
+++ b/server/game/cards/13.4-BtRK/KnightsOfTheMind.js
@@ -1,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+
+class KnightsOfTheMind extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            handler: () => {
+                this.game.addMessage('{0} plays {1} to give each Maester character a military icon until the end of the phase', this.controller, this);
+                this.untilEndOfPhase(ability => ({
+                    match: card => card.location === 'play area' && card.getType() === 'character' && card.hasTrait('Maester') && card.controller === this.controller,
+                    effect: ability.effects.addIcon('military')
+                }));
+            }
+        });
+    }
+}
+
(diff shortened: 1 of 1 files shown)


server/game/cards/13.4-BtRK/KnightsOfTheMind.js
```diff
@@ -0,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+
+class KnightsOfTheMind extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            handler: () => {
+                this.game.addMessage('{0} plays {1} to give each Maester character a military icon until the end of the phase', this.controller, this);
+                this.untilEndOfPhase(ability => ({
+                    match: card => card.location === 'play area' && card.getType() === 'character' && card.hasTrait('Maester') && card.controller === this.controller,
```
ystros: Since this is a lasting effect, the recipients of the effect are fixed at the time the effect starts. So instead of a function, you should pass an array with all matching cards in play at the time.

See Nymeria Sand as an example:
https://github.com/throneteki/throneteki/blob/master/server/game/cards/02.2-TRtW/NymeriaSand.js#L15-L24

server/game/cards/13.4-BtRK/KnightsOfTheMind.js
```diff
@@ -0,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+
+class KnightsOfTheMind extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            handler: () => {
```
ystros: Should add a `title` for this action.

server/game/cards/13.4-BtRK/KnightsOfTheMind.js
EvadLhorg: thanks for the specific feedback, i really appreciate it. fixed the targeting

PR #2696, opened 2019-07-28
implement shadow of the rose

EvadLhorg:
not 100% sure if the order of actions happening is correctly implemented this way. the card reads that first you choose the card to put into shadows, then shuffle your deck and then optionally return shadow of the rose to shadows. I do not know if the shuffling of the deck executed by the promptForDeckSearch happens after the complete resolution of the onSelect and onCancel handler functions or not. In any case, i do not think it matters that much if you first return the card to shadows and then shuffle or do it the other way around.
Appreciate any feedback!


EvadLhorg:
progress towards #2640

PR #2702, opened 2019-08-08
add the chosen xValue of Nightflyer to the message

EvadLhorg:
While playing i noticed that the chosen xValue for Nightflyer is not written in the game log so i decided to add it


ystros:
Looks good, thanks!

PR #2715, opened 2019-08-22
Implement The White Book

ystros:
Progress towards #2640


ystros:
Tests here should go green once the data has been merged in.

PR #2728, opened 2019-08-30
Adapt all control changing effects to also target cards you control

dokempf:
Following up on Yoren in #2727, I went through all control-changing effects and changed them. Admittedly, some of these are purely academic, but I do not see why the software should not accept these.

### DIFF
--- a/server/game/cards/08.1-TAK/DaarioNaharis.js
+++ b/server/game/cards/08.1-TAK/DaarioNaharis.js
@@ -7,7 +7,7 @@ class DaarioNaharis extends DrawCard {
                 afterChallenge: event => event.challenge.winner === this.controller && this.isAttacking()
             },
             target: {
-                cardCondition: card => card.location === 'play area' && card !== this && (card.kneeled || card.controller !== this.controller) &&
+                cardCondition: card => card.location === 'play area' && card !== this &&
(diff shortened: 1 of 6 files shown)


server/game/cards/08.1-TAK/DaarioNaharis.js
```diff
@@ -7,7 +7,7 @@ class DaarioNaharis extends DrawCard {
                 afterChallenge: event => event.challenge.winner === this.controller && this.isAttacking()
             },
             target: {
-                cardCondition: card => card.location === 'play area' && card !== this && (card.kneeled || card.controller !== this.controller) &&
+                cardCondition: card => card.location === 'play area' && card !== this &&
```
ystros: Slight additional change that needs to be made - in the handler function, the take control effect is wrapped in an if checking that the controllers aren't equal. That if should be removed as well since the take control effect should always be applied.

dokempf:
I removed that if clause and updated the PR. No worries about late reviews - it is not exactly a pressing issue. You can also turn this down if you think it confuses people - it will not scare me off contributions in the future.

ystros:
Thanks! I think moving us towards more rules accurate play is generally a good thing, and there are enough edge cases around take control that this is probably necessary regardless of any potential mild confusion. Definitely appreciate it!

PR #2742, opened 2019-09-10
fix the azor ahai reborn bug

### LINKED ISSUE #2731: Azor Ahai Reborn participation
Hello,

Stannis has Azor Ahai Reborn, Acolyte is initiating a challenge which makes him paticipate, then the acolyte is killed by Robert Strong and Stannis should not participate any more, though he does. Seems like a bug. See screenshot.

Sergey
Bug

EvadLhorg:
that stannis is still participating in a challenge even though the last rhllor character was removed

fixes #2731

### DIFF
--- a/server/game/effects.js
+++ b/server/game/effects.js
@@ -114,13 +114,18 @@ const Effects = {
                 let challenge = context.game.currentChallenge;
                 if(card.canParticipateInChallenge() && !challenge.isAttacking(card)) {
                     challenge.addAttacker(card);
-                }
+                    card.consideredToBeAttackingWasApplied = true;
+                    context.game.once('afterChallenge', () => {
+                        card.consideredToBeAttackingWasApplied = false;
+                    });
+                }                
             },
             unapply: function(card, context) {
                 let challenge = context.game.currentChallenge;
 
-                if(challenge && challenge.isAttacking(card)) {
+                if(challenge && challenge.isAttacking(card) && card.consideredToBeAttackingWasApplied) {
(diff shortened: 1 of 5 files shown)


EvadLhorg:
The current fix to AzorAhaiReborn is not yet fully functional. If Stannis and another R'hllor character are declared as attackers in a challenge and the other R'hllor character is removed, Stannis will also be removed, which is wrong.

server/game/effects.js
```diff
@@ -114,13 +114,18 @@ const Effects = {
                 let challenge = context.game.currentChallenge;
                 if(card.canParticipateInChallenge() && !challenge.isAttacking(card)) {
                     challenge.addAttacker(card);
-                }
+                    card.consideredToBeAttackingWasApplied = true;
+                    context.game.once('afterChallenge', () => {
```
ystros: I think this would actually work, but there are some bad interactions from the effect engine. Since the effect has a `condition`, I think the engine will continually `unapply` and then re-`apply` the effect. So this event handler will be created more than once. Which in terms of what it's doing is fine - worst case the flag is set to `false` multiple times. It could be bad in terms of introducing an unintended number of event listener - I'm not sure whether there are implications in Node for that.

The alternate solution is to record which characters were declared on the `Challenge` class itself, and check that instead of adding this flag. Even doing that, though, there are edge cases - declare Azor Stannis and another R'hllor character, remove Azor Stannis from the challenge. Even though he was declared, since he was removed, Azor Ahai should make him participate again.

PR #2747, opened 2019-09-11
implement fishmongers square

EvadLhorg:
progress towards #2641

### DIFF
--- a/server/game/cards/13.6-LMHR/FishmongersSquare.js
+++ b/server/game/cards/13.6-LMHR/FishmongersSquare.js
@@ -6,13 +6,19 @@ class FishmongersSquare extends DrawCard {
             title: 'Draw card and gain gold',
             phase: 'marshal',
             cost: ability.costs.kneelSelf(),
-            condition: () => this.opponentDiscardPileHas10(),
+            condition: context => this.opponentDiscardPileHas10() && (context.player.canDraw() || context.player.canGainGold()),
             handler: context => {
+                let messageSegments = [];
                 if(context.player.canDraw()) {
                     context.player.drawCardsToHand(1);
+                    messageSegments.push('draw 1 card');
                 }
-                this.game.addGold(context.player, 1);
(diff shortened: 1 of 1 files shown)


server/game/cards/13.6-LMHR/FishmongersSquare.js
```diff
@@ -0,0 +1,28 @@
...
+        this.action({
+            title: 'Draw card and gain gold',
+            phase: 'marshal',
+            cost: ability.costs.kneelSelf(),
+            condition: () => this.opponentDiscardPileHas10(),
+            handler: context => {
+                if(context.player.canDraw()) {
+                    context.player.drawCardsToHand(1);
+                }
+                this.game.addGold(context.player, 1);
```
ystros: It's possible that you cannot gain gold.

server/game/cards/13.6-LMHR/FishmongersSquare.js
```diff
@@ -0,0 +1,28 @@
...
+                this.game.addMessage('{0} kneels {1} to draw 1 card and gain 1 gold', context.player, this);
```
ystros: It might be nice if the message only reflected what happened. i.e. if they only draw, then leave out the gain gold portion of the message. We're not super consistent about this, though, so don't worry about it too much if you find it difficult. My Arya Stark PR (#2736) has an example

server/game/cards/13.6-LMHR/FishmongersSquare.js
```diff
@@ -0,0 +1,28 @@
+const DrawCard = require('../../drawcard.js');
+
+class FishmongersSquare extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Draw card and gain gold',
+            phase: 'marshal',
+            cost: ability.costs.kneelSelf(),
+            condition: () => this.opponentDiscardPileHas10(),
```
ystros: The condition should also check that the player can either draw or gain gold.

PR #2751, opened 2019-09-11
Implement poisoned dagger

EvadLhorg:
progress towards #2641 

copied and adjusted the handler body from stone crows

### DIFF
--- a/server/game/cards/13.6-LMHR/PoisonedDagger.js
+++ b/server/game/cards/13.6-LMHR/PoisonedDagger.js
@@ -21,6 +21,11 @@ class PoisonedDagger extends DrawCard {
                     activePromptTitle: 'Select character to kill',
                     source: this,
                     gameAction: 'kill',
+                    onCancel: player => {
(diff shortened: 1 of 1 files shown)


server/game/cards/13.6-LMHR/PoisonedDagger.js
```diff
@@ -0,0 +1,38 @@
...
+                let otherPlayer = context.event.challenge.loser;
+                this.game.promptForSelect(otherPlayer, {
+                    cardCondition: card => (
+                        card.isParticipating() &&
+                        card.getType() === 'character' &&
+                        card.controller === otherPlayer),
+                    activePromptTitle: 'Select character to kill',
+                    source: this,
+                    gameAction: 'kill',
+                    onSelect: (player, card) => {
```
ystros: There should also be an `onCancel` here that adds an alert in the game log saying the player cancelled the resolution of the ability.

EvadLhorg: i looked at ser mandon moore for the onCancel handler as i figured it would be worth an alert if the player does not choose any characters.

PR #2752, opened 2019-09-11
implement recruit from the dungeons

EvadLhorg:
progress towards #2641 

not 100% sure on the afterChallenge method and how i bound it with this.game.once as i do not fully understand what is happening. would appreciate feedback on it!

### DIFF
--- a/server/game/cards/13.6-LMHR/RecruitFromTheDungeons.js
+++ b/server/game/cards/13.6-LMHR/RecruitFromTheDungeons.js
@@ -10,13 +10,13 @@ class RecruitFromTheDungeons extends DrawCard {
                 this.challengeToParticipateIn = this.game.currentChallenge;
                 this.challengeToParticipateIn.addDefender(this);
                 this.game.addMessage('{0} kneels {1} to have {1} participate in the challenge on their side', context.player, this);
-                this.game.once('afterChallenge', this.afterChallenge.bind(this.challengeToParticipateIn, context));
+                this.game.once('afterChallenge', () => this.afterChallenge(this.challengeToParticipateIn, context));
             }
         });
     }
 
     afterChallenge(challenge, context) {
-        if(challenge.looser === context.player) {
+        if(challenge.loser === context.player) {
             context.player.moveCard(this, 'shadows');
             this.game.addMessage('{0} returns {1} to shadows', context.player, this);
         }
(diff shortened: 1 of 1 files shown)


server/game/cards/13.6-LMHR/RecruitFromTheDungeons.js
```diff
@@ -0,0 +1,28 @@
...
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Have character participate in the challenge',
+            condition: context => this.game.isDuringChallenge({ defendingPlayer: context.player }) && this.canParticipateInChallenge(),
+            cost: ability.costs.kneelSelf(),
+            handler: context => {
+                this.challengeToParticipateIn = this.game.currentChallenge;
+                this.challengeToParticipateIn.addDefender(this);
+                this.game.addMessage('{0} kneels {1} to have {1} participate in the challenge on their side', context.player, this);
+                this.game.once('afterChallenge', this.afterChallenge.bind(this.challengeToParticipateIn, context));
```
ystros: `this.game.once` is the correct method here. It's the makeshift way we currently do delayed effects. It will set up an event listener for the `afterChallenge` event that will be called the next time that event is emitted, then stop listening. It works well for `afterChallenge` in particular because that event is raised only once per challenge.

The second parameter is not correct currently. `this.afterChallenge.bind(this.challengeToParticipateIn, context)` is difficult to explain the space of a comment, but basically it will attempt to call the `afterChallenge` method defined on the current class using the challenge object (`this.challengeToParticipateIn`) as the `this` object inside that method. Which will crash as it tries to move `this` (which is now the challenge, not the card) to shadows.

The correct way to use bind in this instance would be:
```javascript
this.game.once('afterChallenge', this.afterChallenge.bind(this, this.challengeToParticipateIn, context));
```

However, it would be much simpler to write (and read later as):
```javascript
this.game.once('afterChallenge', () => this.afterChallenge(this.challengeToParticipateIn, context));
```

server/game/cards/13.6-LMHR/RecruitFromTheDungeons.js
```diff
@@ -0,0 +1,28 @@
...
+            }
+        });
+    }
+
+    afterChallenge(challenge, context) {
+        if(challenge.looser === context.player) {
```
ystros: Should be `challenge.loser`

server/game/cards/13.6-LMHR/RecruitFromTheDungeons.js
EvadLhorg: thank you for the feedback!

PR #2761, opened 2019-09-19
Add docker support

cryogen:


cryogen:
Note this is incomplete because we have a split client, I'll have to dockerise the client separately

config/default.json5
```diff
@@ -4,7 +4,10 @@
     secret: 'somethingverysecret',
     hmacSecret: 'somethingevenmoresecret',
 
-    dbPath: 'mongodb://127.0.0.1:27017/throneteki',
+    dbPath: 'mongodb://mongo:27017/keyforge',
```
ystros: Wrong app. 😄

PR #2766, opened 2019-09-23
Few small changes to the timer so it counts down and is more responsive

cryogen:
Fix some issues left over from the config changes

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -734,7 +734,7 @@ class Game extends EventEmitter {
         if(this.useGameTimeLimit) {
             let timeLimitStartType = 'whenSetupFinished'; //todo: change to property of game when more kinds of time limit start triggers are implemented/asked for
             let timeLimitInMinutes = this.gameTimeLimit;
-            this.timeLimit.initialiseTimeLimit(timeLimitStartType, timeLimitInMinutes); 
+            this.timeLimit.initialiseTimeLimit(timeLimitStartType, timeLimitInMinutes);
         }
 
         for(let player of this.getPlayers()) {
@@ -766,9 +766,17 @@ class Game extends EventEmitter {
         }
     }
 
+    checkForTimeExpired() {
+        if(this.timeLimit.isTimeLimitReached && !this.finishedAt) {
+            this.addAlert('success', 'The game has ended because the timer has expired.  Timed wins are not currently implemented');
+            this.finishedAt = new Date();
+        }
+    }
+
     beginRound() {
         // Reset phases to the standard game flow.
         this.remainingPhases = Phases.names();
+
         this.raiseEvent('onBeginRound');
         this.queueSimpleStep(() => {
             // Loop through individual phases, queuing them one at a time. This
@@ -1168,6 +1176,10 @@ class Game extends EventEmitter {
         this.router.rematch(this);
     }
 
+    timeExpired() {
+        this.emit('onTimeExpired');
+    }
+
     activatePersistentEffects() {
         this.effectEngine.activatePersistentEffects();
     }
@@ -1231,7 +1243,8 @@ class Game extends EventEmitter {
                 cancelPromptUsed: this.cancelPromptUsed,
                 useGameTimeLimit: this.useGameTimeLimit,
                 gameTimeLimitStarted: this.timeLimit.timeLimitStarted,
-                gameTimeLimitStartedAt: this.timeLimit.timeLimitStartedAt
+                gameTimeLimitStartedAt: this.timeLimit.timeLimitStartedAt,
+                gameTimeLimitTime: this.timeLimit.timeLimitInMinutes
             };
         }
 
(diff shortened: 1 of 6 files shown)


server/game/gamesteps/taxationphase.js
```diff
@@ -44,6 +44,8 @@ class TaxationPhase extends Phase {
         this.game.addAlert('endofround', 'End of round {0}', this.game.round);
         this.game.addMessage(playerPower);
         this.game.addAlert('startofround', 'Round {0}', this.game.round + 1);
+
+        this.game.checkForTimeExpired();
```
ystros: Per tournament rules, should the game end at the beginning of the next round or at the end of the current round? If the latter, this check should come before the `startofround` alert.

server/game/game.js
```diff
@@ -766,9 +766,17 @@ class Game extends EventEmitter {
         }
     }
 
+    checkForTimeExpired() {
+        if(this.timeLimit.isTimeLimitReached && !this.finishedAt) {
+            this.addAlert('success', 'The game has ended because the timer has expired.  Timed wins are not currently implemented');
```
ystros: It doesn't seem like it would be very complicated to record / announce the winner. The most complicated case is if Tycho Nestoris is in play.

Though we could certainly do that as a separate PR. Unless there's concern tying timed win rules to a particular tournament format?

server/game/game.js
```diff
@@ -766,9 +766,17 @@ class Game extends EventEmitter {
...
+            this.finishedAt = new Date();
```
ystros: Elsewhere, we record a `winReason`. Should we record one here as `'timer'`, even if we make no determination about the winner?

server/gamenode/gameserver.js
```diff
@@ -213,6 +213,9 @@ class GameServer {
 
     onStartGame(pendingGame) {
         let game = new Game(pendingGame, { router: this, titleCardData: this.titleCardData, cardData: this.cardData, packData: this.packData, restrictedListData: this.restrictedListData });
+        game.on('onTimeExpired', () => {
+            this.sendGameState(game);
```
ystros: Having the Game Server listen to events on game makes me slightly nervous. Though I don't have a good alternative (e.g. you'd have to pass the `sendGameState` method into the game object itself)

server/game/timeLimit.js
```diff
@@ -23,25 +23,25 @@ class TimeLimit {
...
     checkForTimeLimitReached() {
         if(this.game.useGameTimeLimit && !this.isTimeLimitReached) {
             let differenceBetweenStartOfTimerAndNow = moment.duration(moment().diff(this.timeLimitStartedAt));
             if(differenceBetweenStartOfTimerAndNow.asSeconds() / 60 >= this.timeLimitInMinutes) {
-                this.game.addMessage('Time limit of {0} minutes reached, the game will end after the current round has finished!', this.timeLimitInMinutes);
+                this.game.addAlert('warning', 'Time up.  The game will end after the current round has finished');
                 this.isTimeLimitReached = true;
-                this.game.on('onRoundEnded', () => this.handleEndOfRoundWhenTimeLimitIsReached());
+                this.timeLimitStarted = false;
+                this.game.timeExpired();
```
ystros: What's the trade off here between defining `timeExpired` here vs doing the `emit` here? Both require the `game` object. Is there any game specific logic we anticipate in the future that should live on `Game` instead of `TimeLimit`?

server/game/game.js
cryogen: I think EvadLhorg is working on that so I didn’t want to step on his toes :)

server/game/game.js
cryogen: I was thinking about that, but wasn’t sure I want to record the result yet until the timed win stuff is added(and there’s no point setting winreason if the result isn’t recorded)

server/game/timeLimit.js
cryogen: I felt that it was better abstraction to not have the timer reaching inside of the game state stuff or raising events on his behalf. What do you think?

server/gamenode/gameserver.js
cryogen: I think I get your concern but it kind of fits the pattern. The server is what sends status updates to the connected users, I’d be less happy about the game itself starting to manage state updates

ystros: Yeah, sending game state should be handled here. The alternative is to send in a `handleTimeExpired` method to `Game` that happens to be implemented as sending state. But that's not a lot different than this.

server/game/timeLimit.js
ystros: Agree on not reaching in for state. Not raising the event makes `TimeLimit` feel like it's a leaky abstraction, as `Game` has to know how to signal that time has expired. 

Part of the problem is we've conflated game state (e.g. list of players and their state) and game messaging bus (e.g. something happened in the game and other components might want to know) into a single giant object. But that's a larger problem than this PR.

PR #2767, opened 2019-09-23
backport admin enhancement from throneteki

cryogen:
fix several bugs that caused users not to be banned properly
Add ban list functionality
Add linked accounts and session info to user admin

### DIFF
--- a/server/models/User.js
+++ b/server/models/User.js
@@ -65,6 +65,10 @@ class User {
         return this.userData.verified;
     }
 
+    get registered() {
+        return this.userData.registered;
+    }
+
     get isAdmin() {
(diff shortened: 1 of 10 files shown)


server/models/User.js
```diff
@@ -110,6 +114,16 @@ class User {
         return this.blockList.includes(otherUser.username.toLowerCase());
     }
 
+    getFullDetails() {
+        let user = Object.assign({}, this.userData);
+
+        delete user.password;
```
ystros: Is the password the only thing that needs to be removed? Isn't there also a list of current tokens?

cryogen: This is used in the admin page, which actually uses the session info, so this is ok

PR #2772, opened 2019-09-29
Implement determining winner after time limit expired

EvadLhorg:


server/game/game.js
```diff
@@ -1303,6 +1303,84 @@ class Game extends EventEmitter {
             })
         };
     }
+
+    determineWinnerAfterTimeLimitExpired() {
+        //collect relevant summary of a player in one array, only players who can win the game
+        const playerSummaries = this.getPlayers().map(player => {
```
ystros: This method is sufficiently complex in its current state that I think it could benefit from moving to its own class.

I also think you could make this simpler / more readable by writing out the sequence of rules that determine winner, then continuously filtering the player array:

```javascript
const rules = [
    player => player.canWinGame(),
    player => player.getTotalPower() >= highestPower,
    player => player.drawDeck.length >= highestDrawDeck,
    player => player === wonDominanceThisRound,
    player => player.deadPile.length <= lowestDeadPile,
    player => player.firstPlayer
];

for(const rule of rules) {
    // filter remaining players, declare winner as soon as you only have 1 player left.
}
```

You wouldn't need to create the player summaries at all, just calculate the highest power/draw deck lowest dead pile etc. Just loop through the rules, and record the winner / exit as soon as there's exactly 1 player remaining. If after filtering with all the rules there are no players left, then it's a draw. (If after filtering there are still multiple players left, something went terribly wrong as it means there were two first players)

server/game/gamesteps/dominancephase.js
```diff
@@ -35,6 +35,8 @@ class DominancePhase extends Phase {
         }
 
         if(dominanceWinner) {
+            //save the winner of dominance on the game object in order to use this information in determining the winner of the game after the time limit has expired
+            this.game.winnerOfDominanceInLastRound = dominanceWinner;
```
ystros: Good call on placing this outside the `if`. Rules state you can win dominance even if you can't gain power. 👍

test/server/integration/TimeLimitExpired.spec.js
```diff
@@ -0,0 +1,96 @@
...
+                });
+    
+                it('based on which player was the first player in the last round', function() {
+                    this.game.determineWinnerAfterTimeLimitExpired();
+                    expect(this.game.finishedAt).toBeDefined();
+                    expect(this.game.winner).toBe(this.player1Object);
+                });
+    
+                it('when one player cannot win the game', function() {
+                    this.tycho = this.player1.findCardByName('Tycho Nestoris', 'hand');
```
ystros: Given how many non-Tycho cards are in this test deck, he's not guaranteed to be in hand here. If you just omit the second parameter, it should search through all areas for the player to find a copy.

PR #2775, opened 2019-10-02
Fix The Queens Retinue interacting with Barring the Gates

### LINKED ISSUE #2774: Queen's Retinue draws infinite with Barring the Gates flipped
Queen's Retinue's effect is stated as it enters play, and then the opponent draws. If barring is up, it can't enter play and the opponent does not draw. But as it stands right now on Iron Throne, if barring is flipped, the retinue does not enter play if you click it, but your opponent still is forced to draw two cards.

EvadLhorg:
fixes #2774

### DIFF
--- a/server/game/cards/13.6-LMHR/TheQueensRetinue.js
+++ b/server/game/cards/13.6-LMHR/TheQueensRetinue.js
@@ -5,16 +5,15 @@ class TheQueensRetinue extends DrawCard {
         this.action({
             title: 'Put into play',
             location: 'shadows',
+            condition: context => context.player.canPutIntoPlay(this, 'outOfShadows'),
             handler: context => {
                 context.player.putIntoPlay(this);
-                if(this.location === 'play area') {
-                    for(let opponent of this.game.getOpponents(this.controller)) {
-                        if(opponent.canDraw()) {
-                            opponent.drawCardsToHand(2);
-                        }
+                for(let opponent of this.game.getOpponents(this.controller)) {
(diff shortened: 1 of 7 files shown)


server/game/cards/13.6-LMHR/TheQueensRetinue.js
```diff
@@ -7,12 +7,14 @@ class TheQueensRetinue extends DrawCard {
             location: 'shadows',
             handler: context => {
                 context.player.putIntoPlay(this);
-                for(let opponent of this.game.getOpponents(this.controller)) {
-                    if(opponent.canDraw()) {
-                        opponent.drawCardsToHand(2);
+                if(this.location === 'play area') {
```
ystros: Thanks for picking this up so quick!

This would work right now. In the near term future, every game action will be asynchronous. In this instance, `putIntoPlay` would generate on `onCardEntersPlay` event and the location wouldn't be set to 'play area' until after the interrupts for that event have been resolved. So putting into play and immediately doing an `if` would fail.

The better check would be to prevent the ability from triggering when the card can't be put into play. This could be done by adding a `condition` to the ability that checks that `this` can be put into play. Check the other similar "put into play from shadows for some other cost" cards.

EvadLhorg: Thank you for your feedback. I also checked the other 4 for 4 bicon shadow (4) cards every faction got and added/fixed the condition on them.

PR #2776, opened 2019-10-03
fix the order of the message parameters on the faiths decree

EvadLhorg:
observing a game i noticed that the message parameters were in the wrong order


ystros:
Good catch!

Issue #2693, opened 2019-07-25
Implement Dragons of the East box


ystros:
* [x] 15001 - [Daenerys Targaryen](https://thronesdb.com/card/15001)
* [x] 15002 - [Drogon](https://thronesdb.com/card/15002)
* [x] 15003 - [Rhaegal](https://thronesdb.com/card/15003)
* [x] 15004 - [Viserion](https://thronesdb.com/card/15004)
* [x] 15005 - [Khal Drogo](https://thronesdb.com/card/15005)
* [x] 15006 - [Galazza Galare](https://thronesdb.com/card/15006)
* [x] 15007 - [Jhogo](https://thronesdb.com/card/15007)
* [x] 15008 - [The Shavepate](https://thronesdb.com/card/15008)
* [x] 15009 - [Rakharo](https://thronesdb.com/card/15009)
* [x] 15010 - [The Titan's Bastard](https://thronesdb.com/card/15010)
* [x] 15011 - [Barsena Blackhair](https://thronesdb.com/card/15011)
* [x] 15012 - [Captain Groleo](https://thronesdb.com/card/15012)
* [x] 15013 - [Blue Graces](https://thronesdb.com/card/15013)
* [x] 15014 - [Dothraki Handmaiden](https://thronesdb.com/card/15014)
* [x] 15015 - [Horselord](https://thronesdb.com/card/15015)
* [x] 15016 - [Brown Ben Plumm](https://thronesdb.com/card/15016)
* [x] 15017 - [Womb of the World](https://thronesdb.com/card/15017)
* [x] 15018 - [Pentos](https://thronesdb.com/card/15018)
* [x] 15019 - [Khaleesi](https://thronesdb.com/card/15019)
* [x] 15020 - [Dragon Egg](https://thronesdb.com/card/15020)
* [x] 15021 - [Pit Fighter's Blade](https://thronesdb.com/card/15021)
* [x] 15022 - [Overwhelming Numbers](https://thronesdb.com/card/15022)
* [x] 15023 - [Fire Made Flesh](https://thronesdb.com/card/15023)
* [x] 15024 - [Incinerate](https://thronesdb.com/card/15024)
* [x] 15025 - [Delena Florent](https://thronesdb.com/card/15025)
* [x] 15026 - [Dragonstone Convert](https://thronesdb.com/card/15026)
* [x] 15027 - [Hotho Humpback](https://thronesdb.com/card/15027)
[...]

PR #2799, opened 2019-10-31
implement blue graces

EvadLhorg:
progress towards #2693

### DIFF
--- a/server/game/cards/15-DotE/BlueGraces.js
+++ b/server/game/cards/15-DotE/BlueGraces.js
@@ -1,0 +1,26 @@
+const DrawCard = require('../../drawcard.js');
+
+class BlueGraces extends DrawCard {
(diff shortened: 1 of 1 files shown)


server/game/cards/15-DotE/BlueGraces.js
```diff
@@ -0,0 +1,26 @@
...
+
+class BlueGraces extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Shuffle card back into deck',
+            phase: 'challenge',
+            cost: ability.costs.removeSelfFromGame(),
+            target: {
+                activePromptTitle: 'Select a card',
+                cardCondition: card => (card.location === 'discard pile' || card.location === 'dead pile') &&
```
ystros: The card text specifies your discard or dead pile, so controller needs to be checked.

PR #2802, opened 2019-11-02
implement pentos

EvadLhorg:
progress towards #2693

### DIFF
--- a/server/game/cards/15-DotE/Pentos.js
+++ b/server/game/cards/15-DotE/Pentos.js
@@ -1,0 +1,27 @@
+const DrawCard = require('../../drawcard.js');
+
+class Pentos extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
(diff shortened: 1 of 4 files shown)


server/game/cards/15-DotE/Pentos.js
```diff
@@ -0,0 +1,27 @@
+const DrawCard = require('../../drawcard.js');
+
+class Pentos extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onCardsDrawn: event =>
+                    event.source.controller === this.controller &&
+                    event.cards[0].getType() === 'attachment' &&
```
ystros: Unfortunately this only works if you're drawing a single card and it happens to be an attachment. It should also cover cases where you draw 2 cards at once and the attachment is second, or if you draw multiple attachments at once (in which case you need to be able to choose which attachment to put into play). 

I'll take a closer look at this one as it may require modifying the underlying events.

PR #2820, opened 2019-11-04
implement champions favor

EvadLhorg:
progress towards #2693


server/game/cards/15-DotE/ChampionsFavor.js
```diff
@@ -0,0 +1,32 @@
+const DrawCard = require('../../drawcard.js');
+
+class ChampionsFavor extends DrawCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction({ trait: 'Lady' });
+        this.action({
+            title: 'Give STR',
+            condition: () => this.isNonLadyParticipating(),
```
ystros: You shouldn't need this condition, as it's automatically covered when the engine checks if there are valid targets.

PR #2829, opened 2019-11-06
implement lay waste

EvadLhorg:
progress towards #2693 

i am not 100% sure if the `then(...)` part works like i want it to. I have tried writing a test but i only get to the discarding part although i can observe it entering the `handler` function of the `then ` part.

Test is not included as it is based on a different throneteki-json-data revision than what master is currently on.

### DIFF
--- a/server/game/cards/15-DotE/LayWaste.js
+++ b/server/game/cards/15-DotE/LayWaste.js
@@ -1,0 +1,50 @@
+const DrawCard = require('../../drawcard');
+const GameActions = require('../../GameActions');
+
+class LayWaste extends DrawCard {
+    setupCardAbilities() {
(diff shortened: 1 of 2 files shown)


ystros:
This looks basically correct, but I need to modify the Search game action to support players other than the player that triggered the ability. The `player: context.target.owner` currently won't do anything.

ystros:
I pushed an updated version based on changes to search on the Galazza branch. Tested it out manually and it seems to work. Feel free to add your tests, even if they won't go green until the submodule data is merged.

EvadLhorg:
Thanks for your commit, i added the test and was able to get it to work and test the expected behavior.

PR #2848, opened 2019-11-13
fix The Iron Throne bug when no card name was entered

EvadLhorg:
add check for cardName to have a useable value

### DIFF
--- a/server/game/cards/13.6-LMHR/TheIronThrone.js
+++ b/server/game/cards/13.6-LMHR/TheIronThrone.js
@@ -22,14 +22,16 @@ class TheIronThrone extends DrawCard {
 
     selectCardName(player, cardName) {
         this.game.addMessage('{0} kneels {1} to name {2}', player, this, cardName);
-        this.untilEndOfPhase(ability => ({
-            targetController: 'any',
-            effect: [
-                ability.effects.cannotMarshal(card => card.name.toLowerCase() === cardName.toLowerCase()),
-                ability.effects.cannotPlay(card => card.name.toLowerCase() === cardName.toLowerCase()),
-                ability.effects.cannotPutIntoPlay(card => card.name.toLowerCase() === cardName.toLowerCase())
-            ]
-        }));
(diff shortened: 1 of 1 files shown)


server/game/cards/13.6-LMHR/TheIronThrone.js
```diff
@@ -22,14 +22,16 @@ class TheIronThrone extends DrawCard {
...
         this.game.addMessage('{0} kneels {1} to name {2}', player, this, cardName);
-        this.untilEndOfPhase(ability => ({
-            targetController: 'any',
-            effect: [
-                ability.effects.cannotMarshal(card => card.name.toLowerCase() === cardName.toLowerCase()),
-                ability.effects.cannotPlay(card => card.name.toLowerCase() === cardName.toLowerCase()),
-                ability.effects.cannotPutIntoPlay(card => card.name.toLowerCase() === cardName.toLowerCase())
-            ]
-        }));
+        if(cardName) {
```
ystros: I think it would be better to handle this at the prompt level, rather than directly here. There are other "Name a card" abilities that likely have the same problem, they're just currently less popular than the Iron Throne.

Ideally, it should work the same as clicking "Done" when targeting without selecting a card - it should cancel the overall ability and print a warning message to game chat.

It would also be good to understand how / why so many people are submitting blank values. I would have thought, at worst, a blank string would be submitted, not null.

EvadLhorg: Need a little help as i don´t know how i would implement the check at the prompt level to make it available for all prompts. Would you be so kind to push me in the right direction?

ystros: Take a look at `IconPrompt` and `Game#promptForIcon` for an example of separating it out into it's own prompt. The cancellation message can be seen here (but you'd put it in the new prompt's onCancel handler): https://github.com/throneteki/throneteki/blob/master/server/game/gamesteps/abilityresolver.js#L125

PR #2882, opened 2019-12-25
Fix Champion's Challenge can be triggered multiple times per phase

atkeranov:
Fix Champion's Challenge can be triggered multiple times per phase if the character strength condition has been met. The plot says "Limit once per phase" so this restriction should be applied as well.


ystros:
Thanks!

PR #2933, opened 2020-02-21
Refactor boolean effects into flags

### LINKED ISSUE #2310: Effects should not use boolean flags
Effects that set and unset boolean flags can remove other instances of that effect when they leave play.

* Both players reveal Withering Cold. Player 1 uses Rains of Castamere to reveal a new plot. The effect on Player 2's Withering Cold is undone, so the standing phase is no longer skipped.
* Blood of the Dragon is out, and a character receives an additional burn effect (Dracarys, Plaza of Punishment, etc). When the phase ends, the burn effect for Blood of the Dragon will be removed from that character. So if the character's STR is reduced to 0 somehow in the new phase, it won't be killed as expected.

To fix this, these properties should be reference counted instead, probably using `ReferenceCountedSetProperty` to manage them as a group.

ystros:
Fixes #2310


cryogen:
There's test failures on this, but it's possible I did the conflict resolution wrong

PR #2988, opened 2020-05-22
[bugfix] Fix string interpolation in Dornish Man's wife message

dokempf:
The message of Dornish Man's Wife used faulty string interpolation,
which resulted e.g. in the following text (copied from game):

hechtlinger uses "The Dornishman's Wife" to choose Harren and gain {1} gold,
2 , and draw 1 card.


ystros:
Thanks for submitting this! It seems like this doesn't fix the problem:
Screen Shot 2020-05-23 at 9 32 51 PM

I think this is because of two factors:
1. The second argument when `push`ing the message onto the array is ignored.
2. Nested rendering of messages like this on early cards was not really solidified.

Probably the best way to do this now is to require the `Message` class and use `Message.fragment` to create a nested message. See Friends at Court as an example:

https://github.com/throneteki/throneteki/blob/9a00dfc152c732a587446a1d78f31c8dd88aa2f4/server/game/cards/13.6-LMHR/FriendsAtCourt.js#L40-L43

PR #2991, opened 2020-05-25
New Gameoption: Disable chat for spectators

EvadLhorg:
In worldcup games to prevent random people joining the game and spamming the chat, most games are created with a standard password, but opinions differ on which password should be used and it is a real pain in the ass to figure out the "correct" "standard" password... So i thought to add an option to games that disables the chat for spectators to have a greater incentive to host games without passwords. 
Please also look at the client PR!


EvadLhorg:
Closed in favor of a direct port of the same feature from keyteki

PR #2999, opened 2020-06-12
Fix Pulling the Strings + At the Gates interaction

ystros:
Gets rid of `this.controller` within each plot that Pulling the Strings can interact with, so we no longer need to do a fake take-control to initiate the ability. This means that the At The Gates plot being initiated no longer counts as a City plot in your own used pile.


cryogen:
There comes a time in every *teki implementation, when *this* has to be hunted down and destroyed...

PR #3013, opened 2020-06-30
Chessclocks

EvadLhorg:
Players can now choose to use a chess clock to time their games.
Chess clocks provide a time limit per player. The time will start running if a player has an active prompt and after an initial delay of 5 seconds.
There are also two new chat commands:

- /pause-clock to pause the chess clocks (other time limit not yet implemented) so both clocks will stop running until /pause-clock is called a second time.
- /add-chess-clock-time to add time to the calling players clock (in case someone disconnects and the clock keeps running and players agree to add the lost time)

### DIFF
--- a/server/game/ChessClock.js
+++ b/server/game/ChessClock.js
@@ -1,0 +1,98 @@
+class ChessClock {
+    constructor(player, time, delayToStartClock) {
+        this.player = player;
+        this.mainTime = time * 60;
+        this.timeLeft = time * 60;
+        this.mode = 'inactive';
+        this.timerStart = 0; //when the clock is running, this contains the instant on which it started
+        this.paused = false; //has the clock been paused via chat command?
+        this.running = false; //is the clock ticking for the player?
+        this.stateId = 0;
+        this.name = 'ChessClock';
+        this.activated = false; //indicates if the game has started/setup is finished 
+        this.delayToStartClock = delayToStartClock;
+
+        this.player.game.on('onSetupFinished', () => this.activateChessClock());
+    }
+
+    activateChessClock() {
+        this.activated = true;
+    }
+
+    togglePause() {
+        if(this.activated) {
+            this.paused = !this.paused;
+            if(this.paused) {
+                this.stop();
+            }
+        }
+    }
+
+    modify(secs) {
+        this.timeLeft += secs;
+    }
+
+    updateStateId() {
+        this.stateId++;
+    }
+
+    start() {
+        if(this.activated) {
+            if(!this.paused && !this.running) {
+                this.mode = 'down';
+                this.running = true;
+                this.timerStart = Date.now();
+                this.updateStateId();
+            }
+        }
+    }
+
+    stop() {
+        if(this.activated) {
+            if(this.timerStart > 0 && this.running) {
+                this.running = false;
+                this.updateTimeLeft(Math.floor(((Date.now() - this.timerStart) / 1000) + 0.5));
+                this.timerStart = 0;
+                this.updateStateId();
+            }
+            this.mode = 'stop';
+        }
+    }
+
+    timeRanOut() {
+        this.player.game.addMessage('{0}\'s clock has run out', this.player);
+        this.player.game.recordWinner(this.player.game.getOpponents(this.player)[0], 'time');
+        return;
+    }
+
+    updateTimeLeft(secs) {
+        if(this.timeLeft === 0 || secs < 0) {
+            return;
+        }
+        if(secs <= this.delayToStartClock) {
+            return;
+        }
+        
+        secs = secs - this.delayToStartClock;
+        if(this.mode === 'down') {
+            this.modify(-secs);
+            if(this.timeLeft < 0) {
+                this.timeLeft = 0;
+                this.timeRanOut();
+            }
+        }
+    }
+
+    getState() {
+        return {
+            mode: this.mode,
+            timeLeft: this.timeLeft,
+            stateId: this.stateId,
+            mainTime: this.mainTime,
+            name: this.name,
+            delayToStartClock: this.delayToStartClock
+        };
+    }
+}
+
+module.exports = ChessClock;
--- a/server/game/ChessClock.js
+++ b/server/game/ChessClock.js
@@ -20,11 +20,12 @@ class ChessClock {
     }
 
     togglePause() {
-        if(this.activated) {
-            this.paused = !this.paused;
-            if(this.paused) {
-                this.stop();
-            }
+        if(!this.activated) {
+            return;
+        }
+        this.paused = !this.paused;
+        if(this.paused) {
+            this.stop();
         }
     }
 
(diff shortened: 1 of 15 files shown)


server/game/gamesteps/uiprompt.js
```diff
@@ -14,14 +14,25 @@ class UiPrompt extends BaseStep {
 
     complete() {
         this.completed = true;
+        if(this.getPlayer()) {
+            if(this.getPlayer().stopClock) {
```
ystros, line 18: Are the checks like this necessary? I would think that a player object would always have a `stopClock` and `startClock` methods.

server/game/game.js
```diff
@@ -1225,6 +1229,18 @@ class Game extends EventEmitter {
...
+    pauseClock() {
+        this.clockPaused = !this.clockPaused;
+        if(this.useChessClocks) {
+            let players = this.getPlayers();
+            for(let player of players) {
+                player.togglePauseChessClock();
+            }
+        }
+
+        //TODO pause time limit clock, not only chess clock
```
ystros, line 1241: I'm wondering in general if there's a way for the clock mechanism to be generalized. e.g. use the same interface for both time limit clock and chess clock, so that the game logic doesn't need to care so much about the details of which clock is being used.

server/game/gamesteps/uiprompt.js
```diff
@@ -87,6 +98,13 @@ class UiPrompt extends BaseStep {
      */
     onCompleted() {
     }
+
+    /**
+     * Will be implemented in sub classes that have a specific player that is using the prompt
+     */
+    getPlayer() {
```
ystros, line 105: I do notice that `AllPlayerPrompt`, which inherits from this class, does not implement this method. It's used for at least the plot selection prompt, which seems like it should be subject to timers. Any thoughts on making it compatible with the timer?

(you could have this method return an array of players that are currently making choices instead, for instance)

server/game/ChessClock.js
```diff
@@ -0,0 +1,98 @@
...
+        this.player = player;
+        this.mainTime = time * 60;
+        this.timeLeft = time * 60;
+        this.mode = 'inactive';
+        this.timerStart = 0; //when the clock is running, this contains the instant on which it started
+        this.paused = false; //has the clock been paused via chat command?
+        this.running = false; //is the clock ticking for the player?
+        this.stateId = 0;
+        this.name = 'ChessClock';
+        this.activated = false; //indicates if the game has started/setup is finished
```
ystros, line 12: Is this rule (no clocks until after set up) consistent with the current time limit clock?

server/game/ChessClock.js
```diff
@@ -0,0 +1,98 @@
...
+        this.timeLeft += secs;
+    }
+
+    updateStateId() {
+        this.stateId++;
+    }
+
+    start() {
+        if(this.activated) {
+            if(!this.paused && !this.running) {
```
ystros, line 41: Small stylistic choice, but perhaps early returns would be easier to read than these nested `if` statements?

server/game/ChessClock.js
```diff
@@ -0,0 +1,98 @@
...
+                this.timerStart = 0;
+                this.updateStateId();
+            }
+            this.mode = 'stop';
+        }
+    }
+
+    timeRanOut() {
+        this.player.game.addMessage('{0}\'s clock has run out', this.player);
+        this.player.game.recordWinner(this.player.game.getOpponents(this.player)[0], 'time');
```
ystros, line 64: This is not melee friendly. I don't think we need to make it so right now, but perhaps leave a TODO comment so we don't lose track of it?

server/game/gamesteps/uiprompt.js
EvadLhorg: I found out that some of the player objects created during tests are not "real" player objects and only have specific methods. Those tests fail without those checks in place. Another way of fixing it would be to change the creation of the player objects in those particular tests. What would you recommend?

server/game/ChessClock.js
EvadLhorg: yes. "normal" time limit is started after setup is finished

server/game/ChessClock.js
EvadLhorg: Good idea, TODO added

server/game/ChessClock.js
EvadLhorg: good point, fixed

server/game/gamesteps/uiprompt.js
EvadLhorg: Good point regarding the AllPlayerPrompt and its sub-classes. I made a re-test to make sure the plot phase was working "as expected" as in when one player confirms his/her plot, the timer stops. It does work like that at the moment, but the reason for that is not the stopClock call in the completed() method of uiprompt but the stopClock call in the setPrompt() method for the waitingPrompt case. 
After figuring that out, i questioned the necessity of the stopClock call in completed() and the whole getPlayer() stuff. Testing that again, i still need that stopClock in completed for situations like initiating a challenge and having the first action in the after-attacker-declared action window. Without the stopping in completed, the attacking players clock would only stop after it´s the other players turn for taking actions in the after-attackers-declared window.
So all in all, it is not 100% necessary to have the getPlayer implemented in the allPlayerPrompt, but it won´t hurt to have it there. I will follow your advice and return an array of players and stop the clock for all players in the array

server/game/game.js
EvadLhorg: The Keyteki project is doing that (as they have even more types of clocks) but it does make the code a bit more complicated as to what part of code is doing work for which type of clock. But using at least the same interface would be a good idea. I will open up a follow-up issue for that.

server/game/gamesteps/uiprompt.js
cryogen: The test objects are mocks, if you want them to have those methods you need to add it to the mock. Shouldn't be too difficult to set up

EvadLhorg: If i understand correctly, this code creates the mock for a player object: jasmine.createSpyObj('player' ...)
Searching for instances of this code, i find 32 results. Is there an easier way to add the methods to the mocks than adding stop and startClock to each of those 32 mock creation calls? Maybe a helper method for creating player mocks with default methods in addition to those that are added as a parameter?

cryogen: It should only be added where it's needed, if that happens to be all 32 then that's unfortunate :) But yes in that case it would probably make sense to refactor the thing to all get created in one place

EvadLhorg: Fixed it now for the instances where it was needed.

server/game/gamesteps/uiprompt.js
```diff
@@ -14,14 +14,25 @@ class UiPrompt extends BaseStep {
 
     complete() {
         this.completed = true;
+        if(this.getPlayer()) {
+            // if(this.getPlayer().stopClock) {
```
ystros, line 18: Current build is failing due to linting errors. Remove the commented out ifs and re-indent.

EvadLhorg: thanks. fixed

PR #3033, opened 2020-08-04
fix fight to the last so it saves against burn

### LINKED ISSUE #3032: Fight to the Last doesn't save from burn
Hey pals, thanks for all you do for this site!

Just had a game where I didn't get a prompt to use the Fight to the Last in my hand to save from a Drogon (IDP) trigger, which should work as the save sets STR to 1, preventing immediate death from terminal burn. A spectator said they'd found it wasn't implemented properly either. Seems to be fine for other methods of death.

EvadLhorg:
resolves #3032

### DIFF
--- a/server/game/cards/15-DotE/FightToTheLast.js
+++ b/server/game/cards/15-DotE/FightToTheLast.js
@@ -6,14 +6,15 @@ class FightToTheLast extends DrawCard {
         this.interrupt({
             canCancel: true,
             when: {
-                onCharacterKilled: event => event.allowSave && event.card.canBeSaved() && event.card.controller === this.controller
+                onCharacterKilled: event => (event.allowSave || event.isBurn) && event.card.canBeSaved() && event.card.controller === this.controller
             },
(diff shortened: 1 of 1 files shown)


server/game/cards/15-DotE/FightToTheLast.js
```diff
@@ -6,14 +6,15 @@ class FightToTheLast extends DrawCard {
...
+                onCharacterKilled: event => (event.allowSave || event.isBurn) && event.card.canBeSaved() && event.card.controller === this.controller
             },
             location: 'hand',
             handler: context => {
                 context.event.saveCard();
+                let savedCard = context.event.card;
                 if(context.event.card.kneeled && context.event.card.allowGameAction('stand')) {
                     this.game.resolveGameAction(
-                        GameActions.standCard(context.event.card),
+                        GameActions.standCard(() => ({ card: savedCard })),
```
ystros, line 17: Good catch, surprised no one has complained about this portion before.

PR #3034, opened 2020-08-05
fix improved fortifications so it can´t save against non-saveable effects anymore

### LINKED ISSUE #3019: Bug: Improved Fortifications can save against the sacrifice of The Kingsroad
As title says. You are not allowed to save against a sacrifice effect according to the rules, see https://thronesdb.com/rulesreference#Sacrifice

EvadLhorg:
resolves #3019

### DIFF
--- a/server/game/GameActions/DiscardCard.js
+++ b/server/game/GameActions/DiscardCard.js
@@ -20,7 +20,7 @@ class DiscardCard extends GameAction {
         };
         return this.event('onCardDiscarded', params, event => {
             event.cardStateWhenDiscarded = event.card.createSnapshot();
-            event.card.controller.moveCard(event.card, 'discard pile');
+            event.card.controller.moveCard(event.card, 'discard pile', {allowSave: allowSave});
         });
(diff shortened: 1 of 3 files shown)


ystros:
I think more is required to fix Improved Fortifications, though I'm surprised by the test failures here.

EvadLhorg:
I added some stuff now but i am still very confused by the varys test or to be more specific the interaction between Arya Stark (TFM)´s no save effect and the interaction with dupes. The addition of {allowSave: allowSave} as a parameter to the moveCard call breaks the Varys test and it feels like the no-save vs dupe interaction only worked before because of a bug. Please help :D

ystros:
I ran into similar issues when I tried fixing this issue. I think the *real* answer is to make the onCardLeftPlay event simultaneous to the other card move events. Which is a larger, harder change that does need to happen, but it's so large it's hard to get traction on.

EvadLhorg:
not necessary anymore due to #3188 being merged

PR #3035, opened 2020-08-05
make player.shuffleCardIntoPlay raise an onCardReturnedToDeck event so manually saving with dupes works

### LINKED ISSUE #3016: Winterfell Crypt doesn't allow saves
I had a situation where I had a Viserion with a duplicate that had been chosen for a Winterfell Crypt. At the end of the phase, I was not given the choice to save the Viserion which was just shuffled in my deck (the duplicate was discarded)

EvadLhorg:
resolves #3016


EvadLhorg:
Dug this PR out and finally made some more changes. I think the new ReturnCardToDeck game action does the trick in combination with the old ShuffleIntoDeck gameaction. The key to the manual dupes being able to be used is the event with the name onCardReturnedToDeck (which wasn´t raised by player.shuffleCardIntoDeck before the PR)

server/game/GameActions/ReturnCardToDeck.js
```diff
@@ -0,0 +1,26 @@
...
+
+    createEvent({ card, allowSave = true, bottom = false }) {
+        let params = {
+            card: card,
+            allowSave: allowSave,
+            automaticSaveWithDupe: true,
+            bottom: bottom
+        };
+        return this.event('onCardReturnedToDeck', params, event => {
+            event.cardStateWhenMoved = card.createSnapshot();
```
ystros, line 20: Use the `snapshotName` property you introduced a while back instead. 😄

server/game/GameActions/index.js
```diff
@@ -62,7 +63,8 @@ const GameActions = {
         return new SimultaneousAction(actions);
     },
     standCard: props => new AbilityAdapter(StandCard, props),
-    takeControl: props => new AbilityAdapter(TakeControl, props)
+    takeControl: props => new AbilityAdapter(TakeControl, props),
+    returnCardToDeck: props => new AbilityAdapter(ReturnCardToDeck, props),
```
ystros, line 67: Try to keep this (and the corresponding `require` at the top) in alphabetical order.

server/game/player.js
```diff
@@ -893,21 +894,11 @@ class Player extends Spectator {
     }
 
     moveCardToTopOfDeck(card, allowSave = true) {
-        this.game.applyGameAction('moveToTopOfDeck', card, card => {
-            this.game.raiseEvent('onCardReturnedToDeck', { player: this, card: card, allowSave: allowSave }, event => {
-                event.cardStateWhenMoved = card.createSnapshot();
-                this.moveCard(card, 'draw deck', { allowSave: allowSave });
-            });
-        });
+        return this.game.resolveGameAction(ReturnCardToDeck, { card, allowSave });
```
ystros, line 897: The previous `'moveToTopOfDeck'` action is referenced in the implementation of When I Woke.

There are a couple options here:
1. Introduce game actions for both placing on top of deck and bottom of deck (could be same basic implementation, just need the game action name configurable)
2. Update When I Woke to reference "returnCardToDeck" instead
3. Remove the gameAction reference in When I Woke entirely.

I lean toward 3, since there are no cards that provide immunity from being placed on top of deck.

server/game/player.js
```diff
@@ -893,21 +894,11 @@ class Player extends Spectator {
...
     }
 
     moveCardToBottomOfDeck(card, allowSave = true) {
-        this.game.applyGameAction('moveToBottomOfDeck', card, card => {
-            this.game.raiseEvent('onCardReturnedToDeck', { player: this, card: card, allowSave: allowSave }, event => {
-                event.cardStateWhenMoved = card.createSnapshot();
-                this.moveCard(card, 'draw deck', { bottom: true, allowSave: allowSave });
-            });
-        });
+        return this.game.resolveGameAction(ReturnCardToDeck, { card, allowSave, bottom: true });
```
ystros, line 901: Might as well do `GameActions.returnCardToDeck` here like you did with shuffleIntoDeck, rather than include the specific game action.

ystros (changes requested):

server/game/GameActions/ReturnCardToDeck.js
EvadLhorg: fixed that and also merged in the master so the whole snapshotName stuff is working

server/game/GameActions/index.js
EvadLhorg: fixed

server/game/player.js
EvadLhorg: I went with 3 as this was the easiest to do, also tested if when i woke still works

server/game/player.js
EvadLhorg: fixed

PR #3037, opened 2020-08-12
Add event game options and list of valid spectators

EvadLhorg:
I made some extensions to the event functionality:
1. I added game options to be set for every game of the event. These options will be set when a user chooses an event in the event dropdown of a new game and will be fixed so the user can´t change these.
2. I added the option to have a list of valid spectators for an event so that judges/streamers can be put on that list and join any game of that event but no regular users not on that list can´t.

(also have a look at the corresponding client PR)

grafik

grafik

### DIFF
--- a/server/api/events.js
+++ b/server/api/events.js
@@ -17,8 +17,8 @@ module.exports.init = function(server, options) {
             return res.status(403).send({ message: 'Forbidden' });
         }
 
-        const { name, useDefaultRestrictedList, useEventGameOptions, eventGameOptions, restricted, banned } = req.body.event;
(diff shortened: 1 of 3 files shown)


server/api/events.js
```diff
@@ -17,8 +17,8 @@ module.exports.init = function(server, options) {
             return res.status(403).send({ message: 'Forbidden' });
         }
 
-        const { name, useDefaultRestrictedList, restricted, banned } = req.body.event;
-        const event = { name, useDefaultRestrictedList, restricted, banned, pods: [] };
+        const { name, useDefaultRestrictedList, useEventGameOptions, eventGameOptions, restricted, banned, restrictSpectators, validSpectators } = req.body.event;
```
cryogen, line 20: Seems like it's time to put prettier on this codebase too :D

PR #3046, opened 2020-10-03
Add redesigns

EvadLhorg:
* Adding the redesigned cards
* Adding updated copies of the tests related to the redesigned cards
* updating reference to throneteki-json-data to have access to R.json
* updating engine changes related to the redesigned cards
* updated existing tests so that they use the original versions of the redesigned cards

redesignes co-authored by @HannoDiomedes , @ystros , @jzimmer91 , @Martozar , @Duckbringer

### DIFF
--- a/test/server/integration/factionchange.spec.js
+++ b/test/server/integration/factionchange.spec.js
@@ -51,54 +51,54 @@ describe('faction change', function() {
                     expect(this.character.getStrength()).toBe(2);
                 });
             });
-            
-            describe('when a character with gained affiliation is killed', function() {
-                beforeEach(function() {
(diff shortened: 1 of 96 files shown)


test/server/integration/factionchange.spec.js
```diff
@@ -52,54 +52,5 @@ describe('faction change', function() {
                 });
             });
         });
-
-        describe('when a character with gained affiliation is killed', function() {
```
ystros, line 56: I'm not sure this should have been removed?

ystros (commented):
Overall looks good, though not sure why the test around faction change was removed?

PR #3057, opened 2020-10-18
Fix cersei missandei interaction

### LINKED ISSUE #3005: Cersei + Missandei = Server Error
There seems to be some issue with Cersei Lannister when interacting with cards that interrupt their own discard effect. Saw instances of it with both Missandei and Ser Jorah Mormont. See attached screenshots:

Screen Shot 2020-06-24 at 10 14 06
Screen Shot 2020-06-24 at 10 09 27

EvadLhorg:
resolves #3005 

fix the cersei missandei dicard interrupt replacement handler interaction by putting the creation of the cardStateWhenDiscarded snapshot in the params of the event instead of the handler so the creation doesn´t get replaced by missandei´s replacement handler

### DIFF
--- a/server/game/GameActions/DiscardCard.js
+++ b/server/game/GameActions/DiscardCard.js
@@ -17,7 +17,7 @@ class DiscardCard extends GameAction {
             originalLocation: card.location,
             isPillage: !!isPillage,
             source: source,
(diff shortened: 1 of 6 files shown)


ystros (approved):
🚀

PR #3060, opened 2020-10-26
[bugfix] Add limit of once per challenge to redesignbound for the wall

dokempf:
This limit disappeared in the reimplementation. I added exactly the
code that the original implementation is using.


ystros:
Good catch. Thanks!

PR #3063, opened 2020-11-06
sort the restrictedList array by release date

EvadLhorg:
this leads to the redesigned RL coming first and being the default selection for client side dropdowns


ystros (changes requested):
There is sorting of the RLs already happening in `CardService`. It should be adjusted there instead of the API. It's doing so by the card set instead of date.

PR #3071, opened 2020-11-25
add Forgotten Heroes cards

EvadLhorg:
add Forgotten Heroes cards and related tests
add bug fixes for HFK and Mag the Mighty
fix plot phase/plot entering the used pile interaction for Sand Steed and new The Tattered Prince card so reactions to it trigger AFTER when revealed effects

code changes co-authored by @HannoDiomedes

TODO: update throneteki-json-data to latest commit after [89](https://github.com/throneteki/throneteki-json-data/pull/89) is merged

### DIFF
--- a/server/game/cards/05-LoCR/TheRainsOfCastamere.js
+++ b/server/game/cards/05-LoCR/TheRainsOfCastamere.js
@@ -1,5 +1,6 @@
 const AgendaCard = require('../../agendacard');
 const RevealPlots = require('../../gamesteps/revealplots');
+const SimpleStep = require('../../gamesteps/simplestep');
 
 class TheRainsOfCastamere extends AgendaCard {
     constructor(owner, cardData) {
@@ -50,10 +51,12 @@ class TheRainsOfCastamere extends AgendaCard {
         this.game.addMessage('{0} uses {1} and kneels their faction card to reveal {2}',
             context.player, this, context.target);
 
+        this.game.startRevealingPlots();
         context.player.selectedPlot = context.target;
         context.player.removeActivePlot();
         context.player.flipPlotFaceup();
         this.game.queueStep(new RevealPlots(this.game, [context.target]));
+        this.game.queueStep(new SimpleStep(this.game, () => this.game.stopRevealingPlots()));
     }
 
     onPlotDiscarded(event) {
--- a/server/game/cards/05-LoCR/TheRainsOfCastamere.js
+++ b/server/game/cards/05-LoCR/TheRainsOfCastamere.js
@@ -1,6 +1,5 @@
 const AgendaCard = require('../../agendacard');
 const RevealPlots = require('../../gamesteps/revealplots');
-const SimpleStep = require('../../gamesteps/simplestep');
 
 class TheRainsOfCastamere extends AgendaCard {
     constructor(owner, cardData) {
@@ -51,12 +50,10 @@ class TheRainsOfCastamere extends AgendaCard {
         this.game.addMessage('{0} uses {1} and kneels their faction card to reveal {2}',
             context.player, this, context.target);
 
-        this.game.startRevealingPlots();
         context.player.selectedPlot = context.target;
         context.player.removeActivePlot();
         context.player.flipPlotFaceup();
         this.game.queueStep(new RevealPlots(this.game, [context.target]));
-        this.game.queueStep(new SimpleStep(this.game, () => this.game.stopRevealingPlots()));
     }
 
     onPlotDiscarded(event) {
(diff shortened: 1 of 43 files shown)


server/game/cards/08.1-TAK/MagTheMighty.js
```diff
@@ -1,25 +1,25 @@
...
                 this.game.addMessage('{0} is forced by {1} to kill {2}', context.player, this, context.target);
 
-                //TODO Technically should only trigger when the first kill was not saved
-                this.game.promptForSelect(context.event.challenge.loser, {
-                    activePromptTitle: 'Select a character (only when first kill was not saved)',
-                    cardCondition: card => card.location === 'play area' && card.controller === context.event.challenge.loser && card.getType() === 'character',
-                    source: this,
-                    onSelect: (player, card) => this.onCardSelected(player, card)
-                });
+                if(context.target.location && context.target.location === 'dead pile') {
```
ystros, line 15: If we're going to fix Mag, a proper `then` mechanism has been introduced that should solve this issue. We should do that rather than continue this psuedo fix (which I don't think will work due to the way interrupts + events are processed). Maybe drop this from the PR and then PR separately?

server/game/gamesteps/challenge/challengeflow.js
```diff
@@ -144,6 +144,8 @@ class ChallengeFlow extends BaseStep {
 
         defendersToKneel = undefined;
 
+        this.game.raiseEvent('onDefendersDeclared', { player: this.challenge.defendingPlayer, numOfDefendingCharacters: defenders.length });
```
ystros, line 147: Is there a way we can avoid introducing this event? We already got rid of an event very similar to this.

server/game/game.js
```diff
@@ -343,9 +343,8 @@ class Game extends EventEmitter {
             case 'active plot':
                 this.callCardMenuCommand(player.activePlot, player, menuItem);
                 break;
+            //agenda and play area can behave the same now as the alliance agenda allows you to have more than one agenda card that are clickable
```
ystros, line 346: Ah, I didn't notice the new Agenda was also a banner. We might want to adjust the UI in some way, but this should be fine for now.

ystros (commented):
I haven't reviewed the new cards themselves yet, but took a look at some of the supporting engine changes (and unrelated changes to existing cards)

server/game/game.js
EvadLhorg: Created #3072 for this

server/game/cards/08.1-TAK/MagTheMighty.js
EvadLhorg: Will cut this from the PR and open up a seperate one for good old mag

server/game/cards/05-LoCR/TheRainsOfCastamere.js
```diff
@@ -50,10 +51,12 @@ class TheRainsOfCastamere extends AgendaCard {
         this.game.addMessage('{0} uses {1} and kneels their faction card to reveal {2}',
             context.player, this, context.target);
 
+        this.game.startRevealingPlots();
```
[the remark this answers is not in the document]
EvadLhorg: The changes to the plot phase were necessary (in my opinion) because currently reactions to a plot card entering the used pile during the plot phase triggers BEFORE the resolution of the When Revealed effects of the chosen plots. This is wrong from a rule perspective. You can only trigger Sand Steed/Tattered Prince´s Reaction AFTER all When Revealed effects have been resolved.
I tried to figure out how to make the whole "revealing of new plots and placing the old one in the used pile" atomic, but this didn´t solve the issue or i wasn´t trying the correct changes, so i went for a distinction between placing plots in the used pile during the revelation of plots in the plot phase and not.
I´ve tested the changes with Tatteres/Sand Steed and Rains and also with Change of Plans (which moves plots to the used pile without the revealing part) and so far i haven´t encountered any bugs/sideeffects

server/game/gamesteps/challenge/challengeflow.js
EvadLhorg: I´ve implemented this event for Left-Hand Lucas Codd. His reaction triggers to a player not declaring ANY defenders. As far as i have found out, there is an event for a character being declared as a defender (onDeclaredAsDefender) but those do not get created when there are no defenders declared.
Do you see another way to implement this kind of triggering conditon?

EvadLhorg:
throneteki-json-data now updated to the latest commit so the FH cards are included

server/game/cards/05-LoCR/TheRainsOfCastamere.js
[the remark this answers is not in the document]
ystros: Reverted these changes in favor of #3073, which has actual simultaneous reaction windows.

server/game/gamesteps/challenge/challengeflow.js
ystros: The ability is sufficiently strange, it's either this or a dedicated "declared no defenders" event. It's probably fine for the moment, lets just try not to rely on it too much.

PR #3073, opened 2020-11-30
Fix timing of discarding plot to used pile

### LINKED ISSUE #2125: Sand Steed timing
Sand Steed power should trigger after when revealed effects have been resolved. 

Currently creating some odd situations such as being able to trigger when opponent is revealing king in the north, or letting someone trigger and win a game before a valar morghulis, when actually the characters should die first.

ystros:
Fixes #2125

### DIFF
--- a/server/game/GameActions/LeavePlay.js
+++ b/server/game/GameActions/LeavePlay.js
@@ -1,0 +1,50 @@
+const GameAction = require('./GameAction');
+const ReturnCardToHand = require('./ReturnCardToHand');
+const DiscardCard = require('./DiscardCard');
+
+class LeavePlay extends GameAction {
+    constructor() {
+        super('leavePlay');
+    }
+
+    canChangeGameState({ card }) {
+        return ['active plot', 'faction', 'play area', 'title'].includes(card.location);
+    }
+
+    createEvent({ card, allowSave = false }) {
+        const params = {
+            card,
+            allowSave,
+            automaticSaveWithDupe: true
+        };
+
+        return this.event('onCardLeftPlay', params, event => {
+            event.cardStateWhenLeftPlay = event.card.createSnapshot();
+            event.card.leavesPlay();
+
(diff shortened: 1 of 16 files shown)


server/game/GameActions/LeavePlay.js
```diff
@@ -0,0 +1,50 @@
...
+
+    createEvent({ card, allowSave = false }) {
+        const params = {
+            card,
+            allowSave,
+            automaticSaveWithDupe: true
+        };
+
+        return this.event('onCardLeftPlay', params, event => {
+            event.cardStateWhenLeftPlay = event.card.createSnapshot();
```
EvadLhorg, line 22: IIRC we wanted to seperate the snapshot creation from the event handler like we did for `DiscardCard.js` so interrupts that overwrite the handler don´t overwrite the snapshot creation. I suggest adding `snapshotName: 'cardStateWhenLeftPlay'` to the params and then the snapshot will be created automatically inside `event.js`

server/game/gamesteps/plotphase.js
```diff
@@ -51,27 +49,15 @@ class PlotPhase extends Phase {
...
-        }
-    }
-
     recyclePlots() {
         for(const player of this.game.getPlayers()) {
             player.recyclePlots();
         }
     }
 
     getActivePlots() {
```
EvadLhorg, line 58: maybe rename to `getSelectedPlots()` as the plots are not active yet?

EvadLhorg (changes requested):

server/game/GameActions/LeavePlay.js
ystros: Good catch, this class was originally put together before you had added that mechanism.

server/game/gamesteps/plotphase.js
ystros: Yup. I want to totally change how plot selection is done so this method will go away eventually, but for now, the name will be updated.

PR #3075, opened 2020-12-04
Use proper "Then" on Mag the Mighty + Arianne Martell

### LINKED ISSUE #1642: Mag the Mighty and Saves
Currently it appears you are allowed to save a character using a dupe for Mag the Mighty and still force the opponent to kill one.  This appears to be wrong as the pre-then was not fully resolved

Mag the Mighty:  Forced Reaction: After you win a challenge in which Mag the Mighty is participating, choose and kill a character you control. Then, the losing opponent chooses and kills a character he or she controls.

RRG:  Save:  When a card is saved from an effect that attempts to remove that card from play, the card remains in play as the effect resolves and is not considered to have been killed, discarded, or otherwise removed from play by the effect

RRG: Then:If the pre-then aspect of an effect does not successfully resolve in full, the post-then aspect does not attempt to resolve.

http://www.cardgamedb.com/forums/index.php?/topic/37865-risen-from-the-sea-vs-mag-the-mighty/

### LINKED ISSUE #2053: Arianne Martell (Core) + reactions to entering play
When Arianne puts a character into play, any reactions to that character entering play should trigger after Arianne returns to hand (e.g. Ramsay Snow, Areo Hotah, Hotah's Axe).

ystros:
Fixes #1642 
Fixes #2053


server/game/cards/08.1-TAK/MagTheMighty.js
```diff
@@ -7,28 +8,32 @@ class MagTheMighty extends DrawCard {
...
-                    onSelect: (player, card) => this.onCardSelected(player, card)
-                });
+                this.game.resolveGameAction(
+                    GameActions.kill(context => ({
+                        card: context.target
+                    })).then(preThenContext => ({
+                        target: {
+                            cardCondition: { location: 'play area', type: 'character', controller: preThenContext.event.challenge.loser },
+                            gameAction: 'kill'
+                        },
```
EvadLhorg, line 23: The choosing of the target should be done by the losing opponent. Currently, the winner/controller of mag chooses the opponent´s character to kill. I think adding `choosingPlayer` to the `target` should work, but i am not sure if you can use `preThenContext.event.challenge.loser` as the value for `choosingPlayer`

EvadLhorg (changes requested):

server/game/cards/08.1-TAK/MagTheMighty.js
ystros: Good catch, thanks!

PR #3080, opened 2020-12-05
fix dreadfort maester so that a player other than the owner also has their claim raised

EvadLhorg:


server/game/cards/08.1-TAK/DreadfortMaester.js
```diff
@@ -9,6 +9,7 @@ class DreadfortMaester extends DrawCard {
             cost: ability.costs.sacrificeSelf(),
             handler: context => {
                 this.untilEndOfChallenge(ability => ({
+                    targetController: 'any', //here any is necessary in case dreadfort maester changes it´s controller to a player other than the owner
```
ystros, line 12: This looked incorrect (since it should only work with the triggering player) but is actually correct, since the `match` function already narrows it down to the right player. Good find!

PR #3081, opened 2020-12-07
Update data to contain all FFG sample decks

ystros:


server/scripts/exportstandalonedeck.js
```diff
@@ -0,0 +1,36 @@
+/*eslint no-console:0 */
+
+const monk = require('monk');
+
+const CardService = require('../services/CardService');
+const DeckService = require('../services/DeckService');
+
+class ExportStandaloneDecks {
+    constructor() {
+        this.db = monk('mongodb://127.0.0.1:27017/throneteki');
```
EvadLhorg, line 10: does this have to be changed depending on the dev´s setup or should we use the property dbPath from the local/default.json5 config file here?

ystros: Going to leave it for the moment, as the same thing exists in `ImportStandaloneDecks` (which is run / is meant for even production)

PR #3088, opened 2020-12-20
fix Hotho Humpback 

### LINKED ISSUE #2878: Players are able to keep 1 more card than reserve after Hotho Humpback trigger
Players consistently keep 1 more card than required to after a Hotho humpback trigger.  For example, if their reserve is 3, they'll be able to keep 4 after Hotho triggers.  I've only noticed this from opponents, so not 100% sure if it happens when you yourself are over reserve.

EvadLhorg:
so the draw happens first, then the checking for reserve

resolves #2878

### DIFF
--- a/server/game/cards/15-DotE/HothoHumpback.js
+++ b/server/game/cards/15-DotE/HothoHumpback.js
@@ -13,12 +13,14 @@ class HothoHumpback extends DrawCard {
                         ...this.game.getPlayersInFirstPlayerOrder().map(player =>
                             GameActions.drawCards({ player, amount: 1 })
                         )
-                    ]).then({
-                        gameAction: GameActions.checkReserve()
-                    }),
+                    ]),
                     context
                 );
-            },
+                this.game.resolveGameAction(
+                    GameActions.checkReserve(),
+                    context
+                );
+            },            
             limit: ability.limit.perRound(1)
         });
     }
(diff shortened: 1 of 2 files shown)


server/game/cards/15-DotE/HothoHumpback.js
```diff
@@ -7,12 +7,18 @@ class HothoHumpback extends DrawCard {
...
-                ),
-                GameActions.checkReserve()
-            ]),
+            handler: context => {
+                this.game.resolveGameAction(
+                    GameActions.simultaneously(() => [
+                        ...this.game.getPlayersInFirstPlayerOrder().map(player =>
+                            GameActions.drawCards({ player, amount: 1 })
+                        )
+                    ]).then({
```
ystros, line 16: This *might* be worth merging, but it is not correct (otherwise I would have implemented this way in the first place). 

Consider if either played has a restriction on the number of cards they can draw (e.g. High Sparrow), and cannot draw the card for Hotho. Since the pre-then condition was unsuccessful - either one or both players did not draw a card - then the post-then check reserve will not resolve. But Hotho's card is phrased in a way such that reserve should always be checked, regardless of the success of the draw.

EvadLhorg: Thanks for your review, you´ve made a really good point i didn´t think about. I made according changes in the latest commit and added a test to check for the edge case of a draw restriction in play.

ystros: Thanks! Theoretically the original code should have worked, but pipeline ordering is hard to understand. Adding the test is a big improvement, and will let us refactor this to something better later.

PR #3089, opened 2020-12-28
Only allow play area for Winterfell Heart Tree

### LINKED ISSUE #3036: Winterfell Heart Tree targets every Stark card
Should only target those that are in play

ystros:
Fixes #3036


EvadLhorg:
lgtm

PR #3090, opened 2020-12-28
Fix the Knight gaining stealth

### LINKED ISSUE #3059: Bug: The Knight doesn´t have Stealth when attacking alone
I think this is a follow up bug regarding the refactoring of #3056 and #3041 
the condition of the knight to gain Stealth is `this.game.isDuringChallenge({attackingPlayer: this.controller, attackingAlone: this })` but i think the `attackingAlone` part is true when the `chooseStealthTargets` step is being processed

ystros:
A previous fix aimed at Randyl Tarly and Lord of the Crossing
changed when a character was considered attacking. This broke
The Knight's ability to gain stealth before the challenge was
officially initiated.

The attacking / defending check has been reverted, and instead
Lord of the Crossing has been modified to only apply the
strength buff after the challenge has been initiated.

Fixes #3059

### DIFF
--- a/server/game/ChallengeMatcher.js
+++ b/server/game/ChallengeMatcher.js
@@ -6,6 +6,7 @@ class ChallengeMatcher {
             Matcher.containsValue(matchers.challengeType, challenge.challengeType) &&
             Matcher.containsValue(matchers.attackingPlayer, challenge.attackingPlayer) &&
             Matcher.containsValue(matchers.defendingPlayer, challenge.defendingPlayer) &&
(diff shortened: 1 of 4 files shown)


EvadLhorg:
lgtm! checked out the branch to see if the stealth prompt shows up, that might be an improvement to the test. testing for having the stealth keyword would also have worked before the changes you made i think, as the prompting for stealth targets happened to early.

ystros:
Good feedback! Added an explicit test about the stealth prompt.

PR #3092, opened 2021-01-15
Count marshalling dupes and shadows for Marwyn

### LINKED ISSUE #2951: Cards marshalled into shadows through Archmaester Marwyn are not tracked
This is a follow up issue to my fix of #2932 that I realized now that the fix is deployed.

Marshalling a card into shadows is not tracked correctly and therefore does not count towards Marwyn's once per phase limit. Also, the user message does not state that the card was marshalled from under the agenda.

ystros:
Fixes #2951


EvadLhorg:
LGTM

PR #3093, opened 2021-01-15
Fix Castle Guard target condition

### LINKED ISSUE #3000: Castle Guard does not return cards to Shadows that were "Put into play from shadows" via a card ability
Example, challenge phase begins. I put the Red Keep Gaoler into play from shadows via it's card ability (move 2 power from your faction card to any baratheon character you control). This triggers cards like Black Cells and Blackwater Rush for "coming out of shadows." However, later in the challenge phase, I pay to bring the Castle Guard out of shadows. Castle Guard reads "After Castle Guard comes out of shadows, choose another character that came out of shadows this phase. Return that character to shadows." The Reaction should be triggered with the option of returning the Red Keep Gaoler to shadows, but it currently isn't firing.

ystros:
Check for the onCardOutOfShadows event instead of whether a card
was put into play with the 'outOfShadows' meta-action.

Fixes #3000

### DIFF
--- a/server/game/EventTrackers/GenericTracker.js
+++ b/server/game/EventTrackers/GenericTracker.js
@@ -10,6 +10,10 @@ class GenericTracker {
         game.on(endingEvent, () => this.clearEvents());
     }
 
+    some(predicate) {
+        return this.events.some(predicate);
+    }
+
     trackEvent(event) {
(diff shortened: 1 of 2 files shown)


EvadLhorg:
LGTM

PR #3094, opened 2021-01-15
Ensure Karhold persistent effect gets recalculated

### LINKED ISSUE #2783: Karhold didn't work
I was playing a game with Karhold on the board on my side, I had A time for wolves and the opponent was playing Greyjoy and had revealed You win or You die.

Despite nor having a winter plot he still was able to claim power with Euron and Corpse Lake (no nightmares, seized or other negative attachments were in play).

ystros:
The cards that Karhold is applied to needs to change as new plots are
revealed. Adding an explicit condition to the effect ensures the effect
is continuously recalculated.

Fixes #2783


EvadLhorg:
LGTM

PR #3095, opened 2021-01-15
Automatically complete prompts for fake players

### LINKED ISSUE #3027: "Waiting for opponent" blockage when sparring alone
When I test decks playing alone, despite of not having an opponent so many card reactions or others ask for your opponent to make a move and as there is no one I need to cancel-prompt every single time.
This happens for instance with the forced reaction from Late Summer Feast or if you play an event like Hidden thorns or for instance Gunthor son of Gurn's reaction to discard from hand. There are lots of "waiting for opponent" blocks and considering this it'd be good whether this never happens and all jumps to the conclusion
Screenshot 2020-07-22 at 12 38 30

ystros:
In single player games, we create a fake defending player during challenges
to avoid special casing every ability to handle a missing defending player.
This leads to a problem on abilities that prompt the defending player, such
as Late Summer Feast, since there's no player to respond to the prompt.

Now the menu and card select prompts will immediately complete themselves
if the player being prompted is the fake player. This may not cover every
case, but it should fix many of the existing hanging prompts.

Fixes #3027

### DIFF
--- a/server/game/challenge.js
+++ b/server/game/challenge.js
@@ -30,6 +30,7 @@ class Challenge {
         let dummyPlayer = new Player('', Settings.getUserWithDefaultsSet({ name: 'Dummy Player' }), false, this.game);
         dummyPlayer.initialise();
         dummyPlayer.resetForStartOfRound();
+        dummyPlayer.isFake = true;
         return dummyPlayer;
     }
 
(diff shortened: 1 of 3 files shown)


EvadLhorg:
LGTM
I will get in touch with the creator of the issue and ask him to keep us posted if he finds any more instances where the fake player not completing a prompt gets in the way.

PR #3104, opened 2021-02-21
fix city of wealth so the plot modifiers really works

EvadLhorg:


server/game/cards/13.6-LMHR/CityOfWealth.js
```diff
@@ -13,7 +13,7 @@ class CityOfWealth extends PlotCard {
     }
 
     numOfCityPlots() {
-        this.controller.getNumberOfUsedPlotsByTrait('City');
+        return this.controller.getNumberOfUsedPlotsByTrait('City');
```
ystros, line 16: Oops

PR #3115, opened 2021-04-20
Fix Citadel Archivist/Heads on Spikes interaction

ystros:


EvadLhorg:
LGTM, tested it locally and everything seems to work like intended

Issue #3062, opened 2020-11-01
startedAt property of a game is saved in differenty types


EvadLhorg:
Sometimes, a game saves the startedAt property as a Date, sometimes it saves it as a string. This makes executing queries against this document quite hard as there is no coherent data type for this property. We should use only one data type for this property and use it everywhere the startedAt property is written.

EvadLhorg:
this can be closed now as #3124 is now deployed on the server @ystros

PR #3124, opened 2021-05-16
 save the startedAt and finishedAt properties as a Date when saving to the database

EvadLhorg:
save the startedAt and finishedAt properties as a Date when saving to the database

cast the input parameters of getAllGames to a date so the comparison works

update stats.js so the local db is used as config.js got removed some time ago

add number of unique players to the stats result

### DIFF
--- a/server/stats.js
+++ b/server/stats.js
@@ -4,8 +4,9 @@ const _ = require('underscore');
 const monk = require('monk');
 
 const GameService = require('./services/GameService.js');
+const config = require('./config.js');
 
-let db = monk('mongodb://127.0.0.1:27017/throneteki');
+let db = monk(config.dbPath);
 let gameService = new GameService(db);
 
 let args = process.argv.slice(2);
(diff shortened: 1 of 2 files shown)


server/stats.js
```diff
@@ -4,9 +4,8 @@ const _ = require('underscore');
 const monk = require('monk');
 
 const GameService = require('./services/GameService.js');
-const config = require('./config.js');
 
-let db = monk(config.dbPath);
+let db = monk('mongodb://127.0.0.1:27017/throneteki');
```
ystros, line 8: Changes in this file seem unrelated (perhaps stuff you've done for local dev?)

ystros (changes requested):

server/stats.js
EvadLhorg: Kind of unrelated, i guess a seperate PR is a good idea. I would like an addition to the stats script to get the number of unique players (see line 139 to 142) and this change is necessary as the config.js file got removed some time ago

EvadLhorg:
fixes #3062

server/stats.js
EvadLhorg: Will revert these changes and open up a seperate PR with the usage of configservice, thanks!

PR #3125, opened 2021-05-16
improve SacrificeCard game action so the card gets placed via another…

EvadLhorg:
… game action instead of moved to the discard pile

### DIFF
--- a/test/server/GameActions/SacrificeCard.spec.js
+++ b/test/server/GameActions/SacrificeCard.spec.js
@@ -62,7 +62,10 @@ describe('SacrificeCard', function() {
             });
 
             it('moves the card to discard', function() {
-                expect(this.playerSpy.moveCard).toHaveBeenCalledWith(this.cardSpy, 'discard pile');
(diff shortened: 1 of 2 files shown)


test/server/GameActions/SacrificeCard.spec.js
```diff
@@ -62,7 +62,10 @@ describe('SacrificeCard', function() {
             });
 
             it('moves the card to discard', function() {
-                expect(this.playerSpy.moveCard).toHaveBeenCalledWith(this.cardSpy, 'discard pile');
```
ystros, line 65: Should remove `moveCard` method from the spy object at the top.

ystros (commented):
This is the exact direction I want to go for all card-moving events. I forget why I hadn't done it on SacrificeCard, but remember running into problems with other events (probably Kill)

test/server/GameActions/SacrificeCard.spec.js
EvadLhorg: good catch, fixed it and removed the playerSpy in the latest commit

PR #3128, opened 2021-05-26
placeCardInPile should invoke card.leavesPlay 

EvadLhorg:
placeCardInPile should invoke card.leavesPlay so that the reworked SacrificeCard->PlaceCard interaction leads to cards leaving play correctly
This was noticed on the playtesting server when a Kingsroad got sacrificed for its reduction but entered the discard pile knelt. Turns out, the "standing" of a card entering a non-play area is done in card.leavesPlay


ystros:
Closing in favor of #3130

PR #3130, opened 2021-05-31
Make sacrifice and leave play simultaneous

ystros:
Replaces #3128 

Also switches to using the `snapshotName` event mechanism to record card state.


EvadLhorg:
LGTM, this fixes the problem i encountered with the previous change that attachments get discarded instead of returned to hand.

PR #3134, opened 2021-06-23
hides the source of an MarshalIntoShadowsAction in the alert 

EvadLhorg:
hides the source of an MarshalIntoShadowsAction in the alert when doing the action outside of your action window

when marshaling a card into shadows during your opponents reaction window by accident, the marshaling action is outside of your action window and it puts an alert in the game log, stating which card you marshal into shadows and thus spoiling secret information to your opponent


ystros:
👍

PR #3135, opened 2021-06-23
Fixes from playtesting

EvadLhorg:
The fixes include:

- fixes for atTheEndOfPhase effects not expiring when the affected card left and enters play again
- fix for green dreams crash when there is no card left in the deck
- fix for weapons at the door to not target facedown cards
- fix for patchface to correctly use the keywords of other fools
- allow donots to trigger on opponent´s lady characters
- fix for left hand lucas codd

### DIFF
--- a/server/game/cards/03-WotN/GreenDreams.js
+++ b/server/game/cards/03-WotN/GreenDreams.js
@@ -4,24 +4,22 @@ class GreenDreams extends DrawCard {
     setupCardAbilities() {
         this.reaction({
             when: {
-                onCardKneeled: event => event.card === this.parent
+                onCardKneeled: event => event.card === this.parent && this.controller.drawDeck[0]
             },
             handler: () => {
                 this.topCard = this.controller.drawDeck[0];
                 this.game.addMessage('{0} uses {1} to look at the top card of their deck', this.controller, this);
 
-                if(this.topCard) {
-                    this.game.promptWithMenu(this.controller, this, {
-                        activePrompt: {
-                            menuTitle: 'Put ' + this.topCard.name + ' on bottom of your deck?',
-                            buttons: [
-                                { text: 'Yes', method: 'placeOnBottom', arg: 'yes', card: this.topCard },
-                                { text: 'No', method: 'placeOnBottom', arg: 'no', card: this.topCard }
-                            ]
-                        },
-                        source: this
-                    });
-                }
+                this.game.promptWithMenu(this.controller, this, {
+                    activePrompt: {
+                        menuTitle: 'Put ' + this.topCard.name + ' on bottom of your deck?',
+                        buttons: [
+                            { text: 'Yes', method: 'placeOnBottom', arg: 'yes', card: this.topCard },
+                            { text: 'No', method: 'placeOnBottom', arg: 'no', card: this.topCard }
+                        ]
+                    },
+                    source: this
+                });
             }
         });
     }
--- a/server/game/cards/03-WotN/GreenDreams.js
+++ b/server/game/cards/03-WotN/GreenDreams.js
@@ -10,16 +10,18 @@ class GreenDreams extends DrawCard {
                 this.topCard = this.controller.drawDeck[0];
                 this.game.addMessage('{0} uses {1} to look at the top card of their deck', this.controller, this);
 
-                this.game.promptWithMenu(this.controller, this, {
-                    activePrompt: {
-                        menuTitle: 'Put ' + this.topCard.name + ' on bottom of your deck?',
-                        buttons: [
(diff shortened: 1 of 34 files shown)


server/game/cards/03-WotN/GreenDreams.js
```diff
@@ -10,16 +10,18 @@ class GreenDreams extends DrawCard {
...
-                    activePrompt: {
-                        menuTitle: 'Put ' + this.topCard.name + ' on bottom of your deck?',
-                        buttons: [
-                            { text: 'Yes', method: 'placeOnBottom', arg: 'yes', card: this.topCard },
-                            { text: 'No', method: 'placeOnBottom', arg: 'no', card: this.topCard }
-                        ]
-                    },
-                    source: this
-                });
+                if(this.topCard) {
```
ystros, line 13: 1. If there isn't a top card, shouldn't we not allow the ability to trigger at all? The better fix would be to check it in the triggering condition.
2. If there is no cards in the draw deck, you lose immediately.

server/game/cards/06.6-TBWB/Patchface.js
```diff
@@ -16,17 +15,8 @@ class Patchface extends DrawCard {
...
-
-        for(let card of fools) {
-            for(let keyword of Keywords) {
-                if(card.hasKeyword(keyword)) {
-                    foolKeywords.push(keyword);
-                }
-            }
-        }
-
+        let foolKeywords = [...new Set(flatten(fools.map(card => card.getKeywords())))];
```
ystros, line 19: Is this to ensure Patchface can gain Bestow? Does this actually work as he enters play?

server/game/cards/09-HoT/SerDontosHollard.js
```diff
@@ -17,8 +17,8 @@ class SerDontosHollard extends DrawCard {
         });
     }
 
-    isControlledLady(card) {
-        return card.location === 'play area' && card.controller === this.controller && card.hasTrait('Lady');
+    isLady(card) {
+        return card.location === 'play area' && card.hasTrait('Lady');
```
ystros, line 21: I have mixed feelings about allowing this to trigger on opponent characters. But you do get a benefit from it (standing Dontos). We should watch for feedback about whether folks find it annoying / confusing.

server/game/cards/18-FH/LeftHandLucasCodd.js
```diff
@@ -5,7 +5,7 @@ class LeftHandLucasCodd extends DrawCard {
     setupCardAbilities() {
         this.reaction({
             when: {
-                onDefendersDeclared: event => event.player !== this.controller && event.numOfDefendingCharacters === 0 && this.hasAttackingRaider()
+                onDefendersDeclared: event => event.player !== this.controller && event.numOfDefendingCharacters === 0 && this.hasAttackingRaider(this.controller)
```
ystros, line 8: This looks equivalent to what was there before. `this.controller` is being passed in instead of hardcoded in the method. What am I missing?

ystros (commented):
Good catch on those atEndOfPhases. Not sure how no one has noticed up until now. 😅 

I'm not fond of the `condition: () => ['play area', 'duplicate'].includes(context.target.location),` pattern, but don't have a great alternative yet.

server/game/cards/18-FH/LeftHandLucasCodd.js
EvadLhorg: You are missing nothing, i had a look at the commit history of LeftHandLucasCodd in the playtesting repo to figure out why this diff exists:
there is another card with the same effect coming up in the next pack and it introduced a bug to LeftHandLucasCodd, so this is what the correction looks like.
I will change it back and also update the new card which should work with the same solution.

server/game/cards/09-HoT/SerDontosHollard.js
EvadLhorg: I also suggest not doing this everywhere and freely, but i had a request from the design discord community about this interaction so i figured i could change it.

server/game/cards/06.6-TBWB/Patchface.js
EvadLhorg: the patchface change enables him to gain the correct versions of "No attachment except X" keywords, bestow wasn´t on my radar, but it should also work in the sense that he gains it (but no bestwo trigger is prompted as he doesn´t enter play).

server/game/cards/03-WotN/GreenDreams.js
EvadLhorg: good catch, now that you mention the lose condition, i wonder how this happened in the first place. I remember seeing this in a stacktrace in the server log as the reason for the gamenode crashing.
Probably pulling it up to the triggering condition is a better idea

PR #3137, opened 2021-07-01
implement pause functionality for general timelimit

EvadLhorg:
also take a look at the client PR here https://github.com/throneteki/throneteki-client/pull/139

### DIFF
--- a/server/game/timeLimit.js
+++ b/server/game/timeLimit.js
@@ -40,11 +40,7 @@ class TimeLimit {
             this.timeLimitInSeconds -= Math.floor(differenceBetweenStartOfTimerAndNow.asSeconds());
         } else {
         //unpause
-            this.timeLimitStarted = true;
-            this.timeLimitStartedAt = new Date();
-            this.timer = setInterval(() => {
-                this.checkForTimeLimitReached();
-            }, 1000);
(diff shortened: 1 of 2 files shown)


server/game/timeLimit.js
```diff
@@ -30,10 +30,28 @@ class TimeLimit {
...
+            this.timer = undefined;
+            let differenceBetweenStartOfTimerAndNow = moment.duration(moment().diff(this.timeLimitStartedAt));
+            this.timeLimitInSeconds -= Math.floor(differenceBetweenStartOfTimerAndNow.asSeconds());
+        } else {
+        //unpause
+            this.timeLimitStarted = true;
+            this.timeLimitStartedAt = new Date();
+            this.timer = setInterval(() => {
+                this.checkForTimeLimitReached();
+            }, 1000);
```
ystros, line 47: For this else, any reason we can't call `startTimer` instead? It seems to be repeating the logic exactly.

ystros (commented):
Other than slight duplication, looks good!

server/game/timeLimit.js
EvadLhorg: good catch, totally missed that. fixed!

PR #3140, opened 2021-07-02
only prompt for sacrifice on heir to the iron throne if the card put into play was not a duplicate

### LINKED ISSUE #2417: Heir to the Iron Throne and dupes
When using Heir to the Iron Throne, if you choose a duplicate you are still being offered a sacrifice window, which can lead to confusion as you have to manually cancel out of it.   Duplicates or failing to find a lord or lady should mean the Then effect does not trigger.  I have not seen if it offers you the sacrifice opportunity if you dont find a character, only on duplicates.

**relevant rules**

If you do not find a Lord or Lady in the top 10 cards, the pre-then aspect of the effect did not resolve successfully and thus you cannot sacrifice a Lord or Lady. You can choose not to find a Lord or Lady despite there being a Lord or Lady in the top 10 cards.

If you find and put into play a duplicate of a Lord or Lady, the pre-then aspect of the effect did not resolve successfully and thus you cannot sacrifice a Lord or Lady.

EvadLhorg:
new players get confused by this rule. you have to know about it and then cancel the sacrifice manually which in turn leads to a warning in the game log.

fixes #2417

### DIFF
--- a/server/game/cards/12-KotI/HeirToTheIronThrone.js
+++ b/server/game/cards/12-KotI/HeirToTheIronThrone.js
@@ -25,8 +25,6 @@ class HeirToTheIronThrone extends PlotCard {
             this.game.queueSimpleStep(() => {
                 this.promptForSacrifice(player);
             });
-        } else {
-            this.game.addMessage('{0} can not sacrifice a character because {1} is put into play as a duplicate', player, card);
         }
     }
 
--- a/server/game/cards/12-KotI/HeirToTheIronThrone.js
+++ b/server/game/cards/12-KotI/HeirToTheIronThrone.js
@@ -18,11 +18,16 @@ class HeirToTheIronThrone extends PlotCard {
 
     cardSelected(player, card) {
         this.game.addMessage('{0} uses {1} to search their deck and put {2} into play', player, this, card);
+        let dupeCard = player.getDuplicateInPlay(card);
         player.putIntoPlay(card);
-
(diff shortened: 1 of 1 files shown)


server/game/cards/12-KotI/HeirToTheIronThrone.js
```diff
@@ -18,11 +18,16 @@ class HeirToTheIronThrone extends PlotCard {
...
-        this.game.queueSimpleStep(() => {
-            this.promptForSacrifice(player);
-        });
+        //only prompt for sacrifice if the card put into play wasn´t a duplicate
+        if(!dupeCard) {
+            this.game.queueSimpleStep(() => {
+                this.promptForSacrifice(player);
+            });
+        } else {
+            this.game.addMessage('{0} can not sacrifice a character because {1} is put into play as a duplicate', player, card);
```
ystros, line 29: I don't think we should explain rules via game messages. If we add this, we'd also have to go back to all `Then` abilities and add similar things (e.g. you save with a dupe pre-Then for Mag the Mighty, etc).

ystros (changes requested):

server/game/cards/12-KotI/HeirToTheIronThrone.js
EvadLhorg: good point, it is explained on the corresponding thronesdb page and i guess there rule explanations belong.
fixed

Issue #3142, opened 2021-07-05
Hear My Words (HMW) known Issues


EvadLhorg:
Here is a list of known issues regarding cards from the new pack Hear My Words (with a possible workaround for each issue)

- A Mummer´s Farce (https://thronesdb.com/card/20051)
  - Cost Reduction is not working correctly with facedown cards being played from under Fools. As the cards are of the card type "attachment", only "attachment" (and of course any type) cost reductions will be applied.
  - Please apply reductions manually by adding/substracting gold and double check your spent gold when playing/marshaling a facedown card from under a Fool.

- Septa Lemore (https://thronesdb.com/card/20032)
  - Septa Lemore leaving play simultaneously with another card you want to target with her effect might not work as intended. For example, Septa Lemore and her target both dying on a Valar Morghulis reveal will still place both cards in the dead pile.
  - Please use the /move-shadows chat command AFTER the resolution of the effect that makes septa and her target leave play to move the target of Septa Lemores effect to shadows.

- Brave Companions (https://thronesdb.com/card/20042)
  - When triggering the reaction of Brave Companions, selecting a card with multiple matching traits (for example a character with both the Mercenary and Commander traits), the remaining cards to be selected are not correctly highlighted/enabled to be selected.
  - Please change the order of the selected cards, starting with the card with the least traits that you want to pick. If you can´t select all the cards you want with this method, resolve the effect manually by dragging/dropping the cards to your hand and tell your opponent which cards you have chosen if it is not stated already in the game log.
 
- Many-Faced God (https://thronesdb.com/card/20052)
  - When you have Many-Faced God as your agenda and have a Driftwood Cudgel (https://thronesdb.com/card/08112) in play, it counts the number of characters in your dead pile twice, which is wrong.
  - Please try to calculate the strength correctly by hand or by using the /strength x chat command to set a character's strength. 

- Sweetsleep (https://thronesdb.com/card/20047)
  - When attaching an attachment with a passive effect (for example +2 STR of Widow´s Wail) on a character that ALREADY has Sweetsleep attached to it, the passive effect is STILL applied, which is wrong.
  - Please re-attach Sweetsleep to the character (by dragging Sweetsleep to your hand and then onto the play area again, targeting the same character) to blank the new attachment.

Patane97:
Brave Companions has been fixed in #3144

PR #3144, opened 2021-07-08
Updated BraveCompanions & FreshRecruits logic

Patane97:
Fixes a bug with Brave Companions (and Fresh Recruits) by updating its logic. It will now correctly highlight the valid cards based on the remaining _**Traits**_ when looking in the deck, even if a card with multiple of those _**Traits**_ is selected. For example, if you find a character with _**Commander. Mercenary.**_, they could be selected as the _**Commander**_ card or the _**Mercenary**_ card, meaning both traits are still available for selection.
This update uses permutation to find all valid combinations of the selected cards against the given traits, and returns the list of remaining traits that can still be selected.
Discord: Deathlysteve#4587 for any queries.

### DIFF
--- a/server/game/cards/11.1-TSC/FreshRecruits.js
+++ b/server/game/cards/11.1-TSC/FreshRecruits.js
@@ -1,8 +1,10 @@
 const DrawCard = require('../../drawcard');
 const GameActions = require('../../GameActions');
+const Array = require('../../../Array');
 
 class FreshRecruits extends DrawCard {
     setupCardAbilities() {
+        const selectableTraits = ['Ranger', 'Builder', 'Steward'];
         this.action({
             title: 'Search deck',
             gameAction: GameActions.search({
@@ -11,7 +13,10 @@ class FreshRecruits extends DrawCard {
                 numToSelect: 3,
                 match: {
                     type: 'character',
-                    condition: (card, context) => this.remainingTraits(context.selectedCards).some(trait => card.hasTrait(trait))
+                    // Checking if the card is already selected || if it has one of the selectable traits && that trait is remaining for selection
+                    condition: (card, context) => context.selectedCards.includes(card) 
+                    || selectableTraits.some(trait => card.hasTrait(trait)) 
+                    && Array.availableToPair(selectableTraits, context.selectedCards, (trait, card) => card.hasTrait(trait)).some(trait => card.hasTrait(trait))
                 },
                 message: '{player} uses {source} to search their deck and add {searchTarget} to their hand',
                 cancelMessage: '{player} uses {source} to search their deck but does not find a card',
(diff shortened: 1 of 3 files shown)


server/game/cards/11.1-TSC/FreshRecruits.js
```diff
@@ -1,25 +1,56 @@
...
+                var next = current.slice();
+                next.push(r);
+                var newRemaining = remaining.slice();
+                newRemaining.splice(index, 1);
+                this.permutate(generated, next, newRemaining);
+            });
+        } else {
+            generated.push(current);
+        }
+        return generated;
```
ystros, line 53: This logic is pretty hard to follow and is duplicated between both cards.

1. Is there something simpler that would have the same effect?
2. Can we improve the naming so that it is easier to follow? Particularly the numerous one-letter variables.
3. Can would put this logic in a common spot that both cards can use?
4. Some tests might be useful.

ystros (changes requested):

server/game/cards/20-HMW/BraveCompanions.js
```diff
@@ -40,7 +40,46 @@ class BraveCompanions extends DrawCard {
...
+        for(let card of cards) {
+            player.moveCard(card, 'hand');
+        }
+        this.game.addMessage('{0} uses {1} to search their deck and adds {2} to their hand', player, this, cards);
+        return true;
+    }
+
+    cancelSearch(player) {
+        this.game.addMessage('{0} uses {1} to search their deck, but does not add any cards to their hand', player, this);
+        return true;
```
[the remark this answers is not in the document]
Patane97: Yep, these were pasted from the play-testing repository (where it wasnt using GameActions). Removed+resolved.

server/game/cards/11.1-TSC/FreshRecruits.js
```diff
@@ -1,25 +1,56 @@
...
     setupCardAbilities() {
         this.action({
             title: 'Search deck',
-            handler: () => {
-                this.game.promptForDeckSearch(this.controller, {
-                    activePromptTitle: 'Select cards',
-                    numToSelect: 3,
-                    cardCondition: (card, context) => card.getType() === 'character' && this.remainingTraits(context.selectedCards).some(trait => card.hasTrait(trait)),
-                    onSelect: (player, cards) => this.selectCards(player, cards),
-                    onCancel: player => this.cancelSearch(player),
```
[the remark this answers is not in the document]
Patane97: Yep! Removed+Resolved.

server/game/cards/11.1-TSC/FreshRecruits.js
```diff
@@ -1,25 +1,56 @@
...
     }
 
     remainingTraits(selectedCards) {
         const traits = ['Ranger', 'Builder', 'Steward'];
-        return traits.filter(trait => !selectedCards.some(card => card.hasTrait(trait)));
+        let tempCards = selectedCards.slice();
+        for(let i = tempCards.length ; i < traits.length ; i++) {
+            tempCards.push(null);
+        }
+        var remaining = [];
```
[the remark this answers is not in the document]
Patane97: Changed from `var` to `let` on both FreshRecruits + BraveCompanions.

server/game/cards/11.1-TSC/FreshRecruits.js
Patane97: 1. I was working on this for a few days straight looking for a simpler solution (outside of selecting all 3 traits in a separate prompt). Not saying there isn't a simpler solution out there, just not sure I was able to find it. :)
2. Happy to update the 1 letter variables for clarity.
3. I was actually looking for an appropriate place to put this but couldnt find one quickly. I have designed it to work for any combination of traits (eg, if there is a future card that searches for a _**Companion**_, a _**Lord**_, a _**Lady**_ and a _**Bastard**_ character, this could do it by just adjusting the Traits constant. If you could point me to the right class to put this method in, id happily move/call it from there :)
4. I'll look into how the tests work as I am unfamiliar with it at this time. Any suggestions on how to proceed with tests?

EvadLhorg:
relates to #3142

server/game/cards/11.1-TSC/FreshRecruits.js
ystros: 1. I haven't been able to able to devote a lot of time, but I also was trying to find a simpler solution and couldn't quite get there. I think comments about the various stages might be useful, as the code is somewhat dense (even if naming is updated).
3. I'm not sure there's an existing spot that would fit well. I wouldn't want to add the method to the Game, Player, or Card classes. I've toyed with the idea of introducing a `CardConditions` class/file that would include helpers for common things (e.g. "all controlled cards are the same faction", "there's a plot revealed with a specific trait", etc). Perhaps create that as a new file and add the method that gives you the remaining traits?
4. I'd say either write tests directly for your new method (so you'd add it to a file with the same name but with `.spec.js` at the end instead) that just test various input-output combinations, or you could add an integration test that starts a game, activates either Brave Companions or Fresh Recruits, and tries selecting cards. Example integration test: https://github.com/throneteki/throneteki/blob/master/test/server/cards/15-DotE/ArchmaesterMarwyn.spec.js. If you can't figure out, don't worry about it too much, I can probably create one for you if needed.

Patane97: I believe I have resolved these issues (tried to leave as much commenting as possible to explain whats going on). I have not attempted card testing for this yet, but will work on that next. Up to you, but this should be ready to push if youre happy! :)

ystros (approved):

PR #3151, opened 2021-07-13
do not log the target and name of an attachment when the attachment is attached facedown

EvadLhorg:
side effect of https://github.com/throneteki/throneteki/commit/97c2e4bb94ee6c4490e2dc131a4c6e0414913ec9
image0

### DIFF
--- a/server/game/Message.js
+++ b/server/game/Message.js
@@ -43,6 +43,9 @@ class Message {
         if(Array.isArray(arg)) {
             return this.formatArray(arg);
         } else if(arg instanceof BaseCard) {
+            if(arg.facedown) {
+                return 'a facedown card';
+            }
             return { code: arg.code, label: arg.name, type: arg.getType(), argType: 'card' };
         } else if(arg instanceof Spectator) {
             return { name: arg.user.username, argType: 'nonAvatarPlayer' };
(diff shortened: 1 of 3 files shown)


ystros:
I'm wondering if it would be better to handle this when messages are formatted. This fixes it for the new attachment message. Consider abilities that can target the facedown attachments though. We'd have to special case the message in each ability.

Instead, we could check the facedown property when the message is being prepared for the client, and render something else instead if it's facedown. I believe this is the relevant piece:
https://github.com/throneteki/throneteki/blob/3c0d069060ec1d1ce4003ff45c668249d4543ee3/server/game/Message.js#L46

Thoughts?

EvadLhorg:
That is a good idea to fix all of these issues at once. I´ve tested this locally with a quick version like this:
`if(arg.facedown) {
    return 'a facedown card';
}
return { code: arg.code, label: arg.name, type: arg.getType(), argType: 'card' };`

The problem with this i noticed is that the cards a player sets up are also now formatted as they are facedown when the message is formatted. To fix this i think we would have to change the order of the setupphase as the message is written first and then the cards are revealed. Changing the order of that without adjustments doesn´t work unfortunately because of duplicates. Any ideas to fix this?

ystros:
I think the problem is that the message announcing setup cards happens before they're flipped face up by the `revealSetupCards` method. Could either explicitly set facedown = false on the cards just prior to the message, or try re-ordering things (though `revealSetupCards` will place dupes, which makes announcing the setup cards trickier).

https://github.com/throneteki/throneteki/blob/5d10d59af58b4c305adc9840f6a0fc2611c49f3d/server/game/gamesteps/setupphase.js#L85-L98

PR #3152, opened 2021-07-13
fix the smith game message when the attachment knelt for the cost is …

EvadLhorg:
…a facedown attachment so the game message doesn´t spoil the name of the facedown attachment


ystros:
See comment on the other "facedown attachment" issue.

ystros:
Should now be fixed with #3151

PR #3155, opened 2021-07-15
Fix Kindly Man discarding tokens

### LINKED ISSUE #3154: BUG: kindly man does not remove betayal token off of jorah even though game log says so
_what's going on?_

kindly man does not remove betrayal token off of jorah, even though it get's logged in the in-game chat as such.
the token stays on jorah.

see screenshot for example:

jorahkindlyman

(never mind the  background, i'm running thronesdb proper with custom browser extension)

ystros:
Introduces a new game action explicitly about discarding tokens from
cards. The existing place card action explicitly does not allow to
"place" negative tokens.

Fixes #3154

### DIFF
--- a/server/game/GameActions/DiscardToken.js
+++ b/server/game/GameActions/DiscardToken.js
@@ -1,0 +1,20 @@
+const GameAction = require('./GameAction');
+
+class DiscardToken extends GameAction {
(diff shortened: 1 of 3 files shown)


EvadLhorg:
lgtm

PR #3156, opened 2021-07-15
fix elias lance so the game doesn´t crash anymore because no handler is detected

EvadLhorg:


server/game/cards/20-HMW/EliasLance.js
```diff
@@ -21,8 +21,11 @@ class EliasLance extends DrawCard {
                     card.getType() === 'character' &&
                     card.isParticipating()
             },
-            message: '{player} uses {source} to have {opponent} return {target} to their hand',
-            handle: context => {
```
ystros, line 25: ☹️

PR #3161, opened 2021-07-16
Fix "place on bottom of deck" effects

### LINKED ISSUE #3160: BUG: Acolyte of the Flame - ability not working
reported in the tournament chat:

> Acolyte of the Flame is bugged. If you choose to put the cards at the bottom the opponent would draw them regardless. 

has been confirmed by multiple players.

ystros:
Returning a card to the deck assumes that the card is already outside of
the deck. For cards that take the top card of the deck and place it on
bottom, PlaceCard should be used instead.

Fixes #3160


EvadLhorg:
lgtm

PR #3163, opened 2021-07-18
fix shagwell and other fools with gold giving each other keywords when the source of those keywords has left play

EvadLhorg:
As pointed out in the HMW FAQ https://docs.google.com/document/d/1iab_YjLgR_80SR0DcSoQ9aBZHrwc9Le9B5wMK0YV22g/edit under 45. Shagwell:

... 
If a character that was giving two or more Fools a keyword via Shagwell's constant effect leaves play or loses that keyword, those characters will lose that keyword too, unless there is another "source". They cannot continue to be each other's only source of that keyword if none of them are gaining it from anywhere else.
...

Two fools with gold should not give each other keywords when the source of those keywords has left play

I think my current fix is more of a workaround, i would appreciate feedback and other solution ideas

### DIFF
--- a/server/game/effects.js
+++ b/server/game/effects.js
@@ -388,10 +388,15 @@ const Effects = {
                 for(let keyword of context.dynamicKeywords[card.uuid]) {
                     card.removeKeyword(keyword);
                 }
-                context.dynamicKeywords[card.uuid] = keywordsFunc(card, context);
-                for(let keyword of context.dynamicKeywords[card.uuid]) {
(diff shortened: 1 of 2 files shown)


ystros:
Like you said, this seems like more of a workaround than a fix. I'm also not 100% convinced it would fix it in all cases as a workaround - it may depend on the order the cards came into play.

Feedback loops like this (one character gains something based on condition A, then another character gains based on the first character gaining while it also fulfills condition A) are tricky.

PR #3165, opened 2021-07-18
fix ShuffleIntoDeck gameaction´s order of actions

EvadLhorg:
Bug description:
When using Balon Greyjoy (KotI)´s ability to put a card into play until the end of the phase and then the phase ends, the ShuffleIntoDeck gameaction is called. After refactoring this gameaction to use an ReturnCardToDeck event (https://github.com/throneteki/throneteki/commit/54c901a01ae0a2b9d39d2c68069d0bb82836adab) instead of calling player.moveCard directly, the ShuffleIntoDeck gameaction is broken as FIRST the shuffling happens and THEN the card is move to the draw deck, ending up on top of the draw deck.

My debugging has gotten me so far as to figure out why that is: player.moveCard raises and event onCardLeftPlay which then calls synchronousMoveCard. This leads to the following order of execution:

ShuffleIntoDeck -> returnCardToDeck -> raise event onCardLeftPlay -> shuffleDeck -> onCardLeftPlay Handler gets executed

I´ve already tried making the sub-events of ShuffleIntoDeck atomic but i couldn´t get it to work and would appreciate help!
I´ve buit a test for Balon Greyjoy which currently succeeds, testing 100 times if the target card of his ability is on the top of the draw deck after the abilitiy´s effect expires (which shouldn´t be the case 100% of the time)


ystros:
I've updated ReturnCardToDeck to have the leave play event simultaneous to returning the card to deck. This doesn't fix the problem, but puts us in a better position.

What needs to happen is that the shuffle events should execute after the handler for returning (and placing) the cards in the draw deck. The event system doesn't make this type of sequencing very easy currently. Need to think more

EvadLhorg:
your latest fix looks good, tested it manually and the balon interaction works as intended. the target card ends up being shuffeld into the deck.

ystros:
It does, but confusingly Lost Message is still broken.

ystros:
Give it a try now.

EvadLhorg:
looks good, tested balon and lost message as well as breaking ties to see if the ShuffleCardIntoDeckCost also works. thank you!

PR #3167, opened 2021-07-21
Update to G.O.T. restricted lists

ystros:
* Uses the newer code and name fields that have been added to
  the RL data instead of calculating it.


EvadLhorg:
lgtm

PR #3174, opened 2021-08-02
Make discard + leave play + placement simultaneous

### LINKED ISSUE #3102: Bound for the Wall and Citadel Archivist
There is a slight problem when Bound for the Wall (R) discards Citadel Archivist. You are prompted to trigger the Archivist before having to choose the character for Bound, so Bound just fizzles. What should happen is that Bound should resolve in full, and only then can the reaction of Citadel Archivist be triggered, if it's still in the discard pile. :)

### LINKED ISSUE #3166: Enter discard triggers before pillage reactions although first player decides
Citadel Archivist is allowed to trigger when pillaged before the first player gets a chance to react with pillage cards such as Asha or Gregor. The triggers are simultaneous, though.

First player also decides "After (military) claim is applied (e.g. Qohor)" vs "after a character is killed (for military claim)". I believe currently, after a character is killed triggers first.

Thanks!

ystros:
* Leave play event is now simultaneous to the discard card event
* Reactions to placing a card in discard are now simultaneous to reactions to discarding the card

Fixes #3166 
Fixes #3102

### DIFF
--- a/server/game/GameActions/DiscardCard.js
+++ b/server/game/GameActions/DiscardCard.js
@@ -1,4 +1,6 @@
 const GameAction = require('./GameAction');
+const LeavePlay = require('./LeavePlay');
+const MoveCardEventGenerator = require('./MoveCardEventGenerator');
 
 class DiscardCard extends GameAction {
     constructor() {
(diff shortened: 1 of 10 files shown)


EvadLhorg:
lgtm!

PR #3177, opened 2021-08-05
Fix cost of Corn, Corn, Corn

### LINKED ISSUE #3148: No prompt for Corn, Corn, Corn
Was playing a game and the challenge phase ended and the necessary conditions were in play to use Corn, Corn, Corn. There was no prompt at the end of the challenge phase to play the event.

ystros:
Also implement the Melee clause about which opponents can be selected,
and fix the message.

Fixes #3148


EvadLhorg:
wow, don´t know how i missed that obvious bug. good fix and never give up on melee 👍

PR #3183, opened 2021-08-09
fix winter maid interrupt so it checks the previous plots traits

### LINKED ISSUE #3176: The Winter Maid is not returning to hand when opponent just revealed summer, even tho summer was not revealed during the interrupt window
Repro steps:
1) start game -> no revelaed plots
2) choose summer plot
3) play winter maid
4) **Problem: winter maid does not return to hand**

**Expected behaviour:**
Winter maid will return to hand UNLESS there was revealed summer plot during the interrupt window.

/-----------------------------

Examples of correct behaviour:

First:
1) start of plot phase, where there isnt summer plot revealed
  2) reveal summer plot
  3) play Winter Maid
  4) winter maid returns to hand

Second:
  1) summer plot revelaed
  2) end of round -> plot phase
  3) choose any plot
  4) play winter maid
  5) winter maid does NOT return to hand

EvadLhorg:
fixing the way the plot phase works was too much of a hassle so i figured i "just" fix winter maid in regards to the interrupt window timing by creating a workaround.
fixes #3176

### DIFF
--- a/server/game/cards/19-JS/TheWinterMaid.js
+++ b/server/game/cards/19-JS/TheWinterMaid.js
@@ -35,7 +35,7 @@ class TheWinterMaid extends DrawCard {
         this.game.addMessage('{0} uses {1} to give the Winter trait to {2}',
             player, this, card);
 
-        if(!this.game.anyPlotHasTrait('Summer')) {
+        if(!this.game.anyPlotHasTraitDuringPlotInterrupt('Summer')) {
             this.game.addMessage('{0} uses {1} to return {1} to their hand instead of their discard pile', player, this);
             player.moveCard(this, 'hand');
         }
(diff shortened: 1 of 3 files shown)


ystros:
This would work, but maybe it's better to fix the timing in general? The main reason the soon-to-be-revealed plot is moved to the `activePlot` location before the interrupt window is so that it'll show up in the UI when there's an interrupt for it (like Winter Maid). Otherwise the interrupt wouldn't show you which plots you would potentially be modifying. Similar to Missandei or other cards that have an interrupt while the card is still in an out-of-play area.

(This fix may be a lot easier than fixing the timing to be correct + associated UI changes)

EvadLhorg:
there are currently only two cards that share this interrupt window (old nan being the other) and only winter maid references plots in play. i agree that it would be better to fix these things in general, but going the route of least resistance for a single card issue might be ok. What do you think?

PR #3188, opened 2021-08-13
Fix improved fortification triggering condition and saving

EvadLhorg:
previously, improved fortification reacted to late for a "would leave play" interrupt. I now used the same triggering condition as dupes (minus the killing part).
furthermore, using context.event.saveCard() should be called, just like for dupes, which in turn calls cancel()


ystros:
Sigh. I guess. We're getting closer to having onCardLeftPlay simultaneous to all events, but not there yet.

Issue #3206, opened 2021-09-17
Implementing Draft


ystros:
* [ ] Admin
  * [ ] Allow a draft cube to be created / updated, along with associated starter deck
  * [ ] Allow an event to be configured to use draft
* [ ] Drafting Table
  * [ ] A drafting table for the event can be started
  * [ ] Players can join the drafting table
  * [ ] Packs are generated from the draft cube
  * [ ] Players draw a pack, pick a card, then pass the remaining card to the next player
  * [ ] Picking continues until all cards have been chosen. Then next pack is drawn.
  * [ ] Drafting continues until all packs for the table have been drafted
  * [ ] Drafted cards are saved to a deck associated with the event.
* [ ] Deck construction
  * [ ] Players construct a final deck based on the starter deck and drafted cards
  * [ ] Decks validated to ensure it only includes cards from the starter deck or were drafted
  * [ ] Decks validated against the event-based deck sizes (default 40 draw cards, 5 plot cards)
* [ ] Draft game
  * [ ] A game can be started for the draft event
  * [ ] Players participating in the draft event can join the game
  * [ ] Players can only select the deck that they constructed from the drafting session
  * [ ] Game starts and proceeds as normal

PR #3207, opened 2021-09-17
Add API endpoint for draft cubes

ystros:
Adds API endpoint to manage draft cubes, and allows events to be
configured with a format and draft options.

Progress toward #3206

### DIFF
--- a/server/services/DraftCubeService.js
+++ b/server/services/DraftCubeService.js
@@ -1,0 +1,56 @@
+const logger = require('../log.js');
+
+class DeckService {
+    constructor(db) {
+        this.draftCubes = db.get('draftCubes');
+    }
+
+    async getAll() {
+        return this.draftCubes.find({})
+            .then(draftCube => {
+                return draftCube;
+            })
+            .catch(err => {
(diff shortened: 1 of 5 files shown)


server/services/DraftCubeService.js
```diff
@@ -0,0 +1,56 @@
+const logger = require('../log.js');
+
+class DeckService {
```
EvadLhorg, line 3: class name does not match file name

server/services/DraftCubeService.js
```diff
@@ -0,0 +1,56 @@
...
+
+        return this.draftCubes.update({ _id: draftCube.id }, { '$set': properties });
+    }
+
+    async delete(id) {
+        return this.draftCubes.remove({ _id: id });
+    }
+}
+
+module.exports = DeckService;
```
EvadLhorg, line 55: export name does not match file name

server/services/DraftCubeService.js
ystros: Fixed!

PR #3217, opened 2021-11-08
fix family duty honor trying to react to cards that have no canBeSaved function

EvadLhorg:


server/game/cards/20-HMW/FamilyDutyHonor.js
```diff
@@ -2,7 +2,7 @@ const DrawCard = require('../../drawcard.js');
 
 class FamilyDutyHonor extends DrawCard {
     setupCardAbilities(ability) {
-        let leftPlayCondition = event => event.allowSave && event.card.canBeSaved() && event.card.hasTrait('House Tully') && event.card.controller === this.controller && event.card.getPower() > 0;
+        let leftPlayCondition = event => event.allowSave && event.card.canBeSaved && event.card.canBeSaved() && event.card.hasTrait('House Tully') && event.card.controller === this.controller && event.card.getPower() > 0;
```
ystros, line 5: I assume this is to handle the left play event for plots. Would it be better to check that the card type isn't plot, or (probably better), define canBeSaved so that it *is* available on all cards?

EvadLhorg: good idea, i´ve put canBeSaved in the baseCard now with the same implementation as in drawCard before

PR #3248, opened 2022-04-28
Fixing bug for "shares a Trait" abilities and added Traits

Patane97:
This should fix a bug with Jinglebell & A Mission in Essos for when Traits are added to that character (eg. The Many-Faced God).

**Jinglebell**: Previously was calculating triggerable traits when Jinglebell entered play (House Frey & Fool only). Now it is calculated when the ability actually triggers.

**A Mission in Essos**: Was checking traits for that character after they were returned to hand, rather than before. I have confirmed with the Design Team that it should take into account that characters card-state whilst in-play for the search condition, and this adjustment does that.

### DIFF
--- a/server/game/cards/17.1-R/AMissionInEssos.js
+++ b/server/game/cards/17.1-R/AMissionInEssos.js
@@ -14,13 +14,13 @@ class AMissionInEssos extends DrawCard {
             handler: context => {
                 this.game.resolveGameAction(
                     GameActions.returnCardToHand(context => ({ card: context.target }))
-                        .then(preThenContext => ({
+                        .then((preThenContext, preThenEvent) => ({
                             gameAction: GameActions.search({
                                 title: 'Select a character',
                                 match: {
                                     type: 'character',
-                                    trait: preThenContext.target.getTraits(),
-                                    printedCostOrLower: preThenContext.target.getPrintedCost() - 1
+                                    printedCostOrLower: preThenEvent.cardStateWhenReturned.getPrintedCost() - 1,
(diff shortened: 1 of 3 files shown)


server/game/cards/20-HMW/Jinglebell.js
```diff
@@ -11,7 +11,7 @@ class Jinglebell extends DrawCard {
                 title: 'Select a card',
                 match: {
                     type: 'character',
-                    trait: this.getTraits()
+                    condition: card => card.getTraits().some(trait => this.hasTrait(trait))
```
ystros, line 14: Ah, that was pretty subtle. Good catch.

server/game/cards/17.1-R/AMissionInEssos.js
```diff
@@ -14,13 +14,13 @@ class AMissionInEssos extends DrawCard {
...
-                        .then(preThenContext => ({
+                        .then((preThenContext, preThenEvent) => ({
                             gameAction: GameActions.search({
                                 title: 'Select a character',
                                 match: {
                                     type: 'character',
-                                    trait: preThenContext.target.getTraits(),
-                                    printedCostOrLower: preThenContext.target.getPrintedCost() - 1
+                                    printedCostOrLower: preThenEvent.cardStateWhenReturned.getPrintedCost() - 1,
+                                    condition: card => card.getTraits().some(trait => preThenEvent.cardStateWhenReturned.hasTrait(trait))
```
ystros, line 23: I think with this one you could have just done `trait: preThenContext.event.cardStateWhenReturned.getTraits()`, since execution is delayed until the `then` clause is evaluated. But this also works. 🤷‍♀️

PR #3265, opened 2022-05-05
fix the princess pass after the refactoring of the loseIcon effect so that the then part of the card triggers in a queued simple step

### LINKED ISSUE #3263: second part of The Prince's Pass doesn't trigger
Not sure if it already exists: the second part of The Prince's Pass doesn't trigger nor open a reaction window. Referring to this part: Then, if it has no challenge icons, you may sacrifice The Prince's Pass to discard it from play.

EvadLhorg:
fixes #3263

i think refactoring the `ability.effects.removeIcon(icon)` part so that it now is a gameaction breaks this particular card. the check for 
`if(this.targetCharacter.getNumberOfIcons() === 0) {` returns false as the number of icons is still 1 at that point as the game action hasn´t resolved yet
seems more like a workaround to be honest, other ideas welcome

### DIFF
--- a/server/game/cards/04.5-GoH/ThePrincesPass.js
+++ b/server/game/cards/04.5-GoH/ThePrincesPass.js
@@ -22,18 +22,21 @@ class ThePrincesPass extends DrawCard {
                     this.game.addMessage('{0} kneels {1} to remove {2} {3} icon from {4}',
                         this.controller, this, icon === 'intrigue' ? 'an' : 'a', icon, this.targetCharacter);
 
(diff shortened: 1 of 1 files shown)


Patane97:
Feels sufficient enough for now. In regards to a more "proper" solution, would it be to update Princes Pass to use GameActions (Might require a "PromptForIcon" gameaction to be added), and add a "thenExecute" to check if icons are 0?
So something like:
- GameActions.promptForIcon (needs to be added first)
- GameActions.loseIcon
- .thenExecute() -> if card has 0 icons, do GameActions.discardCard

PR #3266, opened 2022-05-06
Do not allow Glamor to resurrect killed parent

ystros:
The text for Glamor says "another character killed this phase", so you cannot
select the card killed as a cost to activate its ability.


EvadLhorg:
LGTM

PR #3271, opened 2022-05-07
Add locked deck feature for events

### LINKED ISSUE #3257: Locked decks for tournaments/events
For some tournaments it would be useful if a player can´t change their deck in between rounds.
we would need a way to specify if an event will use locked decks
we would need a way to specify that a deck will be used for a specific event.
we would need to prevent changes to that deck after saving it for the event
we would need to change the deck selection for a player when joining a game of such an event to only show the deck configured for that event

EvadLhorg:
resolves #3257

### DIFF
--- a/server/services/DeckService.js
+++ b/server/services/DeckService.js
@@ -45,6 +45,16 @@ class DeckService {
             });
     }
 
+    removeEventIdAndUnlockDecks(eventId) {
+        return this.decks.find({ eventId: eventId })
+            .then(eventDecks => eventDecks.forEach(deck => {
+                let properties = {
+                    eventId: undefined
+                };
+                this.decks.update({ _id: deck._id }, { '$set': properties });
+            }));
+    }
+
     getStandaloneDecks() {
         return this.decks.find({ standaloneDeckId: { $exists: true } }, { sort: { lastUpdated: -1 } });
     }
@@ -54,8 +64,7 @@ class DeckService {
         if(deck.eventId) {
             //if a deck for the event already exists, do not create the new deck
             if(await this.userAlreadyHasDeckForEvent(deck.username, deck.eventId)) {
-                logger.error('User ' + deck.username + ' already has a deck configured for event ' + deck.eventId);
-                throw new Error('User ' + deck.username + ' already has a deck configured for event ' + deck.eventId);
+                throw new Error(`User ${deck.username} already has a deck configured for event ${deck.eventId}`);
             }
         }
         
@@ -95,7 +104,6 @@ class DeckService {
         let previousVersion = await this.getById(deck.id);
         //do not save the deck if the deck is locked
         if(previousVersion.locked) {
-            logger.error('Locked decks can not be updated');
             throw new Error('Locked decks can not be updated');
         }
 
@@ -103,8 +111,7 @@ class DeckService {
         if(deck.eventId) {
             //if a deck for the event already exists, do not update the deck
             if(await this.userAlreadyHasDeckForEvent(previousVersion.username, deck.eventId)) {
-                logger.error('User ' + previousVersion.username + ' already has a deck configured for event ' + deck.eventId);
-                throw new Error('User ' + previousVersion.username + ' already has a deck configured for event ' + deck.eventId);
+                throw new Error(`User ${previousVersion.username } already has a deck configured for event ${deck.eventId}`);
             }
         }
 
@@ -127,7 +134,6 @@ class DeckService {
         let previousVersion = await this.getById(id);
         //do not delete the deck if the deck is locked
         if(previousVersion.locked) {
-            logger.error('Can not delete a locked deck');
             throw new Error('Can not delete a locked deck');
         }
         return this.decks.remove({ _id: id });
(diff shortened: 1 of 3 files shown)


server/services/DeckService.js
```diff
@@ -37,13 +37,23 @@ class DeckService {
         return this.decks.find({ standaloneDeckId: { $exists: true } }, { sort: { lastUpdated: -1 } });
     }
 
-    create(deck) {
+    async create(deck) {
+        //if the eventId is set on a new deck, check if the user already has a deck with the same eventId
+        if(deck.eventId) {
+            //if a deck for the event already exists, do not create the new deck
+            if(this.userAlreadyHasDeckForEvent(deck.userName, deck.eventId)) {
```
cryogen, line 44: This needs to be awaited or will have unpredictable results

server/services/DeckService.js
```diff
@@ -37,13 +37,23 @@ class DeckService {
...
+                return () => Promise.resolve();
```
cryogen, line 45: Just throw an error instead of resolving to nothing

server/services/DeckService.js
```diff
@@ -37,13 +37,23 @@ class DeckService {
...
+        }
+        
         let properties = {
             username: deck.username,
             name: deck.deckName,
             plotCards: deck.plotCards,
             bannerCards: deck.bannerCards,
             drawCards: deck.drawCards,
+            eventId: deck.eventId,
+            locked: deck.eventId ? true : false, // lock the deck from further changes if the eventId is set
```
cryogen, line 56: This shouldn't be a property of the deck itself, it should be computed when it's needed

server/services/DeckService.js
```diff
@@ -69,13 +79,29 @@ class DeckService {
         return this.decks.insert(properties);
     }
 
-    update(deck) {
+    async update(deck) {
+        let previousVersion = await this.getById(deck.id);
+        //do not save the deck if the deck is locked
+        if(previousVersion.locked) {
+            return () => Promise.resolve();
```
cryogen, line 86: again, throw an error here

server/services/DeckService.js
```diff
@@ -69,13 +79,29 @@ class DeckService {
...
+        }
+
+        //if the eventId is set on the deck, check if the user already has a deck with the same eventId
+        if(deck.eventId) {
+            //if a deck for the event already exists, do not update the deck
+            if(this.userAlreadyHasDeckForEvent(deck.userName, deck.eventId)) {
```
cryogen, line 92: await needed

server/services/DeckService.js
```diff
@@ -69,13 +79,29 @@ class DeckService {
...
+                return () => Promise.resolve();
```
cryogen, line 93: throw an error

server/services/DeckService.js
```diff
@@ -69,13 +79,29 @@ class DeckService {
...
+        }
+
         let properties = {
             name: deck.deckName,
             plotCards: deck.plotCards,
             drawCards: deck.drawCards,
             bannerCards: deck.bannerCards,
             faction: deck.faction,
+            eventId: deck.eventId,
+            locked: deck.eventId ? true : false, // lock the deck from further changes if the eventId is set
```
cryogen, line 104: As above

server/services/DeckService.js
```diff
@@ -84,9 +110,20 @@ class DeckService {
         return this.decks.update({ _id: deck.id }, { '$set': properties });
     }
 
-    delete(id) {
+    async delete(id) {
+        let previousVersion = await this.getById(id);
+        //do not delete the deck if the deck is locked
+        if(previousVersion.locked) {
+            return () => Promise.resolve();
```
cryogen, line 117: throw an error

server/services/DeckService.js
```diff
@@ -84,9 +110,20 @@ class DeckService {
...
+        }
         return this.decks.remove({ _id: id });
     }
+
+    async userAlreadyHasDeckForEvent(userName, eventId) {
+        let otherDecksForThisUser = await this.findByUserName(userName);
```
cryogen, line 123: This should operate on the decks collection directly and filter at source, otherwise you're returning far more records than necessary and the work can be done on the db

cryogen (changes requested):

server/services/DeckService.js
cryogen: I see now that it's used in the client, so if you want to keep that, compute this property when fetching decks

server/services/DeckService.js
EvadLhorg: fixed

server/services/DeckService.js
EvadLhorg: fixed

server/services/DeckService.js
EvadLhorg: how would i go about computing it when fetching decks? do i need to add a handler to every find() method and add a property to the returned decks or is there some base functionality i can use that adds computed properties to fetched entities?

server/services/DeckService.js
EvadLhorg: fixed

server/services/DeckService.js
EvadLhorg: fixed

server/services/DeckService.js
EvadLhorg: fixed

server/services/DeckService.js
EvadLhorg: fixed

server/services/DeckService.js
EvadLhorg: good idea, done

cryogen (approved):

PR #3273, opened 2022-05-08
[WIP] Implement RevealCards GameAction, revamping card reveals & implement Alla Tyrell

### LINKED ISSUE #3252: Implement Alla Tyrell
Alla Tyrell requires a restructure of any ability that "reveals" a card, as currently that reveal has simply been part of the message text in chat. This needs to be handled correctly and tested for each ability, as in some cases Alla's interrupt will prevent part of the ability from completing. For example, if the player plays Summons, and chooses Alla from their deck, she will be revealed and her interrupt will put her into play. Because she is no longer within the original search location (draw deck), the rest of that sentence fizzles out, and she cannot be added to hand (the deck will still shuffle though).

This is the closest search I can find to detect all "reveal" cards (some don't belong here, primarily those relating to revealing plots): https://thronesdb.com/find?q=x%3A%22reveal+it%22%7C%22reveal+them%22%7C%22reveal+that%22&sort=name&view=list

This is also probably the greatest time to properly implement the RevealCard (or RevealCards) game action.

Patane97:
This is Pull Request is still a work in progress. Closes #3252, I'd recommend a "Squash & Merge" when approved.

The intention here is to split the reveal portion of any ability into a separate GameAction, which can be interrupted/reacted to. Currently, we have decided to do the following:

Required tasks:
- [x] Implement GameAction.RevealCards core functionality
- [x] Implement Alla Tyrell
- [x] Implement Reveal into GameActions.Search with 'reveal' boolean
- [x] Implement Reveal into game.promptForDeckSearch with 'reveal' boolean (for legacy reasons)
- [x] Update Reveal Cost builders to fit GameActions.RevealCards
- [x] Adjust messaging for any search/reveal/do something cards
- [x] Check & Update all cards which any form of reveal (potentially affected cards: https://thronesdb.com/find?q=x%3A%22reveal+%22&sort=set&view=list)
- [x] Update card tests to work for new reveal process (eg. accounting for an extra button press to acknowledge revealed cards)
- [ ] Update GameActions & Search functions to "build" the resulting message for chat, based on what actions were actually performed. This could also be used outside of RevealCards
- [x] Updating any non-revealing searches (to simply not reveal, as it reveals by default)


server/game/cards/13.3-PoS/JonConnington.js
```diff
@@ -13,7 +13,7 @@ class JonConnington extends DrawCard {
             message: '{player} uses {source} to reveal {target}',
             handler: context => {
                 this.game.resolveGameAction(
-                    GameActions.revealCard(context => ({ card: context.target })),
+                    GameActions.revealCards(context => ({ card: context.target })),
```
ystros, line 16: Don't modify any of the existing `revealCard` cards, I'm updating them on my own branch.

server/game/GameActions/RevealCards.js
```diff
@@ -0,0 +1,106 @@
...
+        const eventParams = {
+            player,
+            cards: Array.isArray(cards) ? cards : [cards],
+            source: context.source
+        }
+        return this.event('onCardsRevealed', eventParams, event => {
+            this.whileRevealed = whileRevealed || this.defaultWhileRevealed;
+            this.playerRevealFunc = (card, player) => event.cards.includes(card) && player === event.player;
+            this.allPlayersRevealFunc = card => event.cards.includes(card);
+            this.revealingTo = context.game.getPlayers();
```
ystros, line 31: Do not add data to the `this` object. GameActions are intended to be stateless; by modifying a `this` property inside the event handler, you are modifying it for ALL `RevealCards` objects.

The `defaultWhileRevealed` in the constructor is fine because it's the same for all RevealCards. But everything else here needs to change.

server/game/costs/RevealCost.js
```diff
@@ -1,14 +1,16 @@
+const GameActions = require('../GameActions');
+
 class RevealCost {
     constructor() {
         this.name = 'reveal';
     }
 
     isEligible(card) {
-        return card.location === 'hand';
+        return ['hand', 'plot deck', 'shadows'].includes(card.location);
```
ystros, line 9: For consistency, this would probably be better expressed using the `RevealCards.allow` method itself, so that we're not repeating implementations.

server/game/gamesteps/DeckSearchPrompt.js
```diff
@@ -47,38 +49,67 @@ class DeckSearchPrompt extends BaseStep {
...
+                return;
+            }
+        }
+        // Do not continue to onSelect if single result is not in draw deck
+        else if(result.location !== 'draw deck') {
+            result = null;
+            return;
+        }
+
+        this.properties.onSelect(player, result);
```
ystros, line 110: What is this trying to prevent?

Additionally, you don't need to set `result = null` if you're returning immediately after.

server/game/costs/RevealCost.js
Patane97: This is slightly different to `RevealCards.allow`, for revealing as a cost will require a hidden location which you have access to, but your opponents do not (your hand, plot deck or shadows). For example, you cannot reveal cards in your deck as a cost without them already being revealed (thus a cost being impossible).

server/game/gamesteps/DeckSearchPrompt.js
Patane97: This is the logic to ensure that it only continues to `onSelect` if there are search results left in your draw deck after the reveal has happened.

The `result = null` _can_ be removed I believe; this is more of an artefact to me copying the same process as `GameActions.Search`, and wasn't sure if I needed to update result or not. In hindsight, its irrelevant here as result is only used in `onSelect`.

server/game/GameActions/RevealCards.js
Patane97: Will be adjusted/fixed in latest commit.

Related to this then, should I also be changing `this.previousSelectCards` (in `highlightRevealedCards` and `hideRevealedCards`)? Not sure of the best way of tracking that otherwise.

server/game/cards/13.3-PoS/JonConnington.js
Patane97: Reverted `Jon Connington` and `Highgarden` adjustments in 67ba4a5b9b3ac39ba257adcbc3de7d867395be2f.

I do recommend we keep it as the updated `GameActions.revealCards`, to keep things consistent; might be best to do that in your branch or whichever branch is being pushed last. 😄

server/game/GameActions/RevealCards.js
ystros: Anything that uses `this` needs to be changed.

Patane97: Fixed in 4ea24e76c5eb05fa56ddaffcccd34aa31b4856fc and 1155082d26c69f55c0204874c4080920304f301d

server/game/GameActions/RevealCards.js
```diff
@@ -0,0 +1,107 @@
...
+                    card,
+                    cardStateWhenRevealed: card.createSnapshot(),
+                    player: event.player,
+                    opponents: event.opponents,
+                    source: event.source
+                }
+                
+                event.thenAttachEvent(this.event('onCardRevealed', revealEventParams, () => {}));
+            }
+            context = Object.assign({ revealed: event.cards, player: event.player, opponents: event.opponents }, context);
```
ystros, line 48: I'm not sure we want to do this. This is creating a plain Javascript object, whereas the `context` being passed in is likely an `AbilityContext` object.

While it might not matter in most cases, I think it could be a problem if we have a `whileRevealed` game action, which could be using methods that would not be copied over this way.

I think I'd rather see us create a new context object (as we do in the ThenClauseAbility), or do something simple like:
```
context.revealData = { ... }
```

Or even just:
```
context.revealedCards = event.cards
```

since `player` and `opponents` are not used in here and may not be needed in the `whileRevealed` action.

server/game/GameActions/RevealCards.js
```diff
@@ -0,0 +1,107 @@
...
+        cards = Array.isArray(cards) ? cards : [cards];
+        return cards.length > 0 && cards.some(card => this.isInHiddenArea(card));
+    }
+
+    createEvent({ cards, player, whileRevealed, context }) {
+        const allPlayers = context.game.getPlayers();
+        const eventParams = {
+            player,
+            opponents: allPlayers.filter(p => p !== player),
+            cards: Array.isArray(cards) ? cards : [cards],
```
ystros, line 26: I think I'd rather see us always pass it as an array, rather than try and figure out whether it's an array or not. `cards` already implies that it could be more than one.

server/game/GameActions/RevealCards.js
```diff
@@ -0,0 +1,107 @@
...
+                    context.revealed = context.revealed.filter(card => this.isInHiddenArea(card));
+                    event.revealed = context.revealed;
+                    // Reveal all remaining cards to all players
+                    if(context.revealed.length > 0) {
+                        context.game.cardVisibility.addRule(allPlayersRevealFunc);
+                        this.highlightRevealedCards(event, context.revealed, allPlayers);
+                    }
+                }
+            });
+            event.thenAttachEvent(prepareRevealCards.createEvent(context));
```
ystros, line 64: No need to wrap this in a `HandlerGameActionWrapper`. Since we always want to do this and don't need to check the `allow`, you can just create an event directly.

server/game/GameActions/RevealCards.js
Patane97: This has been fixed/addressed in e2661e091c5a94eb156a653cf08385a597cf4e74. I'd recommend resolving this comment.

server/game/GameActions/RevealCards.js
Patane97: Adjusted in e2661e091c5a94eb156a653cf08385a597cf4e74. I've had to adjust `GameActions.Search` and `DeckSearchPrompt` to suit.

server/game/GameActions/RevealCards.js
Patane97: This event has been removed in e2661e091c5a94eb156a653cf08385a597cf4e74, as it's luckily not required anymore due to revealing happening before reveal interrupts, with `context.revealed` being filtered within the  `onCardRevealed` event.

server/game/costs/RevealCost.js
Patane97: On second thought, it's probably better to keep this as flexible as possible. Eg. there _could_ be an ability which says to "Search your deck for a card. Reveal that card to [do something unusual]", so I'll implement your thoughts now.

Patane97:
This code update has been pushed to the Playtesting instance instead, and will be merged with Live when `As High as Honor` is implemented.

PR #3293, opened 2022-07-10
update throneteki-json-data to include the newest restricted list

EvadLhorg:


EvadLhorg:
not necessary anymore

PR #3300, opened 2022-08-05
Fix pyromancers cache so that the action works even though the parent card is blanked

EvadLhorg:
progress towards #2377

### DIFF
--- a/server/game/cardaction.js
+++ b/server/game/cardaction.js
@@ -51,6 +51,7 @@ class CardAction extends BaseAbility {
         this.location = properties.location || DefaultLocationForType[card.getPrintedType()] || 'play area';
         this.events = new EventRegistrar(game, this);
         this.activationContexts = [];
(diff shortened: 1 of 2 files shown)


ystros:
This seems more like a bandaid than a proper engine fix. So while I'd be willing to merge this, I do not believe we should remove the TODO entirely. Some TODO should remain as a clue that there's more work that needs to be done here.

A proper engine fix would likely look like creating the action in `gainText` with the proper source card (instead of pretending it was generated by the card it's being applied to), or instead of checking blank in the ability then rephrase all action abilities as a persistent effect that causes the card to gain the action ability text. (this is similar to how keywords work currently and probably is the correct solution).

EvadLhorg:
Finally getting back to answering your comment:

I am not 100% sure i understand you suggestion for a fix right. I think if we add the gainText with the proper source card (in this case pyromancer´s cache, an attachment), the action (draw a card) would be treated as an attachment action and not as a location action.

PR #3324, opened 2022-12-22
Implementing As High as Honor, Card Reveal & relevant upgrades

### LINKED ISSUE #3252: Implement Alla Tyrell

Patane97:
**When approved, I would recommend Squash & Merge!!**

**What this merge achieves:**
- Implementation of all 40 AHaH cards (all cards implemented).
- Relevant engine tweaks to fit certain cards, such as Lysono Maar & Anya Waynwood/Lord Protector of the Vale.
- Adding quite a few relevant `GameActions`, to continue building our stock of valid game action abilities.
- Implementation of revealing cards (originally scheduled for BtB release). This should close #3252 and close #3273, and has been playtested for a few months on the PT server. Most of this has already been reviewed, but a final review is probably good.

**What is remaining:**
- [x] Identify & remove any development/playtesting fluff (such as [.github/workflows/node.js.yml](https://github.com/throneteki/throneteki/blob/58dd9551d65244ff29324d99cd39958ed91d0ff0/.github/workflows/node.js.yml) changes)
- [x] Approve & Merge https://github.com/throneteki-playtesting/throneteki/pull/271
- [x] Approve & Merge https://github.com/throneteki/throneteki-json-data/pull/143
- [x] Approve & Merge https://github.com/throneteki/throneteki-client/pull/181
- [x] Approve & Merge https://github.com/throneteki/throneteki-deck-helper/pull/15
- [x] Unit test for Contributing Strength (Within contribute-str merge: https://github.com/throneteki-playtesting/throneteki/pull/271)
- [x] Unit test for Anya Waynwood ~~or Lord Protector of the Vale~~
- [x] Unit test for A Traitor's Whisper (as it tests a new type of selection)

**In my personal code review, I've noted a few things that will eventually need to be done post-AHaH (but is not required for release), and will ensure there is an issue for each:**
- Add lingering/lasting effect GameActions (eg. `UntilEndOfPhase`)
- Add `replaceHandler` GameAction (maybe `replaceGameAction`?)
- Add `selectPrompt` GameAction (Likely a task for @ystros at some point)
- Add `swapCards` GameAction (for cards like Readers Septon; bonus points for preserving the index of the swapped cards)
- Allow GameActions for `target` & `choosePlayer` abilities
- Add `whileLooking` option for `GameActions.lookAtDeck` (similar concept to `whileRevealed` for `GameActions.revealCards`). Update cards like **Doran Martell (AHaH)** to select one of the cards whilst looking.
- Fix Lord Protectors Retainer (and Fear Cuts Deeper Than Swords) to also prevent keywords (eg. stealth & intimidate)
- Resolve https://github.com/throneteki-playtesting/throneteki/issues/203 (Will transfer issue to live repo)
- Update events system to account for aggregates within card files (discussed in https://github.com/throneteki-playtesting/throneteki/pull/269)


.github/workflows/node.js.yml
```diff
@@ -5,9 +5,11 @@ name: Node.js CI
 
 on:
   push:
-    branches: [ "master" ]
+    branches: 
+      - '*'
```
ystros, line 9: It's unclear to me whether this is desirable on the main repo

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "throneteki-json-data"]
 	path = throneteki-json-data
-	url = https://github.com/throneteki/throneteki-json-data
+	url = https://github.com/throneteki-playtesting/throneteki-json-data
```
ystros, line 3: This change definitely should not be made on the main repo

server/game/AbilityTarget.js
```diff
@@ -139,7 +139,7 @@ class AbilityTarget {
                 return true;
             },
             onCancel: () => {
-                if(this.ifAble) {
+                if(this.selector.rejectAllowed(context)) {
```
ystros, line 142: I'm not sure it's a good idea to push down even more targeting game logic into selectors, which were meant to be more about generic card selection prompts. The current implementation is already fuzzy, this breaks it down even further.

server/game/GameActions/PlaceCard.js
```diff
@@ -6,14 +6,9 @@ class PlaceCard extends GameAction {
...
-    canChangeGameState({ player, card, location, bottom = false }) {
-        player = location === 'play area' ? player || card.controller : card.owner;
-
+    canChangeGameState({ card, location }) {
         if(location === 'draw deck') {
-            return (
-                player.drawDeck[0] !== card && !bottom ||
-                player.drawDeck[player.drawDeck.length - 1] !== card && bottom
-            );
+            return true;
```
ystros, line 11: This change implies that moving the top card of your deck to the top of the deck (or bottom to bottom) constitutes a change in board state. Is that correct from a rules perspective?

server/game/GameActions/StandCard.js
```diff
@@ -1,12 +1,17 @@
...
         super('stand');
     }
 
+    message({ card }) {
+        return Message.fragment('stands {card}', { card });
+    }
+
     canChangeGameState({ card }) {
-        return ['faction', 'play area'].includes(card.location) && card.kneeled;
+        return card && ['faction', 'play area'].includes(card.location) && card.kneeled;
```
ystros, line 14: Do you remember what problem required the addition of the `card &&` check? It seems like if there's a problem with null `card` objects being passed into game actions, it would apply to _all_ actions that take a card.

server/game/PlayActions/MarshalCardAction.js
```diff
@@ -42,7 +42,7 @@ class MarshalCardAction extends BaseAbility {
         };
         context.game.raiseEvent('onCardMarshalled', params, () => {
             context.player.putIntoPlay(context.source, 'marshal');
-            context.game.addMessage(this.getMessageFormat(params), context.player, context.source, params.originalController, params.originalLocation, params.originalParent, context.costs.gold);
+            context.game.queueSimpleStep(() => context.game.addMessage(this.getMessageFormat(params), context.player, context.source, params.originalController, params.originalLocation, params.originalParent, context.costs.gold));
```
ystros, line 45: What's the purpose of delaying these messages (along with the other PlayActions) by wrapping them in `queueSimpleStep`? It will make the game log non-linear if there are interrupts (and potentially even reactions). e.g.

With queue:
```
Player 2 plays X (interrupt to player 1 marshalling Y)
Player 1 marshals Y
```

Without queue:
```
Player 1 marshals Y
Player 2 plays X (interrupt to player 1 marshalling Y)
```

server/game/SimultaneousEvents.js
```diff
@@ -9,6 +9,7 @@ class SimultaneousEvents {
     }
 
     addChildEvent(event) {
+        event.parentEvent = this;
```
ystros, line 12: Can't remember if not having this before was an oversight, or if SimultaneousEvent is special in some way and adding this will Cause Problems ™️

server/game/basecard.js
```diff
@@ -366,8 +375,8 @@ class BaseCard {
         return this.keywords.getValues();
     }
 
-    hasPrintedKeyword(keyword) {
-        return this.printedKeywords.includes(keyword.toLowerCase());
+    hasPrintedKeyword(keyword, ignoreValue = false) {
+        return this.printedKeywords.some(printedKeyword => ignoreValue ? printedKeyword.includes(keyword) : printedKeyword === keyword);
```
ystros, line 379: As best as I can tell, this new parameter is never actually used. (And I would prefer a different method rather than this "ignoreValue" boolean if it were)

server/game/PlayActions/MarshalCardAction.js
Patane97: I had definitely done this for a reason, at least initially (whilst implementing all cards for the reveal update), but I'll need to re-investigate why this was the case. Could be legitimate, could also just be left-over changes I forgot to revert.

server/game/AbilityTarget.js
Patane97: @ystros Fair enough, but there were adjustments required for "A Traitor's Whisper" to properly function. This adjustment (as well as adding `rejectAllowed` to `BaseCardSelector.js` and `EachPlayerCardSelector.js`) is there to ensure players can only "cancel" the selection if they have no selections they can make; otherwise they are forced to.

Any alternate suggestions on how to do this? Otherwise, we will need to keep `rejectAllowed`, at least for now.

.github/workflows/node.js.yml
Patane97: Reverted in 31fa026dc6b1e7e530b9fea7fef084789d887b57

Can mark as resolved.

.gitmodules
Patane97: Reverted in 31fa026dc6b1e7e530b9fea7fef084789d887b57

Can mark as resolved.

server/game/GameActions/PlaceCard.js
Patane97: Unsure as to why I had adjusted this a while ago; regardless, reverted in 31fa026dc6b1e7e530b9fea7fef084789d887b57. We may notice later if it was for a reason 😄

Can mark as resolved.

Patane97: Also reverted `PlaceCard.spec.js` in b43c431991dcf7d6ee651be74c8084d00e98b69b (as it was adjusted with the GameAction changes)

server/game/GameActions/StandCard.js
Patane97: I believe this was an old attempt to fix a concern of when `card` was null, but then found another way. Simply some code I forgot to revert, good catch!

Reverted in 31fa026dc6b1e7e530b9fea7fef084789d887b57

Can mark as resolved.

server/game/SimultaneousEvents.js
Patane97: This is, currently, just used as part of an adjustment for **Tywin Lannister (LoCR)** to ensure that only 1 card is being discarded simultaneously. Now that I think about it, it could easily be an aggregated event, and checking if the `event.length === 1`; not entirely sure why I didn't go with that. I'll consider, and maybe revert (with Tywin changes)

server/game/basecard.js
Patane97: Yep, leftover code from when **Doran Martell (AHaH)** was doing something whacky by giving all cards in shadows `Shadow (3)` rather than `Shadow (X)`.

Regardless, reverted in 31fa026dc6b1e7e530b9fea7fef084789d887b57

Can mark as resolved.

ystros:
No glaring issues after looking over the new AHaH cards. The only strange one was The Reader's Septon (swapping the top and bottom card). I'm sure it works, and it would require implementing a new game action to do the swap to really clean up; not worth delaying the pack for.

server/game/AbilityTarget.js
ystros: Ah, so the problem is the `if able` applies per player, rather than the overall selection (e.g. if you can select for one player, you must do so)?

That probably means that the very existence of `EachPlayerCardSelector` doesn't match where I think we should be headed towards. Which means changing it to something else will be a much larger project.

Resolving for now.

server/game/PlayActions/MarshalCardAction.js
Patane97: @ystros Honestly couldn't find the proper reason for this adjustment, and reverting it does not fail any tests. I'll revert it, and if we find any issues in the future, we can deal with it then.

Reverted in https://github.com/throneteki-playtesting/throneteki/commit/143c98f861c602e4ac998bbcf9c06a0b9ad81211

Can mark this as resolved! :)

PR #3330, opened 2023-01-20
Fixing search abilities not filtering valid attachment targets

Patane97:
There is a bug for cards (namely "Black Markey Merchant" and "Gifts for the Widow") which allow the attachment to be put on any character, when the ability specified "character you control". These cards were updated as part of the search/reveal update within AHaH, and this must've been missed.

The resolution here is definitely not the cleanest, but it does the job quickly with what we currently have; I think we need to clean up `GameActions.putIntoPlay` in the future (eg. moving those methods out of `Player` and into GameActions), and this could be a part of that.

### DIFF
--- a/server/game/GameActions/PutIntoPlay.js
+++ b/server/game/GameActions/PutIntoPlay.js
@@ -18,9 +18,11 @@ class PutIntoPlay extends GameAction {
         return Message.fragment(message, { card, controller: card.controller, originalLocation: card.location });
     }
 
-    canChangeGameState({ player, card }) {
+    canChangeGameState({ player, card, attachmentTargets }) {
         player = player || card.controller;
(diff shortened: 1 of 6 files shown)


EvadLhorg:
looks ok to me, maybe rename attachmentTargets to attachmentTargetMatcher to make it more clear that this is a function to match cards and not a list of target cards.
what do you think @ystros  ?

Patane97:
> looks ok to me, maybe rename attachmentTargets to attachmentTargetMatcher to make it more clear that this is a function to match cards and not a list of target cards. what do you think @ystros ?

For reference, I named it `attachmentTargets` as it eventually feeds that function into the `targets` of `player.promptForAttachment(card, playingType, targets)`. Seemed appropriate to keep the same naming scheme there (since it's called targets, not targetMatcher).

PR #3347, opened 2023-02-08
Boros to be able to pick lanni kingsguard

adkateki:
Not sure if this is the way to use CardMatcher, just proposing in case it can save some time in order to fix Boros.

### DIFF
--- a/server/game/cards/19-JS/SerBorosBlount.js
+++ b/server/game/cards/19-JS/SerBorosBlount.js
@@ -11,7 +11,7 @@ class SerBorosBlount extends DrawCard {
             gameAction: GameActions.search({
                 title: 'Select a character',
                 topCards: 10,
-                match: { type: 'character', trait: ['Kingsguard', 'Knight'], not: { faction: 'lannister' } },
+                match: { type: 'character', or: { trait: ['Knight'], not: { faction: 'lannister' } }, or: { trait: ['Kingsguard'] } },
                 reveal: false,
                 message: '{player} {gameAction}',
                 gameAction: GameActions.putIntoPlay(context => ({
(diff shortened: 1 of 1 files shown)


server/game/cards/19-JS/SerBorosBlount.js
```diff
@@ -11,7 +11,7 @@ class SerBorosBlount extends DrawCard {
             gameAction: GameActions.search({
                 title: 'Select a character',
                 topCards: 10,
-                match: { type: 'character', trait: ['Kingsguard', 'Knight'], not: { faction: 'lannister' } },
+                match: { type: 'character', or: { trait: ['Knight'], not: { faction: 'lannister' } }, or: { trait: ['Kingsguard'] } },
```
Patane97, line 14: Close; the `or` is relatively new to the matcher, and simply takes in an array of "or matches" to match. For example, the above should look like this:
```suggestion
                match: { type: 'character', or: [{ trait: ['Knight'], not: { faction: 'lannister' } }, {trait: 'Kingsguard'}] },
```
The above `or` is an array containing the element `{ trait: ['Knight'], not: { faction: 'lannister' } }` and the element `{trait: 'Kingsguard'}`

Happy to approve once adjusted :)

Patane97 (changes requested):
Appreciate the quick suggested fix! Unfortunately didn't pass our unit tests, but I've commented below where the adjustment should go. :)

server/game/cards/19-JS/SerBorosBlount.js
adkateki: Hi! Just updated it!

Patane97 (approved):
Looks good now!

PR #3370, opened 2023-06-07
fix SerDavenLannister reaction triggering when opponent has no cards …

LaranSoft:
Ser Daven Lannister reaction can be triggered only if each opponent has at least one card in hand, due to the "choose" restriction.
Ideally, the check should be done when checking legality of targets. In this case, the number of targets is equal to the number of opponents and each opponent must choose exactly one target.


Patane97 (approved):
Looks good!

PR #3371, opened 2023-06-07
fix OneTwoThree giving insight to all eligible targets instead to the…

LaranSoft:
Steps to reproduce: declare an INT challenge with N characters (N > 1), win the challenge and use One, Two, Three on one of the participating character. N cards will be drawn for Insight.
The implementation of "Protectors of The Realm" (00002), which have a similar effect, declares only context.target as "match", instead of One, Two, Three which declares context.targets.insight, matching all eligible targets.


Patane97 (approved):
Good catch! Missed this with my own quick check.

PR #3372, opened 2023-06-07
fix WardensOfTheEast weird interactions with card moving in and out f…

LaranSoft:
Steps to reproduce: with Wardens of the East out and a Scheme plot revealed, marshall 2x "A very large Shadow" in the shadows, then bring one of them out and back to hand. Notice that the card is still revealed, even though it is not in the shadow anymore. From this point, if that copy of "A very large Shadow" is marshalled again in the shadow, it remains revealed, but if other cards are marshalled in the shadow then those cards will remain facedown (even in the shadow area).

Problem is caused by the definition of the "revealFunc" that is bound to a very specific "snapshot" of the shadow zone of that player. When effect is applyed again (reApply method) the shadow zone changes, but the revealedFunc continues to track the content of the snapshotted shadow zone.

With the fix, the "revealFunc" bounds to the player itself, so that a card is revealed only if belongs to the shadow area of that player at the time the revealFunc is appied. This way, when "A very large Shadow" goes back in hand from shadow zone, it is not revealed anymore, and when a new card is put in the shadow zone it is revealed as well.

Note: the "revealShadows" effect is only used by card WardensOfTheEast (for now)

### DIFF
--- a/server/game/cards/23-AHaH/Sweetrobin.js
+++ b/server/game/cards/23-AHaH/Sweetrobin.js
@@ -5,7 +5,7 @@ class Sweetrobin extends DrawCard {
     setupCardAbilities(ability) {
         this.interrupt({
             when: {
-                onCardRevealed: event => (!event.card.hasPrintedCost() || (event.card.hasPrintedCost() && event.card.getPrintedCost() <= event.card.owner.getTotalInitiative())) && ['hand', 'draw deck', 'shadows'].includes(event.card.location)
+                onCardRevealed: event => event.card.hasPrintedCost() && event.card.getPrintedCost() <= event.card.owner.getTotalInitiative() && ['hand', 'draw deck', 'shadows'].includes(event.card.location)
             },
             limit: ability.limit.perPhase(1),
             message: {
--- a/server/game/cards/23-AHaH/Sweetrobin.js
+++ b/server/game/cards/23-AHaH/Sweetrobin.js
@@ -5,7 +5,7 @@ class Sweetrobin extends DrawCard {
     setupCardAbilities(ability) {
         this.interrupt({
             when: {
-                onCardRevealed: event => event.card.hasPrintedCost() && event.card.getPrintedCost() <= event.card.owner.getTotalInitiative() && ['hand', 'draw deck', 'shadows'].includes(event.card.location)
+                onCardRevealed: event => (!event.card.hasPrintedCost() || (event.card.hasPrintedCost() && event.card.getPrintedCost() <= event.card.owner.getTotalInitiative())) && ['hand', 'draw deck', 'shadows'].includes(event.card.location)
             },
             limit: ability.limit.perPhase(1),
             message: {
--- a/server/game/effects.js
+++ b/server/game/effects.js
@@ -1452,7 +1452,7 @@ const Effects = {
             targetType: 'player',
             apply: function(player, context) {
                 const shadows = player.shadows;
-                const revealFunc = reveal => shadows.includes(reveal);
+                const revealFunc = reveal => player.shadows.includes(reveal);
 
                 context.revealShadows = context.revealShadows || {};
                 context.revealShadows[player.name] = {
--- a/server/game/effects.js
+++ b/server/game/effects.js
@@ -1451,27 +1451,47 @@ const Effects = {
         return {
             targetType: 'player',
             apply: function(player, context) {
-                const revealFunc = reveal => player.shadows.includes(reveal);
+                // making a snapshot of the shadows area through a shallow copy. This way eventual changes to
+                // player.shadows array content (first level) are not reflected in the snapshot
+                const shadows = [...player.shadows];
 
+                // we define the reveal function just once, so it must be bound to the player to keep track of the
+                // content of the shadows area when the function is called by the engine
+                const revealFunc = reveal => player.shadows.includes(reveal);
+                context.game.cardVisibility.addRule(revealFunc);
+
+                // saving in the context a reference to the revealFunc (so we can remove it from the engine when the
+                // effect expires) and the actual snapshot of the cards in the shadows area for which we are resolving
+                // the reveal game action (to avoid a second resolution of the same game action in case of reapply)
                 context.revealShadows = context.revealShadows || {};
                 context.revealShadows[player.name] = {
                     revealFunc,
-                    controller: player
+                    revealed: shadows
                 };
-                context.game.cardVisibility.addRule(revealFunc);
 
-                context.game.resolveGameAction(GameActions.revealCards({
-                    cards: player.shadows,
-                    player,
-                    revealWithMessage: false,
-                    highlight: false,
-                    source: context.source
-                }), context);
+                // resolving the 'reveal' game action for cards in shadows (if there are any)
+                if(shadows.length > 0) {
+                    context.game.resolveGameAction(GameActions.revealCards({
+                        cards: shadows,
+                        player,
+                        revealWithMessage: false,
+                        highlight: false,
+                        source: context.source
+                    }), context);
+                }
             },
             reapply: function(player, context) {
-                const newReveals = player.shadows.filter(card => !context.revealShadows[player.name].controller.shadows.includes(card));
+                // making a snapshot of the shadows area through a shallow copy. This way eventual changes to
+                // player.shadows array content (first level) are not reflected in the snapshot
+                const shadows = [...player.shadows];
 
-                // Only trigger reveal event for newly revealed cards
+                // calculating the newly revealed cards
+                const newReveals = shadows.filter(card => !context.revealShadows[player.name].revealed.includes(card));
+
+                // updating the context reference with the actual snapshot of the cards in the shadows area
+                context.revealShadows[player.name].revealed = shadows;
+
+                // Only trigger reveal event for newly revealed cards (if there are any)
                 if(newReveals.length > 0) {
                     context.game.resolveGameAction(GameActions.revealCards({
                         cards: newReveals,
--- a/server/game/effects.js
+++ b/server/game/effects.js
@@ -1451,18 +1451,17 @@ const Effects = {
         return {
             targetType: 'player',
             apply: function(player, context) {
-                const shadows = player.shadows;
                 const revealFunc = reveal => player.shadows.includes(reveal);
 
                 context.revealShadows = context.revealShadows || {};
                 context.revealShadows[player.name] = {
                     revealFunc,
-                    revealed: shadows
+                    controller: player
                 };
                 context.game.cardVisibility.addRule(revealFunc);
 
                 context.game.resolveGameAction(GameActions.revealCards({
-                    cards: shadows,
+                    cards: player.shadows,
                     player,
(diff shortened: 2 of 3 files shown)


Patane97 (changes requested):
Great catch; explained requested changes in comment. 😄

server/game/effects.js
```diff
@@ -1452,7 +1452,7 @@ const Effects = {
             targetType: 'player',
             apply: function(player, context) {
                 const shadows = player.shadows;
-                const revealFunc = reveal => shadows.includes(reveal);
+                const revealFunc = reveal => player.shadows.includes(reveal);
```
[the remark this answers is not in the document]
LaranSoft: Something gone wrong. I reverted all my changes and switched to using `let` instead of `const` in the declaration of "shadows":

`let shadows = player.shadows;`

but then the bug riappeared.

I think a possible reason is: `const` cannot be reassigned once assigned, but the referenced value can be changed (if not immutable, like strings). For example:

`var data = [0, 1, 2];`
`const constData = data; //constData is [0, 1, 2]`
`data.push(3); //now constData is [0, 1, 2, 3] as well`
`constData.push(4); //now both data and constData are [0, 1, 2, 3, 4]`
`constData = []; // trying to reassign a new value raise an exception "Uncaught TypeError: Assignment to constant variable."`

Maybe here (but it is only a supposition) is happening a reassignment of `player.shadows` attribute. Something like this:

`const shadows = player.shadows; //both shadows and player.shadows refers to the same array`
`const revealFunc = reveal => shadows.includes(reveal); // revealFunc is bound to the array referenced by const shadows`
`// ...`
`// now my hypothesis is that, in some other part of the code, the player.shadows attribute is reassigned. Something like this:`
`player.shadows = player.shadows.filter(() => {....}); `
`// the above instruction creates a new array (shallow copy) and now the array that was previously referenced by player.shadows and by const shadows is different from the array that is referenced by player.shadows. The result is that changes made to player.shadows are not reflected to the const shadows variable... but the revealFunc is bound to that variable. Continuing my previous example, maybe is happening something like this:`

`//both data and constData are [0, 1, 2, 3, 4]`
`var dataB = data.filter(() => {return true;}); //dataB is [0, 1, 2, 3, 4] but is stored in a different "memory area"`
`dataB.push(5); //now dataB is [0, 1, 2, 3, 4, 5] but data and constData are [0, 1, 2, 3, 4]`
`data.push(6); //dataB is [0, 1, 2, 3, 4, 5] and data and constData are [0, 1, 2, 3, 4, 6]. Values have diverged.`

If my hypothesis is correct, this is also the reason why my previous fix actually works. 
By bounding the `revealFunc` to the `player` object, we are able to access the `player.shadows` attribute when the `revealFunc` is executed, reading from the actual "memory area" referenced by `player.shadows`.

Last note: the `const shadows = player.shadows;` declaration remains unchanged, so line 1460 (and others) are safe. The only change is in the definition of the `revealFunc` function, so there should be no other side effect in my opinion.

Let me know if this make sense for you :)

About test case: absolutely yess. I noticed other test cases, and I would love to do test cases for this case (and maybe also for the other two fixes, if you are ok with it). I will try to write test case by myself, eventually I will contact you :)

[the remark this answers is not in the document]
Patane97: Good investigation. I think you're right in your suspicion of `player.shadows` being reassigned; it seems that the method which handles cards being removed from piles does exactly this within [player.js](https://github.com/Patane97/throneteki/blob/15c772813e6ae0718239a9f47c58847ed7949231/server/game/player.js#L840-L842), which is called when a card is removed from a pile (which I believe is an atomic action of cards leaving shadows, but haven't 100% confirmed this).

That being the case, you're correct in your original adjustment, however the `const shadows` needs to be accounted for; we require this to track which cards are _currently_ being revealed by this effect [here](https://github.com/Patane97/throneteki/blob/15c772813e6ae0718239a9f47c58847ed7949231/server/game/effects.js#L1458-L1461) so that, within the "reapply" of that effect (which gets called whenever the game-state changes), it only applies the "revealCards" GameAction to _new_ cards which have entered shadows; therefore it will need the current list of shadow cards from the last time it was applied or reapplied. Note that the "revealCards" GameAction is ultimately what's responsible for triggering interrupts & reactions to cards being revealed, such as Sweetrobin.

Since `player.shadows` is always being reassigned, we likely should just remove `const shadows = player.shadows`, and always check `player.shadows` directly in the `apply` and `reapply` functions of this effect.

Happy to explain further if you need, as this relatively new reveal code can get a bit strange at times 😄.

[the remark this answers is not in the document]
Patane97: I haven't had the chance to physically test it, but I'm still not sure that the change to what's stored in `context.revealShadows[player.name]` is correct; that **should** be essentially capturing a "snapshot" of that players shadow cards at the moment of **apply**ing or **reapply**ing the effect (essentially doing what the initial issue for the `revealFunc` was). When the effect is reapplied, it uses `context.revealShadows` to know what cards have already been revealed via GameActions.revealCards in a previous apply/reapply, and thus should not be revealed again. If we're storing the player reference in `context.revealShadows[player.name]`, then `player.shadows` will always be equal to `context.revealShadows[player.name].controller.shadows` when the effect is being reapplied, so it won't detect any changes & won't trigger `GameActions.revealCards` on any new cards, I think.

I can give this an actual test sometime tomorrow at work maybe to confirm if I'm right there. :)

The test looks great though!

Patane97 (changes requested):
Great work with the test - explained the rest further in a comment. Maybe I could explain a little more of what GameActions are when I'm next free to help understand the issue here. :)

LaranSoft:
Hey, after your last comment, I've understand better the "revealAction" things you told me [here](https://github.com/throneteki/throneteki/pull/3372#discussion_r1222285505)

One thing is the user experience, and from an user experience perspective the revealing / hiding of the cards entering / exiting the shadows area was working. 
But another thing is the engine itself, which must process correctly all game actions that are happening. So, when a card change the status from "hidden" to "revealed", an appropriate game action must be resolved to give the opportunity to other cards that are interested in to react.

In my last commit I think to have addressed that, taking appropriate snapshot "freezed" at the time of apply / reapply methods being called. 

I've also added some comment on that, to explain what are we doing and why. Let me know If comments are too much verbose and / or you prefer to strip them out.

Thank you :)

Patane97 (approved):
This seems good enough ~ I primarily say this as I want to (eventually) re-visit the `RevealCards` GameAction, as it was quite messily done by me in a rush to get it released in time. When I get to that (which should include actual proper unit tests for it), I'll revisit this effect as well.

The comments aren't always necessary, **but** they're helpful, and I am never against having comments explain code, so welcome to keep it there! :)

**The only thing I'd maybe recommend adding before we merge** would be adding someone like Sweetrobin to the WotE test to ensure that RevealCards gameaction is firing correctly.

Let me know if you intend to do this, and I'll hold back on the merge. :)

LaranSoft:
Good :)

No problem, I'll provide a new use case in the WotE test file, to address the interaction with Sweetrobin ;)

Patane97:
> There was a problem with Sweetrobin and "A Very Large Shadow" and I'm not sure about the correctness of the fix. Sweetrobin condition checks if event printed cost is less that, or equal to, the initiative value of the owner. The event has a printed cost of "-" which means that, for Sweetrobin interrupt purposes, it is 0 and so it is less than, or equal to, the initiative value of the plot card.

This is incorrect; "-" is **not** a printed cost. I believe you're thinking of X. See here for reference (is also in Sweetrobins' FAQ on ThronesDB):
> _**Rules Reference > Shadows > Printed cost of shadows cards**_
> _Some cards with the shadow keyword have a "-" in place of their printed cost. Such a card cannot be marshaled or played from a player's hand, and may only be played by being brought out of shadows. Such a card has no printed cost. If an ability references the printed cost of that card, a null value is returned._

server/game/cards/23-AHaH/Sweetrobin.js
```diff
@@ -5,7 +5,7 @@ class Sweetrobin extends DrawCard {
     setupCardAbilities(ability) {
         this.interrupt({
             when: {
-                onCardRevealed: event => event.card.hasPrintedCost() && event.card.getPrintedCost() <= event.card.owner.getTotalInitiative() && ['hand', 'draw deck', 'shadows'].includes(event.card.location)
+                onCardRevealed: event => (!event.card.hasPrintedCost() || (event.card.hasPrintedCost() && event.card.getPrintedCost() <= event.card.owner.getTotalInitiative())) && ['hand', 'draw deck', 'shadows'].includes(event.card.location)
```
Patane97, line 8: As per my comment [here](https://github.com/throneteki/throneteki/pull/3372#issuecomment-1588256333), this needs to be reverted. May need to adjust your WotE test to compensate in some way. :)

test/server/cards/23-AHaH/WardensOfTheEast.spec.js
```diff
@@ -0,0 +1,83 @@
...
+                    });
+
+                    it('should correctly satisfy reaction condition of sweetrobin', function () {
+                        expect(this.player2.currentPrompt().menuTitle).toBe('Any interrupts?');
+                    });
+
+                    describe('when sweetrobin interrupt is triggered', function () {
+                        beforeEach(function () {
+                            this.player2.clickCard(this.sweetRobin);
+                        })
```
Patane97, line 73: Just needs to pass the eslint test:
```suggestion
                    describe('when sweetrobin interrupt is triggered', function () {
                        beforeEach(function () {
                            this.player2.clickCard(this.sweetRobin);
                        });
```

Patane97 (changes requested):
Thought it best to re-review as changes are required. :)

server/game/cards/23-AHaH/Sweetrobin.js
LaranSoft: omg, totally wrong with that :D You're right, I thought that "X" cost and "-" cost were the same for the purpose of determining the value. Reverted change

test/server/cards/23-AHaH/WardensOfTheEast.spec.js
LaranSoft: Sure, fixed

Patane97 (approved):
Looking good; thought I approved this a while ago haha.

server/game/effects.js
```diff
@@ -1451,31 +1451,47 @@ const Effects = {
...
 
-                context.game.resolveGameAction(GameActions.revealCards({
-                    cards: shadows,
-                    player,
-                    revealWithMessage: false,
-                    highlight: false,
-                    source: context.source
-                }), context);
+                // resolving the 'reveal' game action for cards in shadows (if there are any)
+                if(shadows.length > 0) {
```
ystros, line 1473: This `if` is actually unnecessary, since the `allow` method for `RevealCard` action returns false for empty arrays. But it's consistent with the rest of this effect, so 🤷‍♀️

Patane97: Completely forgot that the `resolveGameAction` method passes through `action.allow(props)` before actually resolving, good catch! I might push a tweak through to remove it *at some point* :P

PR #3373, opened 2023-06-08
investigation on blood of the dragon and house manderly knight wrong …

LaranSoft:
... interaction

As pointed out by deadmeat, there is a wrong interaction between House Manderly Knight and Blood of the Dragon plot.
I wroted a test case to reproduce it and investigated on the reason.

The problem is that Blood of the Dragon doesn't have an explicit "condition" in the definition of the persistent effect, in opposition to House Manderly Knight. 
This means that the effect of the Plot is not considered to be "stateDependent", while the effect on the Knight is.

This in turn cause the `reapplyStateDependentEffects` of the `effectengine` to recalculate the effect of the Knight (when the Winter plot cease to be revaled) and the condition on knight cause the effect to be cancelled (`reapply` method in `effect.js`).
As a consequence, the strength of Knight is reduced by 2 and an event is raised (`modifyStrength` method in `drawcard.js`). That event kills the Knight.

By making the Plot also conditional, his effect is recalculated (and so cancelled) in the same event handler that is cancelling the effect of the Knight. As a result, when the `onCardStrengthChanged` event is processed by the game, the strenght of the Knight has switched back to 1.

The change proposed by me solves the problem, but still I think is not so correct. I mean, the condition for a plot to be the revealed plot for a persistent effect to be active should be a default, without the need of an explicit definition (unless stated otherwise). Maybe the method `persistentEffect` of the class `PlotCard` should be overridden to add the condition (if not present) to the properties? Any other thoughts?

Thank you :)

### DIFF
--- a/server/game/cards/01-Core/TheWall.js
+++ b/server/game/cards/01-Core/TheWall.js
@@ -22,6 +22,7 @@ class TheWall extends DrawCard {
             }
         });
         this.persistentEffect({
+            condition: () => this.location === 'play area',
             match: (card) => card.isFaction('thenightswatch') && card.getType() === 'character',
             effect: ability.effects.modifyStrength(1)
         });
--- a/server/game/cards/02.4-NMG/BloodOfTheDragon.js
+++ b/server/game/cards/02.4-NMG/BloodOfTheDragon.js
@@ -3,6 +3,7 @@ const PlotCard = require('../../plotcard.js');
 class BloodOfTheDragon extends PlotCard {
     setupCardAbilities(ability) {
         this.persistentEffect({
+            condition: () => this.game.getPlayers().some(player => player.activePlot && player.activePlot === this),
             match: card => card.getType() === 'character' && !card.hasTrait('Dragon'),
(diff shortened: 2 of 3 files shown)


LaranSoft:
to elaborate further, I add a condition also to "The Wall" persistent effect (it was causing a fail in burneffects test file).

All changes I made are not correct, there is for sure another way to achieve the same results without the need to write all explicit conditions. I'm not even sure about the correctness of adding a condition to a persistent effect when the printed card has no text that refers to an explicit condition.

ystros:
If this is an engine problem, it should be fixed as part of the engine, not individual cards. Adding conditions to individual persistent effects that check the card is in the right place (which the engine _already_ should be doing) is not a sustainable solution.

Patane97:
@LaranSoft I recommend, whilst we look through potential engine changes, that we convert this into a **Draft** Pull Request, if you are able to do that. Let me know if you cant :)

Patane97:
Will be closing this PR for staleness. Can re-open if anything changes there!

PR #3377, opened 2023-06-10
change the delay to start the chess clock so that it is a game option instead of fixed 5 seconds

### LINKED ISSUE #2952: Feature Request for the upcoming World Cup
I was wondering if it's possible to get a chess clock option for the timer so that games can be self policed when it comes to slow play etc. Chess clocks count down when it's your priority (action window) and you lose if your clock hits 0. This eliminates any enmity towards players for playing certain strategies and eliminates any chance of slow playing. There's a functional version that works for L5R and we use it for that World Cup.

### LINKED ISSUE #3038: Chess clocks cause the play area to shake
Using chess clocks in a game will cause the screen to "shake" left and right when the chess clock icon changes from one player to another.
The issue is caused by the XX:XX H1 text combined with the chess clock icon being wider than the plot area part of the screen. I believe the limit is 101px and the chess clocks are a couple of pixels wider.

### LINKED ISSUE #3368: Add option to set Chess Clock "buffer" amount
Currently the Chess Clock has a 5 second buffer which gives some lee-way between certain action windows. For example:
image

The GOT are looking to allow this value to be adjusted for any game or event, with it being defaulted to 5 seconds; if set to 0 seconds, it should essentially disable this feature.

Note that this would require a value input for both the new game screen, as well as the event customisation screen.

EvadLhorg:
also take a look at the client PR here: https://github.com/throneteki/throneteki-client/pull/189

fixes #3368
fixes #3038 (client PR)
fixes #2952

### DIFF
--- a/server/game/ChessClock.js
+++ b/server/game/ChessClock.js
@@ -63,7 +63,7 @@ class ChessClock {
     }
 
     timeRanOut() {
-        this.player.game.addMessage('{0}\'s clock has run out', this.player);
+        this.player.game.addAlert('warning', '{0}\'s clock has run out', this.player);
         //TODO make this melee friendly
         this.player.game.recordWinner(this.player.game.getOpponents(this.player)[0], 'time');
         return;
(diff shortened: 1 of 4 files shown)


Patane97 (approved):
Looks great to me (along with the Client side after second review) 💪

PR #3379, opened 2023-06-14
Fix hero

LaranSoft:
Ok, this is huge :) I'll convert this to draft because it is an intervention on the engine made by one of the newcomers, needs a deep review. 

I'll try to explain at my best what I did and why.

Lets start with `baseAbility`. Each `baseAbility` has a `gameAction` which is constructed through method `buildGameAction`. In this method there is a guard which throws `'Cannot use gameAction with abilities with choices'`. This guard is essential because, without it, targeted abilities wouldn't be registered to any `AbilityWindow`. Why? Because when an ability is registered to a window (thorugh method `registerAbility` of `BaseAbilityWindow`) a check is made if ability `canResolve`, which means (method `canResolve` of `baseAbility`) it `meetsRequirements`, `canResolvePlayer`, other checks, and finally `this.gameAction.allow`. Here is the problem: at the time this check is made, no target has been yet chosen and resolved. So, the `gameAction.allow` method will always throw an exception if the ability requires targets, because the context won't have the target resolution. 
Take for example the `StandCard` game action: the `canChangeGameState` checks need to checks card `location` and `kneeled` flag, but if no `card` is provided a `Cannot read properties of undefined` TypeError is thrown.
So, the guard in `buildGameAction` method ensures that abilities which requires targets will be constructed with a `HandlerGameActionWrapper`, which overrides the `allow` method to always be true. 

This is, in my opinion, the conceptual problem: if an ability requires targets, it will always be allowed, even when the ability won't produce any state change. This is the reason why `Aggo` (for example) needs to check in the `cardCondition` of its `target` if the card is in the play area, even though, conceptually speaking, its effect (`handler`) is to stand the target, and this can only be done for cards in play area. And there is also another problem with `Aggo`: if there is only one Bloodrider character in play area which is already standing, `Aggo`'s Action would be available for trigger, even though it should not (because it wouldn't produce any state change).

In my proposal, I removed the guard in the method `buildGameAction`: now targeted abilities can be constructed with an explicit GameAction given in the properties. To address the problem explained up, I changed also the `canResolve` method in `baseAbility`, defining a new check `canAlterGameState` in substitution of the previous `gameAction.allow` check. If there are no targets then we simply check `gameAction.allow` like before - not sure if the temporary context is needed though, for now I used it.
If however the ability has one or more targets, then, for each target, we collect the potential cards that could be chosen for that target, and then check, for each of those cards, if the `gameAction` allow a stateChange if that card is chosen for that target. 
Notice that `canAlterGameState` checks is made after the `canResolveTargets` check, so we can assume each target can be satisfied. This means that if a target is optional / isAble / etc... and have 0 cards eligible for choice, it is still allowed even if, from a rules perspective, maybe it shouldn't (checking). The interested condition is `eligibleCards.length <= 0` in the `canAlterGameStateForSome` method definition. Without that condition, some tests fail. Investigating on this.

With this change, now Hero can be fixed by defining directly the `gameAction` instead of wrapping it inside an handler. This is a more elegant solution in my opinion, because there is less boiler plate code. I wrote three test case for Hero:
a) checking if Hero Reaction is prompted when there is a kneeling army in play
b) checking if Hero Reaction is not prompted when there is a standing army in play
c) checking if Hero Reaction is not prompted when there are no armies in play

Discuss :D

### DIFF
--- a/server/game/AbilityTarget.js
+++ b/server/game/AbilityTarget.js
@@ -39,7 +39,7 @@ class AbilityTarget {
         }) && this.subTargets.every(subTarget => subTarget.canResolve(context));
     }
 
-    getEligibleCards(context){
+    getEligibleCards(context) {
         const selector = CardSelector.for({ context, ...this.properties });
 
         return this.getChoosingPlayers(context).reduce((targets, choosingPlayer) => {
--- a/server/game/AbilityTarget.js
+++ b/server/game/AbilityTarget.js
@@ -39,6 +39,17 @@ class AbilityTarget {
         }) && this.subTargets.every(subTarget => subTarget.canResolve(context));
     }
 
+    getEligibleCards(context){
+        const selector = CardSelector.for({ context, ...this.properties });
+
+        return this.getChoosingPlayers(context).reduce((targets, choosingPlayer) => {
+            context.choosingPlayer = choosingPlayer;
+            return targets.concat(selector.getEligibleTargets(context)).concat(this.subTargets.reduce((targets, subTarget) => {
+                return subTarget.getEligibleTargets(context);
+            }, []));
+        }, []);
+    }
+
     buildPlayerSelection(context) {
         // Creating the selector once the target is being selected for effects such as keywords with a changing target amount
         this.selector = CardSelector.for({ context, ...this.properties });
--- a/server/game/baseability.js
+++ b/server/game/baseability.js
@@ -63,10 +63,6 @@ class BaseAbility {
 
     buildGameAction(properties) {
         if(properties.gameAction) {
-            if(properties.target || properties.targets || properties.chooseOpponent || properties.choosePlayer) {
-                throw new Error('Cannot use gameAction with abilities with choices');
-            }
-
             return properties.gameAction;
         }
 
@@ -87,7 +83,7 @@ class BaseAbility {
             this.canResolvePlayer(context) &&
             this.canPayCosts(context) &&
             this.canResolveTargets(context) &&
-            this.gameAction.allow(context)
+            this.canAlterGameState(context)
         );
     }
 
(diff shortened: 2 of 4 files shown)


Patane97:
Hmm, a few things here ~ I will try to summarise:
- Firstly, assuming it was unintentional, but you never actually mentioned the *issue* with Hero here. In a general sense, I would recommend first creating an `issue` within this repository if you've found something, and (if appropriate) linking to any sources of the reported bug, such as a discord message or your own game experience. This would be to primarily save time before you put in work to fix something, as others such as myself or @ystros would likely have some useful insight into the situation, especially if a major engine adjustment is a potential solution. I know you've done this by making this a draft PR, but I'd recommend an issue first as we could likely save you lots of time investigating & coding. :)
- I am assuming based on the investigation/changes that this is in regard to Hero highlighting all ***Army*** characters in the game. A fairly simple fix of limiting the target range to only target cards in-play, for now, is more appropriate. General rule of thumb is that if this sort of ability is working on similar cards (such as Windblown), then a deep engine adjustment is likely not the right path.
- Now, I mentioned above that its likely best to get our insight on something like this beforehand; with this scenario, it's a "issue" we've been well aware of for quite some time now ~ it's a feature gap with `GameActions`. It's a situation which requires much discussion & larger engine adjustments than outlined here, as is a matter of updating **all** current targeting card abilities into `GameActions`, rather than only non-targeting ones ~ this is something that will require various engine-level test files, and changes to hundreds of cards. This is all primarily as it needs to account for passing in all Targetting combinations into the `gameAction.allow` (especially when multiple separate targets or subTargets are chosen), which is a tough task to balance with server performance.

My recommendation here is to apply that simple fix to Hero in a separate branch, and set up a PR for that instead. If you are eager to tackle this major engine change, that's great! It is however something we should discuss well before actual coding changes, possibly within discord with all active developers. :)

Apologies if I missed anything ~ lots of things I've had to read over the past 24 hrs 😄

PR #3382, opened 2023-06-15
fix Hero's reaction can target cards out of play area

### LINKED ISSUE #3380: Hero's reaction can target cards out of play area
As reported by .mak [here](https://discord.com/channels/793188949144764447/853644798706450472/1116491647782506636), Hero's reaction can target all armies, not only those in play area.

LaranSoft:
close #3380 

Just add a condition for `card.location` to be `play area` (like, for example, `Aggo`).

Added also a test case for Hero (failing without the fix, passing with the fix). I don't know if adding a new test case for a bug so simple is correct or not (more test cases means more time to run tests locally). Eventually I will remove it


test/server/cards/24-TSoW/Hero.spec.js
Patane97, line 1: Yeah I would remove this test before merging with live as it's not really necessary, but good idea to make it to ensure it's working as you expect! :)

Will approve, but recommend removing the test file first

Patane97 (approved):
Now this is looking nicer 😄.

Solid work; as mentioned we should just remove this test. :)

LaranSoft:
Perfect! Removed spec file ;) t y

PR #3386, opened 2023-06-16
- fix proposal to Arya Stark duplicates not working sometimes

### LINKED ISSUE #3385: Arya Stark (Core) sometimes can't be saved with the dupe she got on entering
Pointed out by Saraa [here](https://discord.com/channels/793188949144764447/853644798706450472/1118620610701037618).

Steps to reproduce: With a non-winter plot card revealed, marshall Arya Stark (Core) with Nightmare as top deck card. Resolve her reaction so that Nightmare will dupe her.
As the opponent initiate a MIL challenge, use Winterfell (WofN) reaction to prevent each player from triggering card abilities.
Let the opponent win the MIL challenge and choose Arya as claim. The game won't prompt to save her.

LaranSoft:
close #3385 but review needed because it is an engine-level fix.

After investigation, I've found another way to reproduce the problem which doesn't use any trigger restriction.

If Player 1 has `Arya Stark (Core)` with a facedown event as dupe (for example: `Nightmares`) and gold less than the cost of the event (for example: 0 gold), then dupe cannot be used.

That's because during the duplicate ability setup (method `setupDuplicateAbility` of `drawcard`) a new `CardInterrupt` instance is constructed. This class inherits from `TriggeredAbility` which has, in the constructor, a logic for which if the `card.printedType` is `'event'` and the caller doesn't explicitly said to `ignoreEventCosts`, then all three `Costs.playEvent` are added to the cost. As a result, when duplicate ability is setup on `Nightmares`, its costs became 4 in total: `ability.costs.discardDuplicate()` (from `drawCard` `setupDuplicateAbility` method) and three `Costs.playEvent`. If the controller of `Arya` cannot pay one of this 4 costs (for example, because it has no gold or if it has a trigger restriction) then the duplicate interrupt cannot be triggered.

The fix consist in adding the property `ignoreEventCosts: true` when the `setupDuplicateAbility` method is called: could it work?

Tested locally with and without the fix, and it worked in all cases. I didn't add a test case to `Arya Stark` spec file, though. Let me know if it is better to add a test case for this interaction, and if it is correct to add a test case for `Arya` spec or for a more generic spec file, like `Duplicates`.

 Thank you :)

### DIFF
--- a/server/game/drawcard.js
+++ b/server/game/drawcard.js
@@ -71,6 +71,7 @@ class DrawCard extends BaseCard {
             abilitySourceType: 'game',
             canCancel: true,
             cannotBeCanceled: true,
+            ignoreEventCosts: true,
             location: 'duplicate',
             when: {
                 onCharacterKilled: dupeCondition,


Patane97:
This may be something that @ystros will need to give his opinion on. The logic seems to make sense to me in general, but he would be more sure about the potential repercussions of that sort of adjustment. :)

ystros:
Great investigation! While I'd kinda like a test, it doesn't seem critical. Just going to merge as-is.

PR #3390, opened 2023-06-22
fix Drogon's reaction is prompted only if there is at least one killa…

### LINKED ISSUE #3389: Drogon's reaction should be prompted only if there is at least one killable character
As pointed out by .mak [here](https://discord.com/channels/793188949144764447/853644798706450472/1121137562308853911), Drogon's reaction is prompted even if defending player has no characters with STR <= 1 to kill.

LaranSoft:
Close #3389

A simple conversion from handler to gameAction made the job :)

### DIFF
--- a/server/game/cards/15-DotE/Drogon.js
+++ b/server/game/cards/15-DotE/Drogon.js
@@ -12,6 +12,10 @@ class Drogon extends DrawCard {
             when: {
                 afterChallenge: event => event.challenge.winner === this.controller && this.isAttacking()
             },
+            message: {
+                format: '{player} uses {source} to kill all characters {defendingPlayer} controls with STR 1 or lower',
+                args: { defendingPlayer: context => context.event.challenge.defendingPlayer }
+            },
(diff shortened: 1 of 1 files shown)


server/game/cards/15-DotE/Drogon.js
```diff
@@ -11,13 +12,11 @@ class Drogon extends DrawCard {
             when: {
                 afterChallenge: event => event.challenge.winner === this.controller && this.isAttacking()
             },
-            handler: context => {
-                let characters = this.game.currentChallenge.defendingPlayer.filterCardsInPlay(card => card.getType() === 'character' && card.getStrength() <= 1);
-                this.game.killCharacters(characters);
-
-                this.game.addMessage('{0} uses {1} to kill all characters {2} controls with STR 1 or lower',
```
ystros, line 18: Need to convert this to use `message`. Currently, no message will be printed anymore.

ystros (changes requested):
Looks good other than the missing messsage!

Issue #3393, opened 2023-06-27
Wrong interaction between Morghulis and Prized keyword


LaranSoft:
As pointed out by Odrl [here](https://discord.com/channels/793188949144764447/853644798706450472/1122676251589623858), there is a bug in the interaction between Morghulis and Prized keyword.

Easy way to reproduce:
Player A has a character in play with 14 power
Player B has a Prized 1 character in play (for example a >= 6 cost character with Aloof and Apart Agenda)

Player A reveals Valar Morghulis, Player B reveals any plot.

If Player A is chosen as first player, then the game proceed as expected.
Il Player B is chosen as first player though, then Player A wins the game.

This is a bug, because Player A should never win the game in this situation, regardless of the first player choice.

LaranSoft:
Investigated on this one.

The problem is caused by the characters of first player dying before the characters of second player. 

Don't know exactly the correct way to handle this (I imagine two solutions at least).

I will propose a solution that, even if not so elegant in my opinion, actually could solve the problem at hand without twisting too much the engine.

The idea is to change the method `getTotalPower` of `player.js` from

`let cardPower = card.powerOptions.contains('doesNotContribute') ? 0 : card.getPower();`

to 

`let cardPower = card.powerOptions.contains('doesNotContribute') || card.isGoingToDeadPile() ? 0 : card.getPower();`

so that cards that are `goingToDeadPile` (improvable method name :)) cannot contribute to power total.

How to check if a card `isGoingToDeadPile`? First idea was to reuse the existing flag `inDanger` but I think that a better idea is to actually create a new flag to handle this specific condition. For example:

When `promptForDeadPileOrder` method of `KillCharacters` is executed, all cards that are going to die will have this flag switched on, and after the card is actually moved to the dead pile (method `moveCardsToDeadPile` of `KillCharacters`) the flag is switched off.

Tried this solution locally and it actually works in solving the bug.
@ystros, @Patane97 let me know if this can be acceptable: eventually I will open a DPR. 

Thank you :)

Patane97:
I'm not sure I agree with the above solution, primarily as any effect which simultaneously does something (eg. Valar M simultaneously killing each character) should not be checking the "win state" (which uses `getTotalPower`) until that simultaneous effect has completed.

[...]

PR #3396, opened 2023-07-04
Fix valarmorghulis 3393

LaranSoft:
Work in progess #3393 

Refactored `Kill` GameAction to allow multiple kill at once. 
My main goal was to change as little as possible the test suite `Kill.spec.js` and, with that in mind, try to make `Kill` behave like the old `game.killCharacters`. This is the reason why I handled the "force" flag even though I've no idea where it could be used. 
**Note**: `Kill` now accepts an array of "victims" - where each "victim" is an object which specifies the card that is going to be killed, the player killing it, whether the kill `isBurn` and so on.
This way simultaneous kill caused, for example, by different players can be done easily, as well as simultaneous kills with some differences (for example a kill which allow save and a kill which doesn't).
Also note that I didn't changed the "message" method in the `Kill` because I didn't find a way to test it :) not sure where that method is called.

Switched VM to use the new `Kill` and adapted the 4 cards that were already using the new action, to make them work with the new version.

The last two commits (fix: <something>) are there to make all the jasmine tests pass. They are explained in detail in the second half of my [note](https://github.com/throneteki/throneteki/issues/3393#issuecomment-1616954539) - from where I say "_Did some test with this solution and it seems to work_".
Note that I commented out a test case in `Kill.spec.js` because the new `Kill` simply doesn't work anymore that way and I didn't found a way to test specifically that part. 

Finally, they are solving both the "Prized and VM" interaction and the "Alester Florent and Shireen Baratheon" interaction, but I only tested them manually. Maybe they are important enough to merit a test case? Let me know ;)


Patane97:
Haven't got much time to properly review this anytime soon, but unfortunately that `victim` approach is one we can't really take; values within each `GameAction` needs to be consistent & predictable (eg. anything that references a card is always `card`, and multiple cards as `cards`); I don't actually expect the inputs for that Kill gameaction to need to change at all, apart from passing "cards" (as an array) instead of "card" (singular).

The adjustment **should** be extremely simple and small, outlined in https://github.com/throneteki/throneteki/issues/3393#issuecomment-1617066120; `GameActions.Kill.createEvent(...)` should simple return a "onCharacter**s**Killed" event. That event should:
1) Check if cards.length > 1. If it is, create a prompt for the player to choose the order of cards. Then continue with creating the child events in that order. If the length === 1, then just create that singular child event.
2) Map each card to a child event and attaches it to "onCharacter**s**Killed" event. These child events should be named "onCharacterKilled" (as to trigger all existing effects to characters being killed), and will handle the actual killing of that card.

As mentioned, I'd observe closely as to how `GameAction.DrawCards` & `GameAction.RevealCards` works; the changes should be super simple. If you're running into issues with your implementation of this, I'm happy to help within Discord as I can expect little bugs here & there which likely have super simple solutions. :)

Patane97:
Closing, as this has been indirectly resolved by #3436

PR #3410, opened 2023-08-05
fix dothraki steed so the str buff only applies when the challenge is…

### LINKED ISSUE #3405: Highgarden Courtier + Dothraki Steed Bug
If a character has Dothraki Steed attached, they get a STR boost every time they attack, and this boost should create a trigger for the Highgarden Courtier to stand them. However, I have not been prompted with an opportunity to trigger the Courtiers. As attackers are declared and knelt at the same time, I am pretty sure that a reaction window for the Courtier should exist.

EvadLhorg:
… really initiated. this leads to highgarden courtier being able to stand the steed character when the str buff happens

fixes #3405


ystros:
Feels like there's probably a more systemic fix that needs to be made, but we should take this in the mine time. Thanks!

PR #3412, opened 2023-08-05
fix the seed is strong so it actually discards the chosen card

### LINKED ISSUE #3402: Double The Seed is Strong (aHaH) wrong interaction
As pointed out by KaiRong [here](https://discord.com/channels/793188949144764447/853644798706450472/1135312156624163018), a situation happened where the Seed is Strong didn't worked.

EvadLhorg:
fixes #3402

### DIFF
--- a/server/game/cards/23-AHaH/TheSeedIsStrong.js
+++ b/server/game/cards/23-AHaH/TheSeedIsStrong.js
@@ -35,7 +35,10 @@ class TheSeedIsStrong extends DrawCard {
 
     onCardSelected(context, card) {
         this.game.addMessage('{0} discards {1} from their {2}', context.event.winner, card, card.location);
-        context.event.winner.discardCard(card);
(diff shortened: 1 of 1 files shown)


server/game/cards/23-AHaH/TheSeedIsStrong.js
```diff
@@ -35,7 +35,7 @@ class TheSeedIsStrong extends DrawCard {
 
     onCardSelected(context, card) {
         this.game.addMessage('{0} discards {1} from their {2}', context.event.winner, card, card.location);
-        GameActions.discardCard({ card });
```
ystros, line 38: We should probably prefer to use game actions, rather than revert to the methods on the Player class. You should just be able to pass the game action to `this.game.resolveGameAction(...)` here.

ystros (changes requested):

PR #3418, opened 2023-08-31
fix the timing issue with the isAttacking() function for effects that…

### LINKED ISSUE #3191: Randyll Tarly's (Core) reaction not triggering from Dragonglass Dagger
as reported in the BotN Discord, and confirmed by myself.

The setup is Core-Randy, with "Sworn to the Watch" and "Dragonglass Dagger" attached. Randy attacks, the +2 STR buff from the dagger is applied. However, the player does not get a prompt from Randy's reaction to stand him back up.

EvadLhorg:
… should only apply after the challenge is initiated. a challenge is only fully initiated when the attackers are knelt.

this fixes #3191


Patane97 (approved):
I didn't test this myself (as I'm almost certain we would have failed tests if this wasn't configured properly), but my only concern was removed after checking the timings in `InitiateChallenge.js`. When the effects engine reapplies effects right before announcing the initiated message, `challenge.isInitiated` should correctly be "true", which means declared attackers with a STR pump based around `isAttacking()` will correctly be using their pumped STR in that initiation message.

PR #3421, opened 2023-09-06
fix the timelimit so that the timelimit does not end when you pause the game after half the time has expired.

### LINKED ISSUE #3223: /pause-clock bug leading to timer being finished/game being over
when you use the /pause-clock chat command in a timed game, it might bug and show "Time up. The game will end after the current round has finished"
this leads to the game determining a winner after the round and furthermore leads to a timeout of the game session after some minutes

EvadLhorg:
fixes #3223

### DIFF
--- a/server/game/timeLimit.js
+++ b/server/game/timeLimit.js
@@ -45,7 +45,9 @@ class TimeLimit {
     }
 
     checkForTimeLimitReached() {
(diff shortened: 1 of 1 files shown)


Patane97 (approved):
Tested this change, and I can accurately reproduce the error, then confirm it's fixed after this has been applied. Nice work 💪

PR #3445, opened 2024-01-18
Fixing PutIntoPlay to include 'Duplicate' location as being fully resolved

Patane97:
Noticed with Ser Mark Mullendure, where a duplicate is put into play. Currently it does not continue the "Then" of his ability, to put him on top of the deck, when ruling should count that as being resolved.

This should fix this scenario, ~~as well as any other put into play scenarios that **should** count duplicating as a successful action.~~


ystros:
While this will fix Mullendore (since it says put a "card" into play), it will break any card that says to put a "character / location / attachment" in play that is part of a pre-then condition. This includes:
* [Arianne Martell (Core)](https://thronesdb.com/card/01104)
* [Old Wyk](https://thronesdb.com/card/05028)
* [Silence](https://thronesdb.com/card/12016)
* [Heir to the Iron Throne](https://thronesdb.com/card/12048)
* [Flea Bottom Alley](https://thronesdb.com/card/13006)
* [Pentos](https://thronesdb.com/card/15018)
* [I Pledge My Life and Honor](https://thronesdb.com/card/20024)
* [As High as Honor](https://thronesdb.com/card/20049)

We probably need a way to distinguish between put in plays that require a card of a certain kind come out vs ones that allow a dupe to come out and be considered resolved.

Patane97:
> While this will fix Mullendore (since it says put a "card" into play), it will break any card that says to put a "character / location / attachment" in play that is part of a pre-then condition. This includes:
> 
> * [Arianne Martell (Core)](https://thronesdb.com/card/01104)
> * [Old Wyk](https://thronesdb.com/card/05028)
> * [Silence](https://thronesdb.com/card/12016)
> * [Heir to the Iron Throne](https://thronesdb.com/card/12048)
> * [Flea Bottom Alley](https://thronesdb.com/card/13006)
> * [Pentos](https://thronesdb.com/card/15018)
> * [I Pledge My Life and Honor](https://thronesdb.com/card/20024)
> * [As High as Honor](https://thronesdb.com/card/20049)
> 
> We probably need a way to distinguish between put in plays that require a card of a certain kind come out vs ones that allow a dupe to come out and be considered resolved.

Good catch.
My initial (and simplest) thought is a defaulted field to give `PutIntoPlay`, maybe called "dupeIsValid", which allows you to determine if a duplicate should still be considered as fully resolvable. This would be defaulted to false, and in special cases like Mark (where it's referred to as "Card" and not a specific card type), we can set `dupeIsValid = true`, and add that logic appropriately into that `isFullyResolved` that I edited.

I'm not sure if a more generalised/better solution is out there, or if it's even worth investing time into.

Thoughts @ystros?

Patane97:
For now I have updated this to the `dupeIsValid` idea above as it's the simplest, but am happy to discuss a different avenue. :)

ystros:
Yeah, this seems like the simplest way to do it for now. Can definitely do exact types (or an array of allowed types, where "dupe" is some sort of extra type) if the need arises in the future.

PR #3454, opened 2024-03-10
Fixing A Cask of Ale

sarabertoncelj:
Changed the event so you first have to select the two cards from the same controller and then which card you are moving the power from. You are apparently also supposed to be able to select two cards without power if you for some reason want the event to not resolve, because it is a referential target, but you can't play the event if there is no power in play.

### DIFF
--- a/server/game/cards/14-FotS/ACaskOfAle.js
+++ b/server/game/cards/14-FotS/ACaskOfAle.js
@@ -6,7 +6,7 @@ class ACaskOfAle extends DrawCard {
         this.reaction({
             when: {
                 afterChallenge: event => (event.challenge.isMatch({ winner: this.controller, challengeType: 'power' }) 
-                    && this.game.allCards.some(card => (card.location === 'active plot', 'faction', 'play area' && card.getPower()>0)))
+                    && this.game.allCards.some(card => (card.location === 'active plot', 'faction', 'play area' && card.getPower() > 0)))
             },
             target: {
                 mode: 'exactly',
@@ -57,11 +57,11 @@ class ACaskOfAle extends DrawCard {
         this.game.addMessage('{0} plays {1} to move {2} power from {3} to {4}', this.context.player, this, amount, this.fromCard, this.toCard);
 
         this.game.resolveGameAction(
-            GameActions.movePower(context => ({
+            GameActions.movePower({
                 from: this.fromCard,
                 to: this.toCard,
                 amount
-            })),
+            }),
             this.context
         );
 
--- a/server/game/cards/14-FotS/ACaskOfAle.js
+++ b/server/game/cards/14-FotS/ACaskOfAle.js
@@ -5,55 +5,70 @@ class ACaskOfAle extends DrawCard {
     setupCardAbilities() {
         this.reaction({
             when: {
-                afterChallenge: event => event.challenge.isMatch({ winner: this.controller, challengeType: 'power' })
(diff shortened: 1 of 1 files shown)


server/game/cards/14-FotS/ACaskOfAle.js
```diff
@@ -5,56 +5,71 @@ class ACaskOfAle extends DrawCard {
     setupCardAbilities() {
         this.reaction({
             when: {
-                afterChallenge: event => event.challenge.isMatch({ winner: this.controller, challengeType: 'power' })
+                afterChallenge: event => (event.challenge.isMatch({ winner: this.controller, challengeType: 'power' }) 
+                    && this.game.allCards.some(card => (card.location === 'active plot', 'faction', 'play area' && card.getPower()>0)))
```
ystros, line 9: The location check is not quite valid syntax. I think you want something more like `['active plot', 'faction', 'play area'].includes(card.location)` instead (similar to what you did below)

server/game/cards/14-FotS/ACaskOfAle.js
```diff
@@ -5,56 +5,71 @@ class ACaskOfAle extends DrawCard {
...
-                    cardType: ['attachment', 'character', 'faction', 'location', 'plot']
-                }
+            target: {
+                mode: 'exactly',
+                numCards: 2,
+                activePromptTitle: 'Select 2 cards',
+                singleController: true,
+                cardCondition: card => ['active plot', 'faction', 'play area'].includes(card.location),
+                cardType: ['attachment', 'character', 'faction', 'location', 'plot'],
+                gameAction: 'movePower'
```
ystros, line 18: This doesn't guarantee that cards chosen are from a player that has power. Is that necessary, or just that power is somewhere on the board?

ystros (changes requested):
Thanks for the PR. Beyond my comments below, I'd ask whether this change is something that can affect a typical game (e.g. some interaction with other cards), or if this is a change to make the card "rules accurate" simply for the sake of being rules accurate.

> You are apparently also supposed to be able to select two cards without power if you for some reason want the event to not resolve

e.g. why would someone want to do this?

server/game/cards/14-FotS/ACaskOfAle.js
sarabertoncelj: I asked Miha (Odrl) about it and you should be able to play it as long as there is power in play (I guess you could play it for some trigger like Tom or DWDW and then let the effect fizzle). The problem before was that it would allow you to select the cards from the opponent and yours, because the first card was read as undefined when selecting the second one. I think there is a similar problem with Richard Horpe that lets you move power back to the same card.

ystros:
Thanks!

Patane97:
@ystros @sarabertoncelj No issues with this being merged, but I would be curious as to why this was even allowing a player to choose cards controlled by different players; the condition of the second target ("to") was meant to check this - did it never work?
```js
(!context.targets.from || card !== context.targets.from && card.controller === context.targets.from.controller)
```

sarabertoncelj:
I think the condition of the second target doesn't work when it references the first target. It also doesn't work in Horpe, he has this condition for the second to target:
`(!context.targets.from || card !== context.targets.from)`
And it lets you select the same card.

With Cask, if I removed the first condition before the OR I would get cannot read properties of undefined when reading controller.

Issue #3440, opened 2023-11-15
Forgotten By History and The Crone wrong interaction


LaranSoft:
Forgotten by History working on a 1-strenght character during The Crone turn. 
The Crone effect is a "can't" effect, and so it should be absolute.

PR #3461, opened 2024-03-29
Crone Forgotten By History Interaction

sarabertoncelj:
Closes: #3440
I added the condition when selecting targets, this way you also can't play the event if there is no valid target. I also added this condition in Bloody Cup, because it now let's you select invalid targets on Crone turn. Not sure if this is okay solution for this, but if it is I can add it on some other events that I know also work like that.


ystros:
This is generally the right idea. A more complete implementation would be to call the `allow` method for the corresponding GameAction class, which can check other conditions (e.g. whether the card is allowed to leave play). Ideally neither will be necessary in the future, once targeting + game actions are better integrated.

I had started a branch for this Forgotten History issue, I can address the remaining cards there.

Issue #3452, opened 2024-02-26
Interaction Sweet Robin and Ruby of R'hllor not implemented


yakushika:
In TiT Sweet Robin does not get highlighted (clickable) after using the reaction of Ruby of R'hllor (assuming the opponent has at least one card with cost <= ini of opponent). The Hand Cards are only revealed in the protocoll textbox and not clickable (like Ghost of High Heart, Red Priest etc.).

PR #3464, opened 2024-04-04
Ruby of Rhllor and Nighttime Marauders Reveal

sarabertoncelj:
Closes: #3452

Added revealCards action so Sweetrobin can interrupt. Also added for Nighttime Marauders which had similar problem.

### DIFF
--- a/server/game/cards/04.1-AtSK/RubyOfRhllor.js
+++ b/server/game/cards/04.1-AtSK/RubyOfRhllor.js
@@ -1,4 +1,5 @@
 const DrawCard = require('../../drawcard.js');
+const GameActions = require('../../GameActions');
 
 class RubyOfRhllor extends DrawCard {
     setupCardAbilities() {
@@ -15,7 +16,17 @@ class RubyOfRhllor extends DrawCard {
             handler: context => {
                 this.game.promptForCardName({
                     player: context.player,
-                    onSelect: (player, cardName) => this.selectCardName(player, cardName),
+                    onSelect: (player, cardName) => {
+                        this.game.resolveGameAction(
+                            GameActions.revealCards(context => ({
+                                cards: [...context.game.currentChallenge.loser.hand],
+                                player: context.player,
+                                revealWithMessage: false
+                            }))
+                            , context);
+                        this.selectCardName(player, cardName);
+                        return true;
+                    },
                     source: context.source
                 });
             }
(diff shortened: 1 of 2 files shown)


server/game/cards/04.1-AtSK/RubyOfRhllor.js
```diff
@@ -15,7 +16,17 @@ class RubyOfRhllor extends DrawCard {
...
+                        this.game.resolveGameAction(
+                            GameActions.revealCards(context => ({
+                                cards: [...context.game.currentChallenge.loser.hand],
+                                player: context.player,
+                                revealWithMessage: false
+                            }))
+                            , context);
+                        this.selectCardName(player, cardName);
+                        return true;
+                    },
```
Patane97, line 29: Good fix - although I can see a small scenario where the card(s) would be discarded (within `selectCardName`) prior to the revealing their hand, and it causes a problem. If the named card is Alla Tyrell, her Interrupt should allow her to enter play prior to the reveal or discard's actually taking affect, which would prevent her from being discarded via this ability (as she's moved from the hand to in-play).

If you wanted to implement for that, it would likely need to be a combination of a `GameActions.Simultaneously` triggering both `GameActions.revealCards` & `GameActions.IfCondition`, with the if condition checking if any _revealed_ cards in that hand match, and if so, triggers a `GameActions.DiscardCard` on those cards. Happy to discuss privately more about how these GameActions work if you'd like!

Otherwise simply commenting a TODO here is also fine to just recognise that this will miss a specific scenario, and it can maybe be cleaned up later. :)

server/game/cards/17.1-R/NighttimeMarauders.js
```diff
@@ -15,6 +15,13 @@ class NighttimeMarauders extends DrawCard {
             handler: context => {
                 let opponent = context.target.controller;
                 let cardsToDiscard = opponent.hand.filter(card => card.hasPrintedCost() && card.getPrintedCost() === context.target.getPrintedCost());
+                this.game.resolveGameAction(
+                            GameActions.revealCards(context => ({
+                                cards: [...opponent.hand],
+                                player: context.player,
+                                revealWithMessage: false
+                            }))
+                            , context);
```
Patane97, line 24: Similar comment to the one in RubyOfRhllor. :)

Patane97 (changes requested):
Nice fixes - just two (identical) comments to address; happy to chat/explain anything you need.

Additionally, would be good to also apply these changes to the non-redesigned `NighttimeMarauders.js`, and `SeptonCellador.js`. :)

PR #3468, opened 2024-04-12
Fixing handleGameState

Patane97:
Fixes THRONETEKI-37K
[Sentry Issue](https://throneteki.sentry.io/issues/5172330382/?project=123019&referrer=slack)

I could not pin-point the exact scenarios that is causing `currentGame` to be null, however regardless of the actual cause, this statement **should** be nullchecked, purely by logic within `handleGameState` itself:
```js
    if(!currentState || !currentState.players[username] || currentState.players[username].left) {
        delete retState.currentGame;
        retState.newGame = false;
    }
```
If this happens, the offending if statement will always error - this aims to resolve this scenario and, hopefully any others which I could not pin-point the reason behind.


cryogen:
The reason this breaks is because this line deletes the currentGame:

Patane97:
> The reason this breaks is because these lines deletes the currentGame:
> 
> ```
> if(!currentState || !currentState.players[username] || currentState.players[username].left) {
>     delete retState.currentGame;
>     retState.newGame = false;
> }
> ```
> 
> the offending code should be made an else on this if and it should solve the issue

On second thought, I actually want to confirm if this is correct - wouldn't this also need to check if the game has not started (eg. `!currentState.started`)?

I am considering doing the following something _like_ the following:
```
    if(!currentState || !currentState.players[username] || currentState.players[username].left) {
        delete retState.currentGame;
        retState.newGame = false;
    } else if(currentState && !currentState.started) {
        retState.newGame = true;
    }
```

cryogen:
Yes that's exactly what I meant. Sorry if I wasn't clear :)

Patane97:
@cryogen Should be good now with 84a103e - happy for you to merge when ready. :)

PR #3472, opened 2024-04-22
Updating Bribery to use GameActions

Patane97:
Was resolving an issue I saw somewhere, although cannot pinpoint where 😅.

### DIFF
--- a/server/game/cards/15-DotE/Bribery.js
+++ b/server/game/cards/15-DotE/Bribery.js
@@ -11,33 +11,25 @@ class Bribery extends DrawCard {
                     (!context.xValue || card.getPrintedCost() <= context.xValue) &&
                     (
                         card.isMatch({ kneeled: false }) && card.allowGameAction('kneel') ||
-                        card.isMatch({ trait: ['Ally', 'Mercenary'] })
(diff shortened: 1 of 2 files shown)


server/game/cards/15-DotE/Bribery.js
```diff
@@ -11,33 +11,25 @@ class Bribery extends DrawCard {
...
-                    });
-                } else {
-                    this[buttons[0].method](context.player);
-                }
+                        elseAction: {
+                            message: '{player} players {source} to kneel {target}',
+                            gameAction: GameActions.kneelCard(context => ({ card: context.target }))
+                        }
+                    }),
+                    context
```
ystros, line 31: I'm not sure how much it matters (since this is the original version of the card, not the redesign), but this now forces you to take control if you can, whereas the card text gives you a choice ('you may take control instead')

If we don't care, then this should be good to merge.

Patane97: @ystros Should be fixed now in 1e58741. There might be some small cases where a kneeling Ally/Mercenary will still give you a choice to kneel it when it shouldn't, but I think we can leave it (as that seems more like something we should fix with `GameActions.choose` rather than this card specifically. Maybe an option within choose to eliminate options if the condition or allow isn't met (where currently, if 1 is available, all should be available).

Can merge this if youre happy! :)

PR #3486, opened 2024-05-19
update throneteki-json-data to latest ref for the new restricted list

EvadLhorg:


EvadLhorg:
did something wrong, will open up new pr

PR #3490, opened 2024-05-23
Aggregate events rework

Patane97:
Ultimately the aim of this update is to improve our aggregate scenario to be more versatile, more efficient, and more readable.

~~The format for aggregated events within the card files themselves is still being improved (I feel like there is a cleaner, more readable way I can do this, possibly even in a single function), but for now it's a "proof of concept" to prove that it works.~~

Todo:
- [x] *Maybe* improve how aggregates are expressed on card files themselves (eg. Cersei Lannister)
Improved into following structure:
```javascript
this.reaction({
    when: {
        onCardDiscarded: {
            // Aggregate by the controller of the discarded cards, and the location they were discarded in
            aggregateBy: (event, context) => ({
                controller: event.cardStateWhenDiscarded.controller,
                location: event.cardStateWhenDiscarded.location
            }),
            // Check if this ability should trigger using the aggregate details defined above
            condition: (aggregate, events, context) =>
                aggregate.controller !== this.controller &&
                ['hand', 'draw deck'].includes(aggregate.location)
        }
    },
    handler: context => {
        // Draw context.events.length worth of cards
    }
});
```
- [x] Create at least 1 additional test (either for an aggregate card or, even better, one to test aggregate functions on their own)
Added to `forcedcardreaction` and `cardreaction` tests to also catch aggregates.
- [x] Add to `implement-cards.md`!!
Added above structure to implement-cards.md
- [x] Although tests are passing, I might be approaching aggregates incorrectly; noticed after trying to update `Tywin Lannister (LoCR)` to use an aggregate. It might not actually be aggregating/grouping as I expect.
- [x] Do a personal review, and likely ensure all tests are being performed properly.

*Close the following PR in playtesting once merged: https://github.com/throneteki-playtesting/throneteki/pull/269*


server/game/cards/05-LoCR/CerseiLannister.js
```diff
@@ -8,22 +9,20 @@ class CerseiLannister extends DrawCard {
...
+                onCardDiscarded: {
+                    aggregateBy: (event) => ({
+                        controller: event.cardStateWhenDiscarded.controller,
+                        location: event.cardStateWhenDiscarded.location
+                    }),
+                    condition: (aggregate, events, context) =>
+                        aggregate.controller !== this.controller &&
+                        aggregate.location === 'hand' &&
+                        !context.player.isSupporter(aggregate.controller)
+                }
```
ystros, line 21: This system seems super flexible, I'm just wondering if there are ways to make it read better. Maybe simplifying the aggregateBy in some way (pre-made functions, less reliance on `cardStateWhenXed`)

server/game/triggeredability.js
```diff
@@ -53,23 +53,32 @@ class TriggeredAbility extends BaseAbility {
...
+        if (typeof listener === 'function') {
+            context.isTriggered = listener(event, context);
+            return context;
+        } else {
+            const aggregates = this.getAggregatedEvents(event, listener, context);
+            // Concat all aggregates which have passed the condition (there could be multiple)
+            context.events = [...aggregates.map((a) => a.events)];
+            context.isTriggered = context.events.length > 0;
+            return context;
+        }
```
ystros, line 73: I'm wondering if there's a way to get rid of the `isTriggered` variable here. It doesn't feel like it should live on the `context` object, since it's only used in this class. It also feels slightly weird that you need to create the full context object before you can determine whether the event can be triggered.

It might be nice to not have this vague `if`, and instead have classes (or functions) encapsulating this logic at the time `TriggeredAbility` is constructed.

Patane97: Yeah so I actually realised that my implementation above _doesn't actually aggregate properly_; I could go into it but I'd be writing a full paragraph - I mostly noticed when I tried to convert Tywin (LoCR) to use aggregates and the whole system crumbled, so I switched this to a Draft and am trying to re-do it.

The next attempt is more doing what you recommend, where I'm making something like an `AggregateCondition` and `SingularCondition` for what events can trigger an ability. I'll send you a message if I have any issues, or when it's ready to review that update. :)

Totally agree with you on removing this `isTriggered` though - felt a little clunky!

server/game/cards/05-LoCR/CerseiLannister.js
Patane97: @ystros have updated this to be a bit more simple, with just providing a map of named properties. They are they accessible through `context.aggregate.{property}`

Patane97 (commented):
Self-review looks good; did some minor cleanup.

ystros (commented):
Looking better overall. I still wish there were helper functions for the `aggregateBy` definition (e.g. a premade `aggregateBy: cardTypeAndController` instead of redefining it in each card), but I think the need to use `cardStateWhen...` hinders that, and there are so few cards that need aggregates that it isn't a big deal.

server/game/triggeredability.js
```diff
@@ -218,4 +204,137 @@ class TriggeredAbility extends BaseAbility {
     }
 }
 
+export class SingularTrigger {
```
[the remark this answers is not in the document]
Patane97: Fair suggestion; moved in c1e8715.

server/game/cards/05-LoCR/TywinLannister.js
```diff
@@ -4,17 +4,22 @@ class TywinLannister extends DrawCard {
     setupCardAbilities() {
         this.interrupt({
             when: {
-                onCardDiscarded: (event) => this.exactlyOneCardDiscardedFromPlayersDeck(event)
+                onCardDiscarded: {
+                    aggregateBy: (event) => ({
+                        controller: event.card.controller,
```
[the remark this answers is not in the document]
Patane97: I believe so yeah.

Ideally I'd love to clean up when "cardStateWhenX" should be used & when "card" is to be used, as it can be confusing for Reactions vs Interrupts. Maybe a future update!

Issue #3526, opened 2024-08-14
Queenscrown should always put the cards on the bottom of the deck


Chrisssie:
Hello team,

During a streamed match last night we found out the [Queenscrown](https://thronesdb.com/card/07019) should always place the cards on the bottom of the deck not just when there is a character in the 3 cards it finds. Also when a character is not chosen all cards should end up on the bottom.
Thank you in advance for fixing it :)! 
great work you all <3

Chris

Patane97:
Fixed in https://github.com/throneteki/throneteki/pull/3541

PR #3541, opened 2024-09-07
Queenscrown fix to send cards to bottom of deck if no character is revealed

PeterShuck:
Issue: #3526 

Added fix for issue with Queenscrown not sending cards to the bottom of the deck when no characters are revealed.

Added unit tests for Queenscrown


Patane97 (approved):
Looks good to me, will merge with the next batch.

PR #3542, opened 2024-09-11
Laughing storm heads on spikes

### LINKED ISSUE #3520: Laughing Storm and Heads on Spikes
When The Laughing Storm is standing, Heads on Spikes still discards a card from your hand. It should not be able to when The Laughing Storm is standing.

PeterShuck:
Fixes #3520 

- Laughing Storm no longer allows discard from Heads on Spikes
- Updated effects and player classes to deny discard at random to GameAction (it was previously only card action)
- Added spec for Laughing Storm

### DIFF
--- a/server/game/GameActions/DiscardAtRandom.js
+++ b/server/game/GameActions/DiscardAtRandom.js
@@ -8,7 +8,11 @@ class DiscardAtRandom extends GameAction {
     }
 
     canChangeGameState({ amount = 1, player }) {
-        return amount >= 1 && player.hand.length >= 1;
+        return (
+            amount >= 1 &&
+            player.hand.length >= 1 &&
+            player.hand.some((card) => card.allowGameAction('discardAtRandom', { card }))
+        );
     }
 
     createEvent({
--- a/test/server/cards/25.3-FUtG/TheLaughingStorm.spec.js
+++ b/test/server/cards/25.3-FUtG/TheLaughingStorm.spec.js
@@ -1,0 +1,105 @@
+describe('The Laughing Storm', function () {
+    integration(function () {
+        beforeEach(function () {
+            const deck1 = this.buildDeck('lannister', [
+                'Heads on Spikes',
+                'A Noble Cause',
+                'Cersei Lannister (LoCR)'
+            ]);
+            const deck2 = this.buildDeck('baratheon', [
+                'Taxation',
+                'The Red Keep (Core)',
+                'The Laughing Storm',
+                'Bastard in Hiding',
+                'Stannis Baratheon (Core)'
+            ]);
+            this.player1.selectDeck(deck1);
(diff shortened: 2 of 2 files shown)


server/game/effects.js
```diff
@@ -1064,6 +1064,17 @@ const Effects = {
...
+        return {
+            targetType: 'player',
+            apply: function (player) {
+                player.cannotBeDiscardedAtRandom = true;
+            },
+            unapply: function (player) {
+                player.cannotBeDiscardedAtRandom = false;
+            }
+        };
+    },
```
Patane97, line 1077: I wouldn't be adding a new effect here (and checking `cannotBeDiscardedAtRandom` within `DiscardAtRandom.canChangeGameState`), as this is ultimately somethign that should be properly handled by `GameAction.allow`, where it checks the name of the game action (in this case, "discardAtRandom") can legally be performed on that player.

I would recommend opening a debug session, starting from `this.game.resolveGameAction` on `HeadsOnSpikes.js` and getting to the point of checking that `allow` function, and seeing why it's not recognising that the action cannot be performed on the Laughing Storm player.

server/game/GameActions/DiscardAtRandom.js
```diff
@@ -8,7 +8,7 @@ class DiscardAtRandom extends GameAction {
     }
 
     canChangeGameState({ amount = 1, player }) {
-        return amount >= 1 && player.hand.length >= 1;
+        return amount >= 1 && player.hand.length >= 1 && !player.cannotBeDiscardedAtRandom;
```
Patane97, line 11: As mentioned in other comment, I'd remove this in favor of ensuring `GameAction.allow` is working properly for the resolved action in Heads on Spikes. :)

Patane97 (changes requested):
Discussed further in discord private message, but to summarise what was said there:
> GameActions are basically any sort of action or effect that can happen in the game, and is something being lowly worked on over time - basically the aim is to move everything you can do within the game into each individual game action which handles the logic of whether it can legally happen, the messaging around it, and the event building structure of performing that action; doesn't necessarily relate to specific card actions or anything.
> 
> Within the chain of this.game.resolveGameAction there should be logic checking if an action can be performed - in this case for some reason HeadsOnSpikes is looking at the discardAtRandom game action, and even though the playing being discarded from is immune to the discardAtRandom action, it is not seeing that.
> 
> The change should likely be at the HOS level, rather than altering the effects on Laughing Storm & adding a separate check for discarding at random. It might help to check where canChangeGameState is called, as im pretty sure the parent GameAction has some "isAllowed" function or what-not which checks that the name of that acton (eg. "discardAtRandom") is legal, alongside checking if the gamestate can change

Patane97 (approved):
Looks good to me!

PR #3606, opened 2025-04-16
Maester Cressen(WK) now triggers on death

fanganga:


fanganga:
A fix for issue https://github.com/throneteki/throneteki/issues/3605 and https://github.com/throneteki/throneteki/issues/3557.

server/game/cards/25.6-WK/MaesterCressen.js
```diff
@@ -5,9 +5,26 @@ class MaesterCressen extends DrawCard {
...
+                    GameActions.placeCardUnderneath({
+                        card,
+                        parentCard: context.player.agenda,
+                        facedown: false
+                    })
+                )
+            )
+        });
+        //reaction has to be split to avoid double trigger in scenario where Cressen dies immediately after coming out of shadows
+        this.reaction({
```
Patane97, line 23: Great catch that this needs to have `location: ['dead pile']` in order to actually trigger from the dead pile, and that the separate reactions ensures he triggers once if he's instantly killed!

My suggestion would be to actually keep this as 1 reaction with a location of `location: ['play area', 'dead pile']`, and in the `onCardOutOfShadows` trigger explicitly check that `event.card.location === 'play area'` to avoid that double trigger. Having this as 2 reactions isn't bad per-say, but it could present unforseen strangeness with the game engine seeing this card has 2 reactions when it's actually meant to be 1. :)

Patane97 (approved):
Thank you for fixing this one!

This does fix those bugs, I'm okay to approve as it, but would _recommend_ merging into a single reaction instead of split (details in comment below) instead.

Awesome work my dude 💪

server/game/cards/25.6-WK/MaesterCressen.js
fanganga: Thanks, I'll give that a go and push when I've had a chance to test it.

fanganga: Pushed commit with your suggested change.

PR #3608, opened 2025-04-18
notConsideredToBeInPlotDeck is now independent of phase

fanganga:
https://github.com/throneteki/throneteki/issues/3602 occurred because the notConsideredToBeInPlotDeck flag on the scheme cards did not apply when recyclePlots() was triggered in the challenges phase as a result of Nymeria of Ny Sar, causing the recyclePlots function to behave as if the plot deck was not empty. I removed the condition from the persistentEffect so that the flag applies in all phases, not just the plot phase.

Other cases where notConsideredToBeInPlotDeck can be checked outside the plot phase are:
 - When At Prince Doran's Behest (SoD) is revealed by Nymeria of Ny Sar - I think it is correct for notConsideredToBeInPlotDeck to be active in this case.
 - In the Taste The Blood (ChoS) reaction - I don't see the relevance of notConsideredToBeInPlotDeck for Taste The Blood because I don't see a way a card with that flag could be in the revealed plots pile

### DIFF
--- a/server/game/cards/05-LoCR/TheRainsOfCastamere.js
+++ b/server/game/cards/05-LoCR/TheRainsOfCastamere.js
@@ -45,7 +45,6 @@ class TheRainsOfCastamere extends AgendaCard {
         });
 
         this.persistentEffect({
-            condition: () => this.game.currentPhase === 'plot',
             match: (card) => card.getType() === 'plot' && card.hasTrait('Scheme'),
             targetController: 'current',
             targetLocation: 'plot deck',
--- a/test/server/cards/05-LoCR/TheRainsOfCastamere.spec.js
+++ b/test/server/cards/05-LoCR/TheRainsOfCastamere.spec.js
@@ -194,4 +194,55 @@ describe('The Rains of Castamere', function () {
             expect(this.player1).not.toAllowAbilityTrigger('The Red Wedding');
         });
     });
+
+    integration(function () {
+        beforeEach(function () {
+            const starkDeck = this.buildDeck('stark', ['Trading with the Pentoshi']);
+            const martellRainsDeck = this.buildDeck('martell',[
+                '"The Rains of Castamere"',
+                'Trading with the Pentoshi',
+                'Filthy Accusations',
+                'A Song of Summer',
+                'Dorne (R)',
+                'Nymeria of Ny Sar',
+                'The Red Viper (Core)',
+                'House Dayne Knight'
+            ]);
+
+            this.player1.selectDeck(starkDeck);
+            this.player2.selectDeck(martellRainsDeck);
+            this.startGame();
+            this.keepStartingHands();
+
+            this.dorne = this.player2.findCardByName('Dorne', 'hand');
+            this.nym = this.player2.findCardByName('Nymeria of Ny Sar', 'hand');
+            this.viper = this.player2.findCardByName('The Red Viper', 'hand');
+            this.martellPentoshi = this.player2.findCardByName(
+                'Trading with the Pentoshi',
+                'plot deck'
+            );
+            this.songOfSummer = this.player2.findCardByName('A Song of Summer', 'plot deck');
+
+            this.player2.clickCard(this.viper);
+
+            this.completeSetup();
+        });
+
+        it('should not prevent plot recycling when usable plots are exhausted in the challenge phase', function () {
+            this.player2.selectPlot('Trading with the Pentoshi');
+            this.selectFirstPlayer(this.player2);
+            this.selectPlotOrder(this.player2);
+            this.player2.clickCard(this.dorne);
+            this.player2.clickCard(this.nym);
+            this.player2.clickCard(this.dorne);
+
+            this.completeMarshalPhase();
+            this.unopposedChallenge(this.player2, 'Power', this.viper);
+
+            this.player2.triggerAbility(this.nym);
+            this.player2.clickCard('A Song of Summer');
+
+            expect(this.martellPentoshi.location).toBe('plot deck');
+        });
+    });
 });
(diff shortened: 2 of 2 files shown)


server/game/cards/05-LoCR/TheRainsOfCastamere.js
```diff
@@ -45,7 +45,6 @@ class TheRainsOfCastamere extends AgendaCard {
         });
 
         this.persistentEffect({
-            condition: () => this.game.currentPhase === 'plot',
```
Patane97, line 48: I'm not sure this is the right fix here, primarily as those ***Scheme*** plots _should_ be considered in your plot deck outside the plot phase for other mechanical use. Easiest example off the top of my head; using **Nymeria of Ny Sar** to flip into **At Prince Doran's Behest,** to then flip into any of your ***Scheme*** plots; you should have them available to you during the challenges phase, but this change prevents that.

That might actually be a good additional test to add to the spec. :)

Patane97 (changes requested):
Love the investigation, and **definitely** love the added unit test - really professional!!

The fix here may need additional consideration, outlined a bit more in my comment. Happy to chat about potential alternatives! :)

server/game/cards/05-LoCR/TheRainsOfCastamere.js
fanganga: Thanks. I'd managed to miss that exact wording on the rains of Castamere card. For this interaction, it looks like the recycle should be triggered at the start of the plot phase when the cards become notConsideredToBeInPlotDeck. My current thought is to have a trigger on the agenda card that does the recycle check - would a ForcedReaction to the start of the plot phase be an appropriate tool, or is there a better one?

fanganga: I've been thinking some more around edge cases and noticed that **"The Rains of Castamere"** doesn't queue a recycle step, which could cause a problem where a Martell player uses **Nymeria of Ny Sar** to reveal their last non-scheme, then uses  **"The Rains of Castamere"** to reveal their last scheme. I'm wondering there's cause to refactor so that recycle plots is part of the RevealPlots step (with maybe an exception to allow additional logic like for **Battle of the Trident**), but I think that's getting beyond the scope of this bugfix.

My current plan is:
  * restore proper behaviour around using **At Prince Doran's Behest** from **Nymeria of Ny Sar**
  * add alternative handling to recycle plots appropriately so a player cannot get themselves stuck with **"The Rains of Castamere"** and **Nymeria of Ny Sar** - I'm still thinking of putting a forced reaction to the start of the plot phase on **"The Rains of Castamere"**, but if there's a more appropriate solution, I'd be happy to hear it .
  * Check the behaviour when a player uses **Nymeria of Ny Sar** followed by **"The Rains of Castamere"** to completely empty their plot deck in the challenges phase, and add a queued recycle step to **"The Rains of Castamere"** if necessary

fanganga: I've tried implementing a ForcedReaction to the plot phase starting on  **"The Rains of Castamere"**, but that introduces an issue when both players have **"The Rains of Castamere"** - the game will generate a prompt for which one happens first, breaking some existing tests and exposing the slightly dirty workaround to the players. 

An alternative is to add recycle step to the plot phase immediately after it starts - all the unit tests pass, but I'm not happy about making a change so deep, and adding extra checks that will be irrelevant when **"The Rains of Castamere"** isn't in play.

Patane97: If I'm honest, I think the actual resolution _should_ be what you mention here, and I'd encourage that than a bandaid fix
> I'm wondering there's cause to refactor so that recycle plots is part of the RevealPlots step (with maybe an exception to allow additional logic like for Battle of the Trident), but I think that's getting beyond the scope of this bugfix.
And ultimately, I don't actually think it will be as difficult as one may think.

I think removing the [recyclePlots step from the PlotPhase](https://github.com/Patane97/throneteki/blob/941c2dd09f2aeb0254978b013d1e9688db7c4c48/server/game/gamesteps/plotphase.js#L17) and moving it to be after the [onPlotRevealed child event within revealplots.js](https://github.com/Patane97/throneteki/blob/941c2dd09f2aeb0254978b013d1e9688db7c4c48/server/game/gamesteps/revealplots.js#L64) seems the best way to handle this.

One important thing to note is that plots are only checked for recycle after the When Revealed effect is fully complete (i.e.. after 1.3 iii in the Rules Reference). This is important for Behest, as it means that if Behest is ever your last plot (whether naturally or flipped via something like Ny Sar), you should have no plots to flip into for it's **When Revealed** effect. Due to this, I think you need to ensure that the recycle child event happens after the onPlotRevealed effect within `revealplots.js`.

Hopefully this makes sense; happy to chat over discord if you wanted a better explanation! (Deathlysteve on discord)

fanganga: Following discussion on Discord this is parked.

Per Rules As Written, under https://thronesdb.com/rulesreference#Plot_Card
_When a player reveals the final card of his or her plot deck, after all just revealed plot cards have entered play and all necessary "When Revealed" abilities have resolved, each plot card in that player's used pile is returned to his or her plot deck. The just revealed plot card is not yet in the used pile, and remains revealed._

The rules as written do not currently account for cases such as this where a plot deck becomes "empty" through means other than revealing a plot, and so a fix would require a change to the rules of the game, approved by GOT.

Patane97:
Going to close this for now, and once we get more official ruling from GOT I will open up a separate issue to be linked to, as it's closer to implementing an update rather than fixing a bug.

PR #3609, opened 2025-04-18
Fix for issue 3592: kill effect from Tarle

fanganga:
Issue https://github.com/throneteki/throneteki/issues/3592 - Tarle the Thrice Drowned's kill effect was able to target characters that could not be killed, due to the target matcher properties not specifying the 'kill' action as eg on Mag the Mighty, meaning the action didn't filter possible targets based on whether or not a kill action would affect game state.

Tested with Valyrian Steel Armor and The Eyrie, but not Long May He Reign


Patane97 (approved):
Spot on, nice catch!

PR #3610, opened 2025-04-22
Incorporated queing of RecyclePlots into  RevealPlots process

fanganga:
Work on the **Nymeria of Ny Sar**/ **"The Rains of Castamere"** interaction revealed inconsistencies in the use of recyclePlots - some actions that could empty the plot deck, such as revealing plots in the plot phase and **Nymeria of Ny Sar**, queued a recyclePlots step while others such as **"The Rains of Castamere"** did not.

This PR:
 - causes the code that creates a WhenRevealed window to be queued to queue a recyclePlots step immediately after
 - removes recyclePlots steps that are now redundant
 - introduces integration tests for motivating cases and edge cases where I anticipated an incorrect approach to this issue causing problems (mainly **At Prince Doran's Behest** interactions)

### DIFF
--- a/server/game/gamesteps/eventwindow.js
+++ b/server/game/gamesteps/eventwindow.js
@@ -70,6 +70,12 @@ class EventWindow extends BaseStep {
 
         if (this.event.getConcurrentEvents().some((event) => event.name === 'onPlotRevealed')) {
             this.openAbilityWindow('whenrevealed');
+            this.event
+                .getConcurrentEvents()
+                .filter((event) => event.name === 'onPlotRevealed')
+                .forEach((event) =>
+                    this.game.queueSimpleStep(() => event.plot.controller.recyclePlots())
+                );
         }
     }
 }
--- a/server/game/gamesteps/revealplots.js
+++ b/server/game/gamesteps/revealplots.js
@@ -61,7 +61,8 @@ class RevealPlots extends BaseStep {
         });
 
         for (let plot of plots) {
-            event.addChildEvent(new Event('onPlotRevealed', { plot: plot }));
+            let plotRevealedEvent = new Event('onPlotRevealed', { plot: plot });
+            event.addChildEvent(plotRevealedEvent);
         }
 
         return event;
--- a/server/game/cards/21-FtR/BattleoftheTrident.js
+++ b/server/game/cards/21-FtR/BattleoftheTrident.js
@@ -92,7 +92,6 @@ class BattleoftheTrident extends AgendaCard {
         this.game.queueStep(new RevealPlots(this.game, [context.target]));
         this.game.queueStep(
             new SimpleStep(this.game, () => {
-                context.player.recyclePlots();
                 if (context.player.plotDiscard.length > 0) {
                     let removeFromGameAction = this.abilities.reactions.find(
                         (reaction) => reaction.title === 'removeFromGameMarker'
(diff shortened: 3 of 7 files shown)


server/game/gamesteps/eventwindow.js
```diff
@@ -70,6 +70,12 @@ class EventWindow extends BaseStep {
 
         if (this.event.getConcurrentEvents().some((event) => event.name === 'onPlotRevealed')) {
             this.openAbilityWindow('whenrevealed');
+            this.event
+                .getConcurrentEvents()
+                .filter((event) => event.name === 'onPlotRevealed')
+                .forEach((event) =>
+                    this.game.queueSimpleStep(() => event.plot.controller.recyclePlots())
+                );
```
Patane97, line 78: This is almost correct; I would not recommend attaching this logic to the `EventWindow` class; instead this should be part of the `RevealPlots` class.

I believe the quickest/easiest way would be to, for each plot being revealed, create a "onPlotsRecycled" event, with the handler of that event simply calling `event.plot.controller.recyclePlots()`, and then `addChildEvent` that event to the "onPlotRevealed" event right after it was created.

server/game/gamesteps/revealplots.js
```diff
@@ -61,7 +61,8 @@ class RevealPlots extends BaseStep {
         });
 
         for (let plot of plots) {
-            event.addChildEvent(new Event('onPlotRevealed', { plot: plot }));
+            let plotRevealedEvent = new Event('onPlotRevealed', { plot: plot });
+            event.addChildEvent(plotRevealedEvent);
```
Patane97, line 65: I mention in the comment above on recycling to happen within RevealPlots; this is probably the spot for it (and you creating that `plotRevealedEvent` is almost there

Patane97 (changes requested):
Definitely on the correct track with this, however the wrong place to implement that change. Comments should outline, and give us a discord message if you need any help! :)

Also absolutely loving the tests, nice work!!

server/game/gamesteps/eventwindow.js
fanganga: I agree that making the change in the `RevealPlots` class rather than the `EventWindow` class is cleaner. Adding an event with a handler that invokes the `recyclePlots()` method as a child of the "onPlotRevealed" event when the "onPlotRevealed" event is created introduces sequencing issues with plots like **At Prince Doran's Behest**, triggering the recycle before the **When Revealed**. I'll have another look for a solution that gets around the sequencing issues without having to put plot-specific code in  the `EventWindow` class, but probably not today.

Patane97: Yeah that makes sense; probably the wrong place then, but if you `thenExecute` it at the end of the '_TOP_LEVEL_REVEAL_' event (notably, after the `openInterruptWindowForAttachedEvents` is called), that should be equivalent of doing it at the same step that's happening now, but will also cover scenarios where `RevealPlots` is called outside of the plot phase.

Patane97: @fanganga I think your above suggestion is good enough to go with, especially as it also fixes that additional interaction you mention with cards like Bitterbridge + Cavalry.

Only thing I'd confirm is whether this will queue multiple plot shuffled for 1 player if they revealed multiple times (eg. will Nysar > Behest > *something* result in 3 recyclePlot triggers?). If so, just fix that and I'm happy to approve 💪

fanganga: Just tested, and it looks good
PlotRecycle

Patane97:
@fanganga is there anything remaining for this one, or does it require a re-review & approval?

Patane97 (approved):

PR #3613, opened 2025-05-08
Gilbert of the Vines is now restricted to the Challenge phase

fanganga:


fanganga:
For issue https://github.com/throneteki/throneteki/issues/3535

PR #3615, opened 2025-05-21
Sea Song, Jon Arryn and War Scorpion test for invalid action on choices

fanganga:
Fix for https://github.com/throneteki/throneteki/issues/3537.

It seems the issue is due to the applicability of the effect not being checked properly when effects are selected from a choice menu. Most cards with multiple choice effects have if statements in the actions that prevent the effect from doing anything when inappropriate - eg the stand effect on Khaleesi.  The GainGold and DrawCard functions implement limits for eg High Sparrow.

Ideally, I'd want to implement a more generalised solution so that prompts for invalid actions are disabled and multi-choice actions where all options are ineligible are untriggerable, but for the sake of having a quick bugfix for now, I've implemented a solution similar to what exists on other cards to prevent Sea Song and Jon Arryn from causing power gain when prevented by Uneasy Truce and War Scorpion from killing a character protected by eg The Eyrie

### DIFF
--- a/server/game/cards/02.2-TRtW/TheReader.js
+++ b/server/game/cards/02.2-TRtW/TheReader.js
@@ -21,15 +21,20 @@ class TheReader extends DrawCard {
                         amount: 1
                     }))
                 },
-                'Discard 3 cards': () => {
-                    this.game.addMessage(
-                        "{0} uses {1} to discard the top 3 cards from each opponent's deck",
-                        this.controller,
-                        this
-                    );
-                    for (let opponent of this.game.getOpponents(this.controller)) {
-                        opponent.discardFromDraw(3);
-                    }
+                'Discard 3 cards': {
+                    message:
+                        "{player} uses {source} to discard the top 3 cards from each opponent's deck",
+                    gameAction: GameActions.simultaneously((context) =>
+                        this.game.getOpponents(context.player).map((opponent) =>
+                            GameActions.discardTopCards({
+                                player: opponent,
+                                amount: 3,
+                                source: context.source
+                            }).thenExecute((event) => {
+                                this.game.addMessage('{player} discards {topCards}', event);
+                            })
+                        )
+                    )
                 }
             }
         });
--- a/server/game/cards/02.2-TRtW/TheReader.js
+++ b/server/game/cards/02.2-TRtW/TheReader.js
@@ -1,4 +1,5 @@
 import DrawCard from '../../drawcard.js';
+import GameActions from '../../GameActions/index.js';
 
 class TheReader extends DrawCard {
     setupCardAbilities(ability) {
@@ -13,11 +14,12 @@ class TheReader extends DrawCard {
             },
             limit: ability.limit.perPhase(1),
             choices: {
-                'Draw 1 card': () => {
-                    if (this.controller.canDraw()) {
-                        this.controller.drawCardsToHand(1);
-                        this.game.addMessage('{0} uses {1} to draw 1 card', this.controller, this);
-                    }
+                'Draw 1 card': {
+                    message: '{player} uses {source} to draw 1 card',
+                    gameAction: GameActions.drawCards((context) => ({
+                        player: context.player,
+                        amount: 1
+                    }))
                 },
                 'Discard 3 cards': () => {
                     this.game.addMessage(
--- a/server/game/cards/24-TSoW/WarScorpion.js
+++ b/server/game/cards/24-TSoW/WarScorpion.js
@@ -30,8 +30,7 @@ class WarScorpion extends DrawCard {
                 this.game.resolveGameAction(
                     GameActions.ifCondition({
                         condition: (context) =>
-                            !context.target.isMatch({ trait: ['Army', 'Dragon'] }) ||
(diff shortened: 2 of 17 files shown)


server/game/cards/24-TSoW/WarScorpion.js
```diff
@@ -30,7 +30,8 @@ class WarScorpion extends DrawCard {
                 this.game.resolveGameAction(
                     GameActions.ifCondition({
                         condition: (context) =>
-                            !context.target.isMatch({ trait: ['Army', 'Dragon'] }),
+                            !context.target.isMatch({ trait: ['Army', 'Dragon'] }) ||
+                            !context.target.allowGameAction('kill'),
```
Patane97, line 34: I'm not sure about this one; the condition is here to check if the chosen character cannot be killed, then it automatically will just remove them from the challenge rather than giving the option.

Was this allowing protected character (eg. The Eyrie) to die?

Patane97 (changes requested):
Your analysis seems solid as per usual, and great tests; the proper fix isn't actually as hard as you think though (promise this time 😅), so I would advocate for that instead:

Update `AbilityChoicePrompt.js` to "disable" any buttons which fail the `choice.gameAction.allow(this.context)` check.
  - This will visibly disable any options which are not allowed.
  - For safety, I'd also update the `chooseAbilityChoice` method to double check they are allowed to execute it before resolving (same "allow" check as above).
  - You'll likely need to update your tests with `toHaveDisabledPromptButton` instead to confirm they can't do it.

In regard to the **War Scorpion** change, I'm not following so much - outlined in my comment there. :)

Also in a general sense, for issues like these (where it's addressing more niche card interaction bugs) I would usually advocate for the more proper fix, as "temporary" fixes like these tend to be more permanent and can cause confusion later down the line. For something game-breaking or problematic, quick fixes are fine. :)

server/game/cards/24-TSoW/WarScorpion.js
fanganga: Yes, in an instance built from the current master branch, it's possible to target eg Drogon protected by the Eyrie, choose to kill and have Drogon die.

fanganga:
I am going to continue working towards a proper fix (maybe prioritise it a bit more now you've outlined how you see it working. I don't know how your release plans are, but I figured since the approach here was consistent with what's on existing cards it would be worth having this PR around in case a release it could slot into came up before the rework of the choice menu was ready.

server/game/cards/24-TSoW/WarScorpion.js
Patane97: Yeah cool, sounds fine with me, although if you implement the fix on the choose logic, not having this allowGameAction check would simply mean they "choose" between a greyed out "kill" option, and the remove option.

Up to you; if you want to automate the choice for War Scorpion by adding this condition, I'm fine with this addition :)

fanganga: I've improved the AbilityChoicePrompt and updated those cards that had choices implemented with functions that could straightforwardly be replaced with GameActions for compatibility. This is now ready for re-review.

server/game/cards/02.2-TRtW/TheReader.js
```diff
@@ -13,11 +14,12 @@ class TheReader extends DrawCard {
...
-                        this.game.addMessage('{0} uses {1} to draw 1 card', this.controller, this);
-                    }
+                'Draw 1 card': {
+                    message: '{player} uses {source} to draw 1 card',
+                    gameAction: GameActions.drawCards((context) => ({
+                        player: context.player,
+                        amount: 1
+                    }))
                 },
                 'Discard 3 cards': () => {
```
Patane97, line 24: Not absolutely necessary, but this could also be updated to be `GameActions.simultaneously`, mapped against all opponents to `GameActions.discardTopCards` with an amount of 3

test/server/cards/11.3-SoKL/UneasyTruce.spec.js
```diff
@@ -0,0 +1,23 @@
...
+            this.startGame();
+            this.keepStartingHands();
+            this.jon = this.player1.findCardByName('Jon Arryn', 'hand');
+            this.player1.clickCard(this.jon);
+            this.completeSetup();
+            this.selectFirstPlayer(this.player1);
+        });
+
+        it('should prevent a player gaining power outside of challenges when the trigger is a multiple-choice action', function () {
+            //TODO - rework choices so choices that have no effect are disabled
```
Patane97, line 19: Is this comment meant to be here?

Patane97 (approved):
Nicely done @fanganga - looks great!

My only two comments are ones of minor improvement or questioning - all in all it looks good. Happy to merge in next update, but if you wanted to address the comments you're welcome to 💪

server/game/cards/02.2-TRtW/TheReader.js
fanganga: Done

test/server/cards/11.3-SoKL/UneasyTruce.spec.js
fanganga: Done

PR #3628, opened 2025-06-03
Corrected Wyman's log text on redesigned card to properly reflect effect

fanganga:


fanganga:
fixes https://github.com/throneteki/throneteki/issues/3627

PR #3632, opened 2025-06-06
Bugfix: Set Claim to Zero.

fanganga:
Fixed cases where game checks claim value to avoid case where set to zero is confused with not set - the issue was causing **Burning on the Sand** and **Lysa's Letter** not to have their required effect on the game state.

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -702,7 +702,7 @@ class Game extends EventEmitter {
             let valueGetter;
             switch (stat) {
                 case 'claim':
-                    if (!player.activePlot.claim.setValue) {
+                    if (player.activePlot.claim.setValue === null) {
                         effect = Effects.modifyClaim(value);
                     } else {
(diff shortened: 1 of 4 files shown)


server/game/game.js
```diff
@@ -702,23 +702,23 @@ class Game extends EventEmitter {
             let valueGetter;
             switch (stat) {
                 case 'claim':
-                    if (!player.activePlot.claim.setValue) {
+                    if (player.activePlot.claim.setValue === null) {
```
cryogen, line 705: I don't really like comparing explicitly with null as there's nothing that stops someone using null and undefined interchangeably. I don't really know what the issue is but can this be if (player.activePlot.claim.setValue > 0) instead?

cryogen: Ah, gotcha. This is one of those effects that pins a specific value and it can be pinned to 0. Yeah if we're using that pattern elsewhere it would be ok to use that here

fanganga: Done

PR #3668, opened 2025-10-08
Implementing Lord of the Waters

Patane97:
# 🆕 Lord of the Waters ⛵ 🎏 
Implements all 20 cards from the new Lord of the Waters pack. All cards are currently available at https://thronesdb.com/set/LotW.

### 🔧 Engine Adjustments
Some engine adjustments to account for certain cards:
- Allowing card abilities to be "cannot be canceled" for **Conspirator's Blade**.
  - Includes an alterations for Daggers in the Dark and Narrow Escape, as they can cancel themselves.
- Adding hand count modifications for **Qartheen Galleas**.
  - All cards which refer to a hand size (eg. comparing sizes, or fixed number like Dothraki Honor Guard) will be using that calculated number rather than the actual hand size. All other effects which target physical cards in hands remain unchanged. Quite tricky, and may have missed some effects.
- Adding a variable of gold to keep for **King's Landing Dromond**.
- Adding maximum power gain per round for **Edric Storm (LotW)**.
  - Includes some improvements to power gain effects and checking.
- Improving properties which rely on stacking effects, such as controller changing, loyalty changing and now cannotBeCanceled changing, with the addition of a simple `StackProperty` to handle the inner logic rather than repeating it all.
- General improvements to wording, such as actually mentioning how much unspent gold players return in taxation (and how much they keep, if applicable).

### ⏫ Potential Improvements
- [ ] When effects are to gain power simultaneously, such as renown, and the amount being gained would exceed the maximum gainable power, it should automatically ask the player to choose the order of resolution, ensuring that it only allows you to trigger abilities that keep you under that maximum.
- [ ] Fix a bug with "first X this round" reducers not considering cards meeting that condition before the effect enters play.

### ☑️ Remaining Tasks
- [x] **Before merging this PR**, merge JSON data (https://github.com/throneteki/throneteki-json-data/pull/186), and pull JSON data changes into this branch.
- [x] **After merging this PR**, pull data changes into playtesting JSON data, update development packs, and push changes to ThronesDB. @Patane97 will handle this.
- [ ] **After merging this PR**, pull card changes into playtesting, resolve conflicts (there should be a few with above engine adjustments; accept all incoming changes), and push.

### DIFF
--- a/deck-helper/AgendaRules.js
+++ b/deck-helper/AgendaRules.js
@@ -340,6 +340,23 @@ const agendaRules = {
             }
         ]
     },
+    // Trading with Braavos
+    26080: {
+        mayInclude: (card) => card.type === 'location' && hasTrait(card, 'Warship') && !card.loyal,
+        rules: [
+            {
+                message: 'Cannot include more than 1 copy of each non-limited location',
+                condition: (deck) => {
+                    const locations = deck.drawCards.filter(
+                        (cardQuantity) =>
+                            cardQuantity.card.type === 'location' &&
+                            !hasKeyword(cardQuantity.card, /Limited/)
+                    );
+                    return locations.every((location) => location.count <= 1);
+                }
+            }
+        ]
+    },
     // Draft Agendas
     // The Power of Wealth
     '00001': {
--- a/server/game/GameActions/GainPower.js
+++ b/server/game/GameActions/GainPower.js
@@ -7,22 +7,26 @@ class GainPower extends GameAction {
     }
 
     message({ card, amount = 1, context }) {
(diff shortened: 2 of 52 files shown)


Patane97 (commented):
All looks good; noticed a bug which is not caused by this released, but noticed by it with **Ibbenese Whaler**; technically our "reduce the first X you do this round" effects do not consider a card of that caliber already being marshaled before that cards entered play.

For Ibbenese Whaler, it means if you marshal something into Shadows, then bring Ibbenese Whaler into play, the next card you marshal into shadows will be reduced when it shouldnt. Same thing would happen with Donella Hornwood.

PR #3697, opened 2026-02-13
fix we light the way

### LINKED ISSUE #3695: We Light the Way doesnt work on shadow cards.
[We Light the Way](https://thronesdb.com/card/20040) seems to have condition to not be applied to shadow cards.

Duckbringer:
Fixes #3695


Patane97:
@Duckbringer I assume merging https://github.com/throneteki/throneteki/pull/3699 means this can be closed/discarded?

Duckbringer:
I think so, it appears I messed up commits/merges and it was contained in the other PR. 

Thanks a lot :)

PR #3729, opened 2026-04-19
Events & Deck System Overhaul

### LINKED ISSUE #3719: Events & Deck System Overhaul
# Events & Deck System Overhaul

## Overview

This update covers a significant rework of the events and deck systems, including draft event support, variant and legality list separation, deck locking fixes, and deck data structure optimisation.

---

## 1. Deck Data Structure
- [ ] Add optional `format` to deck model (backend). Keep as `undefined` on existing decks.
- [ ] Add optional `variant` to deck model (backend). Keep as `undefined` on existing decks.
- [ ] Add optional `pool` property to decks - stored as `{ "cardcode": quantity }` records (e.g. `{ "25001": 8 }`). Keep as `undefined` on existing decks.
- [ ] Update deck storage to save only card codes and quantities rather than full card objects (similar to above). Primarily to save space.
- [ ] Write a MongoDB migration script to convert existing decks from full card objects to `cardcode: quantity` pairs

---

### Format Selections
- [ ] Add a format dropdown & a variant dropdown to the deck editor (frontend), showing all available formats (Joust, Melee, Draft) and their variants. Changing this format + variant will validate against latest legality of that variant (e.g. Joust - Standard would validate on the latest Standard variant of joust)
- [ ] If the deck's `format` & `variant` are both `undefined`, the dropdowns are enabled and the player can select any format/variant. Otherwise, dropdowns are disabled & locked to those values. There should not ever be a scenario where variant is provided but format isn't and vice-versa.
- [ ] When a player attempts to change variant, check if any cards in the current deck are specific to the current variant. If so, display a warning popup informing them those cards will be removed. Allow them to confirm or cancel the change
- [ ] Related to above; filter cards based on variants (e.g. ToJ should show all ToJ cards and no standard). May require some sort of "mapping" logic to determine which variants link to which cards.
### Pool Restrictions
- [ ] If a deck has `format` & `variant` set but `pool` is `undefined`, the player can select any card from that format. Otherwise, restrict available cards in the editor to only those in the pool.
- [ ] Display a chip or flag in the editor when the player is working with a limited pool, so it is clear why certain cards are unavailable
### Draftmancer Import
- [ ] Add "Import from Draftmancer" button (separate to import; maybe update that to "Import from ThronesDB")
- [ ] When a deck is imported via Draftmancer, detect that it is a draft deck based on the import format
- [ ] Set `format` to "Draft" and `variant` to the appropriate variant based on the draftmancer file (e.g. "tower-of-joy")
- [ ] Populate the `pool` property with the cards available from that draft, if they are supplied
### Other changes
- [ ] Add the ability to change the faction card within the deck editor (available for all deck types, not just draft)
- [ ] Update deck read/save to use new card code format of `{ "cardcode": quantity }`. Will need to be applied outside of deck editor also (e.g. loading deck for game)

---

### Event Builder
- [ ] Separate `format`, `variant` and optional `legality` into distinct fields on the event model (backend)
- [ ] Update builder to allow the above three as dropdowns, cascading. If a format/variant has no legality, grey out that last dropdown
- [ ] Selecting any legality populates the rl editor at bottom of event builder (does not have to show). Add option to "edit" legality, which converts it to "Custom" and allows editing those pre-filled rl values
- [ ] Write a MongoDB migration/cleanup script for existing events to align with the updated structure
### Draft Specific
- [ ] Add draft `format` and `variant` options to the event builder
- [ ] Add a toggle for whether `pool` restrictions are enforced for that event
- [ ] Add instructions on the event settings page explaining how to use Draftmancer and import decklists - exact wording to be determined
### Lobby
- [ ] Players can manually select format, variant, and legality list when creating a game
- [ ] Alternatively, players can select an event, which auto-fills and grays out format, variant, and legality list based on event settings
### Deck Locking (Currently broken)
- [ ] Players may change their deck for an event before their first game
- [ ] Once a player completes their first game in an event, their deck is fully locked and cannot be changed. This differs from draft events, where editing within the pool is still permitted between games

---

## Draft Tournament Workflow
Just to help with understanding how draft tournaments should work, here's a loose breakdown of the intended workflow:
1. Event organiser creates the draft event in admin configs, configuring name, date, pool rules, and variant
2. The draft itself takes place on Draftmancer (external)
3. Each player imports their Draftmancer decklist into the website via the Decks section
4. The system detects the import as a draft deck and sets `format`, `variant` and `pool` accordingly (see above)
5. Players can freely edit their deck within their pool at any time - before and between games
6. When a player selects the deck for their first game in the event, they are locked into that deck for all subsequent matches in that event
7. Between games, editing is still permitted but remains constrained to the pool
8. If a player deletes their draft deck, a warning should inform them they will be removed from the event. Alternatively, forbid deleting event decks alltogether?

## Remaining Tasks/Thoughts
- **Pool integrity for draft events:** What prevents a player from editing their Draftmancer export before importing it, adding cards outside their actual pool? This is a sportsmanship concern but relevant for competitive events. To be addressed.
- **Event migration:** The exact shape of the events migration script is not yet determined. To be scoped when event fetch/save logic is being updated.
- **Draftmancer instructions:** Exact wording for the in-app Draftmancer guide to be confirmed with collaborator.

Patane97:
# Events & Deck System Overhaul

## Overview
This update covers a significant rework of the events and deck systems, including draft event support, variant and legality list separation, deck locking fixes, and deck data structure optimisation.

---

Priority of this update has been more leaning toward events in preparation for Online Worlds 2026 - with the major requirement being to fix & improve the "deck locking" logic, as last year this was identified as not actually locking at all.

The major notes:
- Games, events & general deck validation is now split into three avenues; Format > Variant > Legality. It has been decided/designed this way to properly future-proof any new formats or variants being created, such as the current "Draft" format now having two separate variants ("Valyrian", and the new "Tower of Joy" - although only ToJ will be available on TiT initially). This also means that new variants could be introduced cleanly in the future, such as **Hand of the King** for Joust, and **Kingsmoot** for melee. Each format/variant *can have a legality*, but will not always have one.
- Deck locking now actually works, and has been improved to visually show a lock symbol on the appropriate deck. Additionally, selecting the deck on your first event game will present a disclaimer identifying how it works, and that starting that game will lock that deck permanently for the event. Subsequent games now will auto-select the deck (from the server, to avoid client-side tampering). You can view your locked deck in the deck editor, but there are now visuals identifying & preventing you from editing it client-side (with server-side validation remaining).
- Form validation for the event builder & new game have also been cleaned up/improved, as it was noted to be fairly inconsistent/broken during testing. Invalid form properties will now be accurately represented client side.

I will omitt the changes here related to draft, as they are treated as incomplete due to time considerations & priority of the above for online worlds.

More details in issue below
Closes #3719


client/Components/Decks/DeckHelper.js
```diff
@@ -1,5 +1,7 @@
 import { lookupCardByName } from './DeckParser';
 
+const ALLIANCE_CODE = '06018';
```
cryogen: Prefer this was part of a constants object. We don't use all caps anywhere else

client/Components/Decks/PoolInfo.jsx
```diff
@@ -0,0 +1,69 @@
+import React, { useMemo, useState } from 'react';
+import { Chip, Divider, Tooltip } from '@heroui/react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
+import classNames from 'classnames';
+
+const PoolInfo = ({ className, compact = false, isPool = false }) => {
```
cryogen, line 7: Not a fan of the dynamic typing of compact, but it appears not to be used anywhere so can we remove it instead?

client/Components/Games/NewGame.jsx
```diff
@@ -105,168 +113,240 @@ const NewGame = forwardRef(function NewGame(
...
-                                        ?.experimental && (
-                                        <AlertPanel
-                                            variant='warning'
-                                            message={`The ${gameFormat} format is experimental and may not work as expected. Please report any issues to the developers on Github.`}
-                                        />
-                                    )}
-                                    <GameOptions
-                                        formProps={formProps}
-                                        isDisabled={usingEventOptions}
+const NewGameForm = ({ quickJoin, onClosed = () => true }) => {
```
cryogen: This should be in its own file

client/constants.js
```diff
@@ -4,8 +4,35 @@ export const ItemTypes = {
...
 export const GameFormats = [
-    { name: 'joust', label: 'Joust', defaultMaxPlayers: 2 },
-    { name: 'melee', label: 'Melee', defaultMaxPlayers: 4, experimental: true }
+    {
+        name: 'joust',
+        label: 'Joust',
+        defaultMaxPlayers: 2,
+        variants: [
+            { name: 'standard', label: 'Standard' } //,
+            // { name: 'handoftheking', label: 'Hand of the King' }
```
cryogen: Please remove commented code or enable it as needed

client/constants.js
```diff
@@ -4,8 +4,35 @@ export const ItemTypes = {
...
+            // { name: 'kingsmoot', label: 'Kingsmoot' }
+        ],
+        experimental: true
+    },
+    {
+        name: 'draft',
+        label: 'Draft',
+        defaultMaxPlayers: 2,
+        variants: [
+            // { name: 'valyrian', label: 'Valyrian' },
```
cryogen: Please remove commented code or enable it as needed

client/redux/middleware/api.js
```diff
@@ -439,51 +439,21 @@ export const apiSlice = createApi({
...
-            invalidatesTags: [TagTypes.DraftCube]
-        }),
         saveEvent: builder.mutation({
             query: (event) => ({
                 url: `/events/${event._id || ''}`,
                 method: event._id ? 'PUT' : 'POST',
                 body: event
             }),
-            invalidatesTags: (result, error, arg) => [{ type: TagTypes.Event, _id: arg._id }]
+            invalidatesTags: (result, error, arg) => [{ type: TagTypes.Event, id: arg.id }]
```
cryogen: This change seems incorrect. Mongo keys are ._id

client/redux/middleware/api.js
```diff
@@ -439,51 +439,21 @@ export const apiSlice = createApi({
...
         }),
         getEvent: builder.query({
             query: (eventId) => {
                 return {
                     url: `/events/${eventId}`
                 };
             },
-            providesTags: (_result, _error, arg) => [{ type: TagTypes.Event, _id: arg }]
+            providesTags: (_result, _error, arg) => [{ type: TagTypes.Event, id: arg }]
```
cryogen: and this one

server/services/DeckService.js
```diff
@@ -282,12 +300,40 @@ class DeckService {
     }
 
     async userAlreadyHasDeckForEvent(username, eventId) {
-        let deckForEventAlreadyExists = await this.decks
-            .findOne({ username: username, eventId: eventId })
-            .catch(() => {
-                throw new Error(`Unable to fetch deck with parameters ${username} and ${eventId}`);
+        return !!this.getDeckForEvent(username, eventId);
```
cryogen: Needs an await here or it will always return truthy

server/services/DeckService.js
```diff
@@ -282,12 +300,40 @@ class DeckService {
...
+        } catch (err) {
+            throw new Error(`Unable to fetch deck with parameters ${username} and ${eventId}`);
+        }
+    }
+
+    async useDeckForEvent(deckId, eventId) {
+        const deck = await this.getById(deckId);
+
+        //if the eventId is set on the deck, check if the user already has a deck with the same eventId
+        if (deck.eventId) {
```
cryogen: The duplicate check in useDeckForEvent only fires when the deck already has an eventId set, but this method is called precisely when it doesn't, so the check is always skipped:
```

if (deck.eventId) {  // always null/undefined when doing useful work
    if (await this.userAlreadyHasDeckForEvent(deck.username, deck.eventId)) {
        throw ...
    }
}
```
This allows a user to assign multiple decks to the same event by calling this on two unassigned decks in sequence. The check should be against the incoming eventId parameter instead:

```
if (await this.userAlreadyHasDeckForEvent(deck.username, eventId)) {
    throw new Error(`User ${deck.username} already has a deck configured for event ${eventId}`);
}
```

This also makes the if (deck.eventId) guard redundant and can be dropped entirely.

server/services/DeckService.js
```diff
@@ -1,53 +1,65 @@
...
 import { validateDeck } from '../../deck-helper/index.js';
 import logger from '../log.js';
+import ServiceFactory from './ServiceFactory.js';
 
 class DeckService {
-    constructor(db, cardService) {
+    constructor(db) {
         this.decks = db.get('decks');
-        this.cardService = cardService;
+        this.db = db;
```
cryogen, line 9: Should "initialise" the services required here not use ServiceFactory throughout the code

server/services/DeckService.js
```diff
@@ -1,53 +1,65 @@
...
 
-        formattedDeck.status = validateDeck(formattedDeck, {
-            packs: this.packs,
-            gameFormats: this.gameFormats,
-            restrictedLists: this.restrictedLists
-        });
+        let { eventId, format, variant, legality } = options;
+        if (eventId && eventId !== 'none') {
+            const event = await ServiceFactory.eventService(this.db).getEventById(eventId);
+            format = event.format;
```
cryogen: Can event ever be null (like if the  event is deleted but the decks arent updated)? If so this will blow up the server

cryogen (changes requested):

cryogen:
deck.legality is a dynamic field. It could be implemented as two fields which would a) mean the existing code is a simpler and you dont need to do typeof checks and b) makes migrating to sql when it comes easier

client/Components/Decks/DeckHelper.js
Patane97: Moved/fixed

client/Components/Games/NewGame.jsx
Patane97: Moved to its own file; `GameNameLengthLimit` has been exported to be used in both files

client/constants.js
Patane97: Done

client/constants.js
Patane97: Done

client/redux/middleware/api.js
Patane97: This one has been fixed, but notably it was still wrong; redux tags should always be `type` and `id`, not `_id` - but the arg id I changed to was incorrect.

Have updated to `id: arg._id`

client/redux/middleware/api.js
Patane97: This is actually correct; see above comment (`arg` here is the `eventId`)

client/redux/middleware/api.js
Patane97: Nevermind; although it SHOULDb e type & id, I see the rest use _id more frequently. Will grit my teeth and revert this to _id to not blow out the scope of this update :P

client/redux/middleware/api.js
Patane97: Changed (see updated comment above)

server/services/DeckService.js
Patane97: Good catch, fixed.

server/services/DeckService.js
Patane97: Very nice catch, fixed!

server/services/DeckService.js
Patane97: As discussed in discord, have changed this to be a getter

server/services/DeckService.js
Patane97: Technically I dont believe it can based on the client-side requests thta use it, but its definitely a good idea to have a null-check on `event` here regardless.

Have added that now.

client/Components/Decks/PoolInfo.jsx
Patane97: As mentioned over discord, going to leave this as-is as we will be revisiting the draft implementation after online worlds (ran out of time for it here).

Can revisit this at that point

PR #3753, opened 2026-08-10
fix secret schemes phase

Duckbringer:


Patane97:
Already fixing in bulk in https://github.com/throneteki/throneteki/pull/3754 (didn't realise you'd fixed it here, sorry!)

